PR #43, opened 2015-01-05
IOUtils fsync for directory fix

laimis:
UnauthorizedAccessException is being thrown when FSDirectory uses fsync for files and then tries to fsync a directory. The current implementation does not work for directories, and it is not entirely clear what to do in those cases in a way that works across all platforms.

It seems like Lucene version is failing for directories as well and they are handing it by capturing IOException (which won't work with UnauthorizedAccessException):
https://issues.apache.org/jira/browse/LUCENE-5588

The discussion is still on going on the mailing list, checking this in and will adjust based on feedback received.

### DIFF
--- a/src/Lucene.Net.Core/Util/IOUtils.cs
+++ b/src/Lucene.Net.Core/Util/IOUtils.cs
@@ -443,6 +443,12 @@ namespace Lucene.Net.Util
         ///  because not all file systems and operating systems allow to fsync on a directory) </param>
         public static void Fsync(string fileToSync, bool isDir)
         {
+            // Fsync for folders appears to be "undefined" for Windows platform based on the blog article in the comments below
+            if (isDir && Constants.WINDOWS)
+            {
+                return;
+            }
+
             var retryCount = 1;
             while (true)
             {
--- a/src/Lucene.Net.Core/Util/IOUtils.cs
+++ b/src/Lucene.Net.Core/Util/IOUtils.cs
@@ -443,8 +443,11 @@ namespace Lucene.Net.Util
         ///  because not all file systems and operating systems allow to fsync on a directory) </param>
         public static void Fsync(string fileToSync, bool isDir)
         {
-            // Fsync for folders appears to be "undefined" for Windows platform based on the blog article in the comments below
-            if (isDir && Constants.WINDOWS)
+            // Fsync does not appear to function properly for Windows and Linux platforms. In Lucene version
+            // they catch this in IOException branch and return if the call is for the directory. 
+            // In Lucene.Net the exception is UnauthorizedAccessException and is not handled by
+            // IOException block. No need to even attempt to fsync, just return if the call is for directory
+            if (isDir)
             {
                 return;
             }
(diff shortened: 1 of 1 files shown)


uschindler:
In Java, the .NET UnauthorizedAccessException extends IOException in Java (see http://docs.oracle.com/javase/8/docs/api/java/nio/file/AccessDeniedException.html) so it works in Windows. Windows is the only platform where fsync on directories does not work, but we still try  We know for sure that it works in MacOSX and Linux, so we assert on that, but we don't check it in production, because later/earlier versions of those operating systems may not implement it. So the whole thing is "the best we can do".

synhershko:
@uschindler yes that was our understanding (see mail thread), and thanks for confirming that. Why the effort for fsync'ing a dir in the first place though?

Since Windows doesn't support this, and the CLR doesn't seem to have a way of doing that (probably as a result), shouldn't we be ok with just fsyncing all files in a Directory and be done with it?

uschindler:
fsync on directory is more than fsyncing all files in it (there is a difference between syncing metadata and syncing buffers). Mike McCandless did tests on Linux (he has a machine that he hardly switched off in a loop with remote hardware power plug) and he tested it: When machine came up, index was still alive (he tested Linux).

Fsync on directory flushes the file buffers of the directory *+ its metadata*. The first one is also done by syncing metadata of a file contained, yes that works, but the directory's metadata is not necessarily written. Because of that it could happen that the inode of the directory is there, but the entry in the parent directory not yet (because the directory's metadata was not yet synced). After hard kill of machine, the fsck will report a lost+found inode (the directory). Because of the lost dir entry, the directory itsself is not visible in its parent.

I would still try to fsync also in windows, maybe it works at some time. And this is what we are doing in Lucene Java - we just try the POSIX semantics, which *may* work on Windows. So it is just best effort.

uschindler:
This is why we fsync on directory, too: http://man7.org/linux/man-pages/man2/fsync.2.html

"Calling fsync() does not necessarily ensure that the entry in the
       directory containing the file has also reached disk.  For that an
       explicit fsync() on a file descriptor for the directory is also
       needed."

synhershko:
Ok @laimis, so to enable better support of Mono on non-Windows OSs lets bring back this for dirs as well, however lets make 2 changes: (1) handle UnauthorizedAccessException the same as IOException, (2) if its a folder on a Windows OS don't retry (lets avoid IW.Commit slowness for the common use case).

synhershko:
Thanks @uschindler, this makes sense, although all of that doesn't apply to Windows really :)

We'll be bringing this back for the sake of better non-Windows support through Mono

uschindler:
In Lucene 5, the fsync on directory may be more important, be cause we rely on NIO.2's atomic rename functionality (we changed commit logic to make use of that). So we fsync on the direcory after the atomic rename of the commit (segments) file.

synhershko:
Oh dear. We'll cross that bridge when we get there :)

uschindler:
I noticed something: http://msdn.microsoft.com/en-us/library/windows/desktop/aa365258(v=vs.85).aspx

It says, you can only open a direcory with FILE_FLAG_BACKUP_SEMANTICS... Maybe this is why its causing "AccessDenied". 
But on the other hand, the list of functions to call do not contain FlushFileBuffers! Wo it might be useless.

Unfortunately, we cannot do this from Java NIO.2, but you may try this from .NET:

http://msdn.microsoft.com/en-us/library/windows/desktop/aa363858(v=vs.85).aspx

laimis:
@synhershko sounds good, will bring back the fsync support for directories, will do so in a separate branch / PR. Thanks guys!

PR #45, opened 2015-01-10
Field cache sanity checker fixes

laimis:
Fixes for cache sanity checker and to Lucene.Net.Core/Index/CompositeReaderContext as to how it stores some data.

Cache sanity checker was throwing "value too large" exception when reading back the max double stored in the document. The issue is that Convert.ToString on double uses 15 precision digits when serializing so the value is rounded up making it to large when the value is extracted from the doc.  Using double's ToString and specifing "R" format makes it use 17 precision digits if needed.

Lucene.Net.Core/Index/CompositeReaderContext was creating a new array from "IList<IndexReaderContext> children" that was passed to it when the calling code was expecting it to store a reference to the passed in array. Lucene version also makes sure that unmodifiableList is stored (https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/index/CompositeReaderContext.java) so added UnmodifiableList implementation.


laimis:
Let me reopen a different PR, it will be cleaner after the rebase. Too many of the merge branch 'master' here...

PR #46, opened 2015-01-10
Field cache sanitychecker composite reader context fixes

laimis:
Cache sanity checker was throwing "value too large" exception when reading back the max double stored in the document. The issue is that Convert.ToString on double uses 15 precision digits when serializing so the value is rounded up making it too large when the value is extracted from the doc. Using double's ToString and specifing "R" format makes it use 17 precision digits if needed.

Lucene.Net.Core/Index/CompositeReaderContext was creating a new array from "IList children" that was passed to it when the calling code was expecting it to store a reference to the passed in array. Lucene version also makes sure that unmodifiableList is stored (https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/index/CompositeReaderContext.java) so added UnmodifiableList implementation.

UPDATE: after some more thought, decided not to introduce the UnmodifiableList and adjust code to use ReadOnly() of List.


synhershko:
@laimis nice. You should note we have a class `Collections` with implementations of unmodifiable* stuff based on the BCL

laimis:
@synhershko yeah, saw that but could not use it because it does items.ToArray() initialization of the immutable list, in essence losing the reference to the items list passed to it. The same issue that original code was suffering from. The code in that class could be a little bit better in how it passes data around, but since it is somewhat isolated, kept the code structure the same and just fixed the issue and moved on.

PR #47, opened 2015-01-15
Use reference instead of value type to match Lucene's logic

laimis:
Introduce a nullable long in DocValuesConsumer, instead of using just long. The need for this is best illustrated if you look at the changes in Lucene45DocValuesConsumer and in NumericDocValuesWriter. The Writer had iterator implementation that did not match Lucene version, ignoring cases of DocsWithField.Get(doc) evaluating to false in which case a null value needs to be returned. Null value executes a code branch in Lucene45DocValuesConsumer that was simply ignored before. The result of that is the resulting index structure being incorrect, and requests such as DocsWithFields would return true for all documents even though the index had documents that did not have a field in question. 

The tests that used to break are in TestLucene45DocValuesFormat and TestPerFieldDocValuesFormat, e.g. BaseDocValuesFormatTestCase.TestTwoNumbersOneMissing. This fixes about 20 failing tests by my count.

I was considering porting over Numeric class over, but I think we can do that in stages, if necessary. Since all the signatures were using long, migrating to long? was the first step to get the logic right and tests passing. And then we can improve on the approach itself.

I think I see more code places where iterators have been ported incorrectly and numbers are used in non-nullable form causing different code execution paths, but will get those as we go down the list of failing tests.

### DIFF
--- a/src/Lucene.Net.Core/Codecs/Lucene45/Lucene45DocValuesConsumer.cs
+++ b/src/Lucene.Net.Core/Codecs/Lucene45/Lucene45DocValuesConsumer.cs
@@ -123,7 +123,6 @@ namespace Lucene.Net.Codecs.Lucene45
             bool missing = false;
             // TODO: more efficient?
             HashSet<long> uniqueValues = null;
-            // LUCENE TODO: is this necessary because of multiple iterations?
             
             if (optimizeStorage)
             {
--- a/src/Lucene.Net.Core/Codecs/Lucene45/Lucene45DocValuesConsumer.cs
+++ b/src/Lucene.Net.Core/Codecs/Lucene45/Lucene45DocValuesConsumer.cs
@@ -109,12 +109,12 @@ namespace Lucene.Net.Codecs.Lucene45
             }
         }
 
-        public override void AddNumericField(FieldInfo field, IEnumerable<long> values)
+        public override void AddNumericField(FieldInfo field, IEnumerable<long?> values)
         {
             AddNumericField(field, values, true);
         }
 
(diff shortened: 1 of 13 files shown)


src/Lucene.Net.Core/Codecs/Lucene45/Lucene45DocValuesConsumer.cs
```diff
@@ -123,11 +123,13 @@ internal virtual void AddNumericField(FieldInfo field, IEnumerable<long> values,
             bool missing = false;
             // TODO: more efficient?
             HashSet<long> uniqueValues = null;
+            // LUCENE TODO: is this necessary because of multiple iterations?
```
synhershko: Lucene.NET TODOs are marked LUCENENET TODO

laimis: That's good to know! Thanks!

And actually that comment needs to go, I used another list to track objects in one iterator so that I can pass them to the next but then changed the approach.

synhershko:
Merging now - will this get us below 150 failures? :)

synhershko:
No need to port `Numeric` IMO, all usages I've seen of it mattered only for inheritance which I could get by one way or another.

laimis:
@synhershko ha! We wish. If I had to guess it will get us down to 220 range, you know how those random tests go. But what it will allow me to do is move on to the next failures. This particular issue had me stomped. It is so subtle, I just couldn't see what was wrong until looking at the byte level of what the index streams are writing... it was brutal.

PR #50, opened 2015-01-18
Lucene42 codec fixes

laimis:
This PR contains fixes for numerous bugs in Lucene42 codec port, especially around FST functionality. 

First commit is all about changing from long to long? for FST instances to make sure that code branches such as this one https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Core/Util/Fst/FST.cs#L465 evaluate properly. With long, Outputs.Merge gets called when it should not have and then an exception "Specified method is not supported." is thrown. (here is one example in TC: http://teamcity.codebetter.com/viewLog.html?buildId=178533&tab=buildResultsDiv&buildTypeId=LuceneNet_Core#testNameId8117509285720903580)

Then once the fix allowed further code execution, noticed that FST and some other classes related to it use cast to object when comparing generic type values. Equals should be used as that will work for reference and value types properly. Next commits take care of that.

This also gets a bunch of TestLucene42DocValuesFormat tests to pass.

### DIFF
--- a/src/Lucene.Net.Core/Codecs/Lucene42/Lucene42DocValuesProducer.cs
+++ b/src/Lucene.Net.Core/Codecs/Lucene42/Lucene42DocValuesProducer.cs
@@ -69,7 +69,7 @@ namespace Lucene.Net.Codecs.Lucene42
         private readonly IDictionary<int, NumericDocValues> NumericInstances = new Dictionary<int, NumericDocValues>();
 
         private readonly IDictionary<int, BinaryDocValues> BinaryInstances = new Dictionary<int, BinaryDocValues>();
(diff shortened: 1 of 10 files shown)


geobmx540:
@laimis you're a committer, you should be able to commit without a PR - unless you just want to review :)

synhershko:
I'm good with this process - let's always have an extra set of eyes for intrusive changes

geobmx540:
I like it.

laimis:
@synhershko @geobmx540 I like the PR approach too. So if you guys don't mind it either, let's continue that way. I am still new with this code base so it is good to have someone look over the changes.

Hopefully the explanations are useful. Let me know if you want me to change anything.

PR #53, opened 2015-01-21
Expressions & other contrib projects

hakeemsm:
Expressions compiles & some tests pass. The others fail due to Antlr issues.
Contrib projects don't even compile and are not added to the sln


synhershko:
Thanks for this PR

Can you please resend this PR, this time with only the Expressions bits (code + tests), and also rebase this against the current master so there are no merge conflicts?

It will be probably easier for you to create a new branch following apache/master and copying just the Expression projects files over, committing and pushing that

PR #55, opened 2015-01-22
Fixing Compilation for Build Server

wwb:
Trying to keep up with development I tested the build on TeamCity and I found that things did not quite build correctly in that environment. The project files for Lucene.net.Tests.Queries and Lucene.net.Tests.Classification needed a slight tweak to make msbuild look in the right place all the time.

Once we get these added I can turn on the build / tests for those libraries on the CI server.


synhershko:
Thanks - merged now

For future PRs, please try to work on a branch, and also squash committs to fewer when posssible...

PR #57, opened 2015-01-24
Fixes in Util and Store packages.

guidotag:
This changes make almost all Util's test pass. The exceptions are the "ignored" tests, and the TestIndexableBinaryStringTools' ones. The latter ones won't succeed unless we change IndexableBinaryStringTools' implementation, which is wrong. Nevertheless this class is used nowhere in the project.

Also, they fix some of the Store's tests and related classes.


synhershko:
Thanks

However, this fixes too many things, some of the fixes are wrong or needs further discussion.

I'll cherry-pick some commits based on what I think are good fixes, and then for the rest please open a dedicated PRs on a branch tracking `apache/lucene.net/master` for each fix and we will discuss each separately in its own PRs.

Sorry but there are too many intrusive changes in this PR that need discussions separately.

guidotag:
Following Itamar's suggestion, I started to create separate PRs for each group of related fixes.
I'll close this PR as soon as I finish the new ones.

PR #59, opened 2015-01-25
Merge logic fixes

laimis:
This PR contains fixes for segment merging logic for numeric and binary fields. Lucene.Net port was ignoring docsWithField parameter and as a result the merged segment had incorrect values in field infos.  

SegmentMerger needs to pass docsWithField downstream so that the iterators could use it to determine which value to return for byte or numeric fields. Here is the Lucene version, https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/index/SegmentMerger.java#L159, you can see how docsWithField is passed to Merge* calls.

Subsequently, re-introduced docsWithField usage in DocValuesConsumer. Here is the Lucene version for reference: https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/codecs/DocValuesConsumer.java. Note how it uses docsWithField to determine if to return null or a value.

And then related to that BinaryDocValuesWriter also needs to use docsWithField list to return proper BytesRef values. Here is Lucene version for reference: https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/index/BinaryDocValuesWriter.java

Also removed commented out ports of iterators. They were taking up space and making things more confusing.

You can see the tests that fail without these changes in TestLucene42DocValuesFormat or TestLucene45DocValuesFormat, but there might be more. Example failing test from TC: http://teamcity.codebetter.com/viewLog.html?buildId=179687&tab=buildResultsDiv&buildTypeId=LuceneNet_Core#testNameId7364447424024347711

### DIFF
--- a/src/Lucene.Net.Core/Codecs/DocValuesConsumer.cs
+++ b/src/Lucene.Net.Core/Codecs/DocValuesConsumer.cs
@@ -112,19 +112,19 @@ namespace Lucene.Net.Codecs
         /// The default implementation calls <seealso cref="#addNumericField"/>, passing
         /// an Iterable that merges and filters deleted documents on the fly.
         /// </summary>
-        // LUCENE TO-DO This is a bit wacky
-        public virtual void MergeNumericField(FieldInfo fieldInfo, MergeState mergeState, IList<NumericDocValues> toMerge/*, IList<Bits> docsWithField*/)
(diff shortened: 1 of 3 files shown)


eladmarg:
awesome! I knew there is a bug in the segments, but couldn't find it, thanks :)

PR #62, opened 2015-01-27
Fixes in the Fsync method and in a helper function in a Store's test.

guidotag:


synhershko:
@laimis can you please review, based on your changes there?

guidotag:
@laimis: Sure. I found that if you set file to null before entering the using's scope and then, inside it, change it to reference a File object, it does not dispose the resources correctly (namely, it does not close the file we are opening). I did an isolated test to reproduce it (constructed a dummy IDisposable object and used the using instruction in the same way), and it also failed. I am no C# expert, so I wasn't completely sure why this happened. Nevertheless, it is reasonable that if you change the reference of the variable inside the using's scope, bad things could happen.

guidotag:
Here is my simple test to reproduce the issue:

static void Main(string[] args) {          
            TestNoDispose();
            TestDispose();
            Console.ReadLine();
}

static void TestNoDispose() {
            DisposableObject obj = null;
            using (obj) {
                obj = new DisposableObject();
                Console.WriteLine("Should NOT dispose after this message...");
            }
}

static void TestDispose() {
            DisposableObject obj;
            using (obj = new DisposableObject()) {
                Console.WriteLine("Should dispose after this message...");
            }
        }
}

class DisposableObject : IDisposable {
        public DisposableObject() {
            Console.WriteLine("Constructing...");
        }
        public void Dispose() {
            Console.WriteLine("Disposing...");
        }
}

synhershko:
Actually that is probably a good catch and can explain many failures we see with files being left open. I'll review and test locally later tonight.

laimis:
@guidotag actually I was referring to TestWindowsMMap.RmDir, was just curious as it seems to be doing the same thing so it would be good to know what was the issue that you found.

The changes in Fsync look good to me.

guidotag:
@laimis sorry, didn't realize you commented on that one. For some reason, when I changed that it wasn't working for me and I couldn't find out why so I rewrote it. Today I tested the original version again and it works just fine, so there should be no need for changing it.

PR #65, opened 2015-01-31
fix LogMergePolicy IsMerged logic

laimis:
This fixes the issue where if LogByteSizeMergePolicy is used, the merges run "forever". At the end of the day the issue was not with the byte size policy, but with the LogMergePolicy "IsMerged" logic. segmentIsOriginal check was inverted, forcing the evaluation to continue and check IsMerged on already merged index and for LogByteSizeMergePolicy that returned false when in reality that check should have never happened. Changed the code to match Lucene:

https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/index/LogMergePolicy.java#L181


synhershko:
Good catch!

PR #67, opened 2015-02-01
Support negative numbers in Lucene40 codec

laimis:
Lucene40 codec picks different methods of storing numeric values based on what data type is large enough to represent them. The issue was with the numbers that can fit in sbyte, they were being returned as byte due to our migration. Negative numbers -128 or greater were being returned as positive numbers.

This passes all remaining failing Lucene40 codec tests. Here is one example:

http://teamcity.codebetter.com/viewLog.html?buildId=180212&tab=buildResultsDiv&buildTypeId=LuceneNet_Core#testNameId-8934306368705483624

Test expects to get back -10 but is getting back 246 instead.

### DIFF
--- a/src/Lucene.Net.Core/Codecs/Lucene40/Lucene40DocValuesReader.cs
+++ b/src/Lucene.Net.Core/Codecs/Lucene40/Lucene40DocValuesReader.cs
@@ -241,7 +241,7 @@ namespace Lucene.Net.Codecs.Lucene40
 
             public override long Get(int docID)
             {
-                return Values[docID];
+                return (sbyte)Values[docID];
             }
         }
 


synhershko:
There's some more such places, I'm on it

synhershko:
FYI Lucene.Net.Index.BaseDocValuesFormatTestCase.TestTwoDocumentsMerged() still fails

laimis:
@synhershko are you sure you are running fully compiled after merging my changes? In TC builds and several local builds that test case passes as expected. Make sure you did a full rebuild if you ran it locally after merging my changes, that has gotten me before.

The sbyte that you added will not make a difference because the header value is either 0 or 1 value, works in byte and sbyte format.

synhershko:
Yes, works now

PR #68, opened 2015-02-01
Marking a few things long running. (+1 squashed commits)

wwb:
here is the long running test attribute and also the attribute applied to one of the longer-running tests.

[4f28661] adding LongRunningTest attribute.


synhershko:
What does this give us exactly? I assume it depends on appropriate TC configs?

wwb:
Or any other test runner -- it adds a category called "LongRunningTest" to decorated items. You can then use your test runner to include or exclude that attribute depending on what you want to do.

PR #69, opened 2015-02-01
Lucene.net.codecs.memory

geobmx540:
Finalizing conversion of the Lucene.Net.Codecs (Memory), as well as refactoring to account for change from long to long?

### DIFF
--- a/src/Lucene.Net.Codecs/Memory/DirectPostingsFormat.cs
+++ b/src/Lucene.Net.Codecs/Memory/DirectPostingsFormat.cs
@@ -108,7 +108,7 @@ namespace Lucene.Net.Codecs.Memory
         public override FieldsProducer FieldsProducer(SegmentReadState state)
         {
             var postings = ForName("Lucene41").FieldsProducer(state);
-            if (state.Context.Context != IOContext.Context.MERGE)
+            if (state.Context.Context != IOContext.Context_e.MERGE)
             {
                 FieldsProducer loadedPostings;
                 try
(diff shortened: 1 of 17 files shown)


synhershko:
Looks good! one lesson we learned post-sbyte removal is some calculations are relying on the value to be -128 - 128 in range, so sbyte conversion is required before making the actual calc. Just a thing to bear in mind, though once tests are ported and compiling those should float up.

src/Lucene.Net.Codecs/Memory/DirectPostingsFormat.cs
```diff
@@ -353,40 +344,30 @@ public DirectField(SegmentReadState state, string field, Terms termsIn, int minS
...
+                hasOffsets_Renamed = fieldInfo.FieldIndexOptions.Value.CompareTo(IndexOptions.DOCS_AND_FREQS_AND_POSITIONS) > 0;
+                hasPayloads_Renamed = fieldInfo.HasPayloads();
 
                 BytesRef term;
                 DocsEnum docsEnum = null;
                 DocsAndPositionsEnum docsAndPositionsEnum = null;
-//JAVA TO C# CONVERTER WARNING: The original Java variable was marked 'final':
-//ORIGINAL LINE: final index.TermsEnum termsEnum = termsIn.iterator(null);
-                TermsEnum termsEnum = termsIn.iterator(null);
+                TermsEnum termsEnum = termsIn.Iterator(null);
```
synhershko: `GetEnumerator` ?

PR #70, opened 2015-02-03
Minor fixes in store

guidotag:
See the comment on the penultimate commit.


synhershko:
This is very weird that the file isn't being created despite the call to `file.Create();`

I'd like to have @uschindler's eyes on this before pulling this in

uschindler:
@rcmuir 's patches for Java 7 changed that, as far as I remember. I will ask him. In fact if a directory does not exist FSDirectory in Lucene 5.0 will create it in its constructor. It calls Files.createDirectories(Path), which is "mkdir -p": http://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createDirectories(java.nio.file.Path,%20java.nio.file.attribute.FileAttribute...)

uschindler:
I still have no idea why this one test in Lucene seems to se a different behaviour. Maybe its some bug, I will look into it possibly a bit later. Just came back from FOSDEM have to catch up first.

synhershko:
I don't mind Lucene.NET 4.8 being non-compatible with Java Lucene 4.8 in that regard since it doesn't affect anything but our "create new index" story and the exception we throw when the index doesn't exist, but I'd rather make sure first there isn't special reason for things being the (weird) way they are now

/cc @rmuir

guidotag:
@uschindler I didn't know what Lucene 5.0 does. I just tried to reproduce Lucene 4.8 behavior.
Essentially the only difference between TestDontCreate in Java and C# was that the first one didn't create the directory in the FS. That didn't make a difference in the final result. When I did the change in GetCanonicalPath, Lucene.Net's test started to fail because there was no directory to delete (Lucene does not fail because it checks if the directory exists before attempting the deletion).

Note that the tests TestNRTCachingDirectory/TestNoDir and TestFileSwitchDirectory/TestNoDir now pass thanks to this change.

laimis:
@synhershko @guidotag I just setup Lucene 4.8 on my local with source code and IDE so that I can debug (some of those multithreading bugs are brutal to catch otherwise), and can confirm that in Java version, FSDirectory does not create the directory, it only says that it does in the comment. If you look at the code, it make sense:

https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/store/FSDirectory.java#L128

There is no explicit create being called, and constructing a File class with a path does not create a file that represents such a path. When the port was being done, whoever read that comment for "getCanonicalPath" that said "creating it if it doesn't exist." simply got mislead.

uschindler:
Hi,
I checked the Lucene Core 4.8 code. In fact it does not create the directory. The comment is just bogus. I removed it in Lucene 4.10: http://svn.apache.org/r1657033

In Lucene 5, the move to Java NIO.2 now creates the direczory in the ctor. This is in my opinion wrong to do, so I started a discussion. In fact, Mike McCandless stated, that there was already a bug open about this: https://issues.apache.org/jira/browse/LUCENE-1464

The thread is CC'ed to the dev@lucenenet list.

synhershko:
ok @laimis @guidotag let's mimic the actual behavior and not the intended behavior. The tests should pass without changing them, and the exceptions thrown as meaningful as they can be. In fact, lets remove `GetCanonicalPath` altogether if possible.

Thanks @uschindler !

guidotag:
OK @laimis. TestDontCreate needs to be changed just a little bit to check that the directory exists (it actually won't) before trying to delete it. Besides, there seems to be no way of getting rid of GetCanonicalPath by using a DirectoryInfo method.

uschindler:
Hi,
In Lucene Java the GetCanonicalPath is needed for the following: Java has the problem that creating a native lock only affects *other* processes from using the directory. It does not prevent the same process from writing to the same index directory. Because of that NativeFSLockFactory keeps a static set of Path names to also ensure locking in the same JVM/process. To make sure that this locking works we need the "canonical name" of the directory, otherwise it is not guaranteed that locking really works.
If this is not an issue on windows (if you lock a file, the file lock is visible to the same process, too), then you won't need that.

synhershko:
I'm assigning this to myself. I'll work on this next week and get some well-defined behavior suitable for the CLR in place.

uschindler:
Thanks! Looks fine.

PR #73, opened 2015-02-04
replace dictionary with hashmap to support null keys

laimis:
Replace Dictionary with HashMap for tracking SegmentsToMerge in the IndexWriter. Lucene is using HashMap, and I can confirm by running Lucene 4.8 tests that sometimes there are cases when IndexWriter will insert null value as key. It is happening here:

https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L1788

When instance of OneMerge is created and passed around there is no guarantee that segment info will be set.

This now passes TestAddIndexesWithThreads and TestAddIndexesWithRollback in Lucene.Net.Index.TestAddIndexes.

### DIFF
--- a/src/Lucene.Net.Core/Index/IndexWriter.cs
+++ b/src/Lucene.Net.Core/Index/IndexWriter.cs
@@ -247,7 +247,7 @@ namespace Lucene.Net.Index
         internal readonly IndexFileDeleter Deleter;
 
         // used by forceMerge to note those needing merging
-        private IDictionary<SegmentCommitInfo, bool?> SegmentsToMerge = new Dictionary<SegmentCommitInfo, bool?>();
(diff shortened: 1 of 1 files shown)


synhershko:
I would rather we don't use the `HashMap` support class if that is the only difference

Going through the IndexWriter code, there's no real meaning for a null segment marker - if it doesn't point to any specific segment it doesn't really matter if we track the boolean value for it or not.

I pushed a different fix, let's see how the tests behave

laimis:
@synhershko sounds good, let's see what happens. The tests look good, I guess we will catch the issues if there are any and adjust. Closing this PR.

PR #74, opened 2015-02-04
Don't modify the collection while iterating over it

laimis:
Collection was being modified while iterated, which throws an exception in .NET. I am not 100% sure if try / finally is necessary but the comment was suggesting to do it as such. @synhershko any thoughts on this one? Is the idea here that we want to remove the ones that were successfully iterated over before the exception occurred?

Several tests fail due to this, Lucene.Net.Index.TestRollingUpdates.TestUpdateSameDoc is one of them as an example from the latest TC master build.

### DIFF
--- a/src/Lucene.Net.Core/Index/DocumentsWriterFlushControl.cs
+++ b/src/Lucene.Net.Core/Index/DocumentsWriterFlushControl.cs
@@ -734,20 +734,28 @@ namespace Lucene.Net.Index
         {
             lock (this)
             {
-                IEnumerator<BlockedFlush> iterator = BlockedFlushes.GetEnumerator();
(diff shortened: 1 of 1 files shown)


synhershko:
No idea why it was implemented that way. Not sure why it needed locking as well - the method isn't synchronized in the original java code.

Maybe it was an attempt at freezing state - but again I see no reason for that

I just pushed another attempt at a fix, which is more in line with the original Java code - please review

https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterFlushControl.java#L563

laimis:
Cool, yeah it looked strange. The error is gone from the test logs.

PR #78, opened 2015-02-06
Expressions test cases fixed

hakeemsm:
Some still fail due to changes in TokenStream


synhershko:
Thanks I'll review and squash shortly

PR #83, opened 2015-02-06
Fix for TestThreadSafety.

guidotag:


src/Lucene.Net.Tests/core/Store/TestDirectory.cs
```diff
@@ -134,6 +134,7 @@ public override void Run()
                             try
                             {
                                 IndexInput input = outerBDWrapper.OpenInput(file, NewIOContext(Random()));
+                                input.Dispose();
```
synhershko: probably makes more sense to put a using statement here?

PR #88, opened 2015-02-08
fix lru hash map

eladmarg:


synhershko:
What is this fix about?

Also, have you implemented an LRU cache from scratch? any reason why you didn't use other implementations such as http://www.codeproject.com/Articles/23396/A-High-Performance-Multi-Threaded-LRU-Cache ?

eladmarg:
my previous implementation was wrong, this is not implemented from scratch, and its working.
(tests are passing)

laimis:
@eladmarg can you point us to the tests that are being fixed by this PR?

laimis:
I see one test in the PR request, but the comment referenced plural tests so want to make sure we look at them all.

eladmarg:
currently only TestHashMap

synhershko:
To be frank my main concern is with the actual implementation - the Java version is relying on a JVM implementation, and if we are to write our own we should at least try and get other proven ones

eladmarg:
Itamar, this is proved one, 
Taken from here: https://gist.github.com/nakhli/6686251

the previous was mine implementation, and it was wrong.

eladmarg:
can it be merged?

synhershko:
I want to review this properly and see what our options are for trusted LRU implementations, not going to merge for now until we get it right for sure.

PR #92, opened 2015-02-10
Fix on the PQ.

guidotag:
Important: Please cherry pick Feb09's commits. Feb06 was already pushed to apache/master.

The minor change is just something I saw while trying to fix another bug. A String's length might not be the same as the represented utf16's length.

The second change is on the PQ. Heapify was swapping incorrectly. This solved TestFuzzyQuery/TestFuzziness and TestFuzzyQuery/TestTieBreaker.

### DIFF
--- a/src/Lucene.Net.Core/Search/FuzzyTermsEnum.cs
+++ b/src/Lucene.Net.Core/Search/FuzzyTermsEnum.cs
@@ -132,9 +132,8 @@ namespace Lucene.Net.Search
 
             // convert the string into a utf32 int[] representation for fast comparisons
             string utf16 = term.Text();
-            //LUCENE TO-DO
             //this.TermText = new int[utf16.codePointCount(0, utf16.Length)];
-            this.TermText = new int[utf16.Length];
+            this.TermText = new int[Character.CodePointCount(utf16, 0, utf16.Length)];
             for (int cp, i = 0, j = 0; i < utf16.Length; i += Character.CharCount(cp))
             {
                 TermText[j++] = cp = Character.CodePointAt(utf16, i);
--- a/src/Lucene.Net.Core/Support/PriorityQueue.cs
+++ b/src/Lucene.Net.Core/Support/PriorityQueue.cs
@@ -187,7 +187,8 @@ namespace Lucene.Net.Support
             while (pos > 0)
             {
                 var parentPos = (pos - 1) / 2;
-                if (_baseHeap[parentPos].CompareTo(_baseHeap[pos]) < 0)
+                // Remember it is a Min-Heap.
+                if (_baseHeap[pos].CompareTo(_baseHeap[parentPos]) < 0)
                 //if (_comparer.Compare(_baseHeap[parentPos].Key, _baseHeap[pos].Key) > 0)
                 {
                     ExchangeElements(parentPos, pos);
--- a/src/Lucene.Net.Core/Util/Automaton/SpecialOperations.cs
+++ b/src/Lucene.Net.Core/Util/Automaton/SpecialOperations.cs
@@ -1,3 +1,4 @@
+using System;
 using Lucene.Net.Support;
 using System.Collections;
 using System.Collections.Generic;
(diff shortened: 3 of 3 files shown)


guidotag:
What do you think @synhershko ?

synhershko:
@guidotag this is mainly why I haven't pulled this in yet. I wasn't the one to bring this PQ implementation in so I wanted to review this carefully and haven't got the time to do that just yet.

Take a look here: https://github.com/apache/lucenenet/commit/0471466e0719b8f828e89fc7edaa481ad75ac2d8#diff-86df47eb5fd9dabe41b7624bcb97587aR362

This is another usage of the same PQ - it basically relies on `T` implementing `IComparable` - so while what you are suggesting is true, it is probably way more trickier than you think.

The right plan of attack here is to switch to using `IComparer<T>` (passed via a mandatory ctor param)

As a side-note, I'd very much like to verify this PQ implementation. I don't believe we have tests for it, and since its going to be quite central I want to validate its correctness, performance and thread-safety guarantees. Worth looking up in the history who brought it in and take it from there /cc @paulirwin

guidotag:
Why changing to an IComparer<T>? It is much more natural to think PriorityQueue<T> as a PQ containing elements of type T, rather that letting T be a comparator class.

Regarding the PQ's tests, I'll start working on that. Since it seems to be a critical piece of code, I'll send an email just to tell I'll be on it.

synhershko:
Because as you indicate, items are not compared based on one strategy always. See the link I posted for another example of this.

`IComparer` is more suitable for those scenarios. I'm not saying make `T` an `IComparer`, I'm saying lets remove the requirement for `T` to implement `IComparable` and instead provide PQ with an `IComparer`

guidotag:
Oh, I got it! Yep, that seems a much more flexible approach. Unless you want to do it, I'll take care of that too.

synhershko:
Go ahead!

guidotag:
I realized that there are two different implementations of a PQ, one in Util and this one in Support. Do we have two different ones for some reason? A couple of things I've noticed:

- The Java version uses the Util's PQ. Moreover there seems to be no Support package, which seems to contain emulations of Java features not present in C#. 
- The Support's one is only used in the Search/TopTermsRewrite class. The Util's one is much more referenced.
- The only PQ's methods invoked from TopTermsRewrite are: PriorityQueue(int capacity), PriorityQueue(), T Peek(), T Poll() (dequeueing without failure if the PQ is empty), bool Offer(T item) (insertion) and int Count. All of these methods are provided by Util's PQ, with the exception of Poll which could be added with little effort.

All this facts make me think that we should replace the Support's duplicate by the Util's preexisting PQ.

guidotag:
One more thing: The Util's one has some unit tests while the other one, as you remarked previously, does not.

synhershko:
The Util one is taken directly from Java Lucene:

https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/util/PriorityQueue.java

This is also why it has unit tests.

Can you please verify the port of this class, and the tests, and move to using that one?

Worth noting the TTR usage is basically wrong - it should have used the one in Util: 
https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/search/TopTermsRewrite.java#L23

So let's verify the port of that PQ and move to using it, deprecating the one in Support. Once we stabilized everything I'll dig to see why it was introduced in the first place, and remove it after all required lessons were learned.

Good catch

guidotag:
That's a good plan! I'll start working on this.

guidotag:
@synhershko I finished reviewing Util's PQ and creating some more unit tests.
I could change this PQ to require a comparer in the constructors, rather than needing to inherit and overwrite the LessThan method, but, as you imagine, this would require changes all over the code.

This way of defining comparers via inheriting is flexible enough to allow two different comparers for the same data type, though you'll need to create two different PQ type's which is much less natural than just creating two different comparer functions and using them with the same generic PQ.

Passing a comparer seems a much better (and natural) approach to me, but since this would require a lot of changes I'd like to discuss it before doing any move.

eladmarg:
@guidotag I Agree with you, passing comparer seems more naturally, more generic and more readable code.
We're at .net, so we should follow the attitude.

how many files involved with this change?

guidotag:
It has 60 references from 20 different files.

eladmarg:
maybe it is possible to assign default comparer <T> as default value

synhershko:
@guidotag I think this can be closed now that #97 is merged? can the PQ from `Support` be removed now?

guidotag:
Not yet. It remains changing all the references to Support's PQ. I'll take care of this right now, and after PRing that I'll close this one.

PR #95, opened 2015-02-11
Use CompareOrdinal to compare field_renamed because unit tests can creat...

Chand2048:
...e a string with a trailing \0 char which is ignored in the default culture.

### DIFF
--- a/src/Lucene.Net.Core/Index/Term.cs
+++ b/src/Lucene.Net.Core/Index/Term.cs
@@ -149,13 +149,14 @@ namespace Lucene.Net.Index
         /// </summary>
         public int CompareTo(Term other)
         {
-            if (Field_Renamed.Equals(other.Field_Renamed))
+            int comp = string.Compare(Field_Renamed, other.Field_Renamed, StringComparison.Ordinal);
+            if (comp == 0)
             {
                 return Bytes_Renamed.CompareTo(other.Bytes_Renamed);
             }
             else
             {
-                return Field_Renamed.CompareTo(other.Field_Renamed);
+                return comp;
             }
         }
 


synhershko:
Can you point at some of those tests? I'm not sure we actually want to not-ignore \0

Chand2048:
The test I was working on was Lucene.Next.Index.TestPrefixCodedTerms.TestMergeRandom
I added a huge amount of test code to narrow down the problem and found that the MergedIterator's use of the PriorityQueue ended up returning terms out of order. An alternate solution would be to modify the random Unicode string generator to not include \0 chars. This also fixed it but digging deeper the default comparison for java is ordinal so this should match a port better.

Chand2048:
Here is some amusing sample code to show the problem.
static void Main(string[] args)
{
    String s1 = "abc";
    String s2 = "abc\0";

    Console.WriteLine("'{0}'", s1);
    Console.WriteLine("'{0}'", s2);
    Console.WriteLine("s1.Equals(s2)    returns {0}", s1.Equals(s2));
    Console.WriteLine("s1.CompareTo(s2) returns {0}", s1.CompareTo(s2));
    Console.WriteLine("string.Compare(s1, s2, StringComparison.Ordinal) returns {0}", string.Compare(s1, s2, StringComparison.Ordinal));
}

Output is
'abc'
'abc '
s1.Equals(s2)    returns False
s1.CompareTo(s2) returns 0
string.Compare(s1, s2, StringComparison.Ordinal) returns -1

synhershko:
@Chand2048 any reason why you closed this PR then ? :)

Chand2048:
I did not realize that deleting the branch would close this.

Chand2048:
Reopened, sorry about the confusion.

PR #96, opened 2015-02-11
Fix for TestIndexableField.TestArbitraryFields - In java, the byte class...

Chand2048:
... is signed and in C# it is unsigned. Changed the unit test to cast to a (byte) insetad of an (sbyte).

An alternate much deeper fix would be to change the Lucene.Net.Util.BytesRef class to use sbyte insetad of byte. Not sure if this class was intentionally setup this way or if it is a conversion error. Comments please.

### DIFF
--- a/src/Lucene.Net.Core/Index/DocumentsWriterDeleteQueue.cs
+++ b/src/Lucene.Net.Core/Index/DocumentsWriterDeleteQueue.cs
@@ -1,3 +1,4 @@
+using Lucene.Net.Store;
 using Lucene.Net.Support;
 using System;
 using System.Threading;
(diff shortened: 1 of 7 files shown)


eladmarg:
this is correct fix,
no need to change bytesref, this is on purpose.

Chand2048:
Great, thanks!

synhershko:
@eladmarg nothing is set in stone, please don't make such statements on behalf of the project

@Chand2048 sbyte on public APIs will break CLS compliance, this is mainly why we moved off it (also because it just looks weird to C# devs). If we spot anything fundamental that breaks because of this we will reconsider. In the meantime , it makes perfect sense to work with C# bytes and convert as needed (its a simple cast - bit representation stays the same)

PR #98, opened 2015-02-16
Chand8

Chand2048:


synhershko:
Thanks, I'll review shortly. Will probably cherry pick and not merge because there's too many discrete unrelated changes here. Would be great if future PRs will be isolated :)

PR #103, opened 2015-02-18
Port from Support's to Util's PQ.

guidotag:
Added a resizing functionality. Now the PQ resizes when it runs out of space. Since the former version was more restrictive (the size was fixed from the beginning) it should be backwards compatible. Nevertheless, it should be checked that no previous usage was expecting an IndexOutOfRangeException when adding elements to the queue.

Added a ToArray method.

Added tests.

With this change, TestsFuzzyQuery/TestFuzziness and TestFuzzyQuery/TestTieBreaker, from the Search package, now succeed.

### DIFF
--- a/src/Lucene.Net.Core/Search/TopTermsRewrite.cs
+++ b/src/Lucene.Net.Core/Search/TopTermsRewrite.cs
@@ -5,7 +5,7 @@ using System.Linq;
 
 namespace Lucene.Net.Search
 {
-    using Lucene.Net.Support;
+    using Lucene.Net.Util;
     using ArrayUtil = Lucene.Net.Util.ArrayUtil;
     using BytesRef = Lucene.Net.Util.BytesRef;
 
@@ -71,7 +71,7 @@ namespace Lucene.Net.Search
         public override Query Rewrite(IndexReader reader, MultiTermQuery query)
         {
             int maxSize = Math.Min(size, MaxSize);
-            PriorityQueue<ScoreTerm> stQueue = new PriorityQueue<ScoreTerm>();
+            PriorityQueue<ScoreTerm> stQueue = new ScoreTermPQ();
             CollectTerms(reader, query, new TermCollectorAnonymousInnerClassHelper(this, maxSize, stQueue));
 
             var q = TopLevelQuery;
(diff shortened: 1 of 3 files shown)


synhershko:
See https://github.com/apache/lucenenet/commit/a3220b5119f628866134ba3e4bd0dabdc4c9664f

Instead of resize, lets figure out what defaults are used in the Java version

guidotag:
Java's priority queue can manage an "unbounded" number of elements and to do so it uses a resizable representation. From PriorityQueue's javadoc: "A priority queue is unbounded, but has an internal capacity governing the size of an array used to store the elements on the queue. It is always at least as large as the queue size. As elements are added to a priority queue, its capacity grows automatically." So yes, it has an internal resizing procedure.

In the TopTermRewrite class it is not clear what size we will need at most:

https://github.com/guidotag/lucenenet/commit/9428451c8ee87b39efeb2c39b94eb153067a066e#diff-226e1a7768fe596cf640cfef43c6f60fL74

so we need the representation's size to grow dynamically.

guidotag:
In Lucene, TopTermRewrite is implemented the same way, relying on the dynamic capacity:

final PriorityQueue<ScoreTerm> stQueue = new PriorityQueue<ScoreTerm>();

synhershko:
I must be missing something - this is not using the Java's PQ, it's using this: https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/util/PriorityQueue.java#L31 which doesn't have a default ctor

guidotag:
Mmm it is using Java's PQ:

https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/search/TopTermsRewrite.java#L23

guidotag:
Anyway, it is strange that they decided to code their own PQ. I suppose they had a good reason for doing that. 

So this is why there where two PQ's in Lucene.Net...

synhershko:
What do you think we should do - keep with one PQ implementation and make it to support resizing (i.e. when `maxSize` is null), or keep using that other implementation from Support?

guidotag:
I definitely think we should only keep Util's PQ, because:
- Having two PQ's will probably confuse somebody else in the future.
- I prefer the raw array representation (and taking care of the resizes) rather than the List one. Anyway, all the operations are implemented in a similar way, and there shouldn't be big differences in their performances.
- It is tested.
- There is a way of going over the resize issue, as you point out.

Deciding in a constructor if it should be resizable sounds good, and it is a reasonable feature for a PQ. Also it shouldn't degrade the performance, since we only need to add a check in the Add method.

What do you think about adding a no-parameter constructor, and only allow resizes if constructed that way?

synhershko:
@guidotag sounds good

guidotag:
@synhershko done!

PR #104, opened 2015-02-18
Use the C# ThreadInterruptedException so the catches will properly contr...

Chand2048:
...ol the flow during interruption.

This fixes the two unit tests Lucene.Next.Index.TestIndexWriter.TestThreadInterruptDeadlock and Lucene.Next.Index.TestIndexWriter.TestTwoThreadsInterruptDeadlock

Lucene uses the ThreadInterruptedException for code flow control so it is important to use the C# ThreadInterruptedException to avoid duplicating the catch everywhere and preserve the flow.


synhershko:
Can you please remove the code you commented out? no reason for it to stay there, the fix looks legit AND we have all of it in the git history either way

Chand2048:
Very happy to do that. Git is better at keeping history than comments. I see lots of code that is commented out and I had assumed if it is related the the original java version we want to leave it. Can we generally remove code instead of comment?

synhershko:
Yes, PLEASE :)

synhershko:
Can you please squash into one commit and then rebase on master?

Chand2048:
Moved to new branch and cleaned up.

PR #106, opened 2015-02-19
The values returned from the uniqueValues does not match the order retur...

Chand2048:
...ned from .Keys.

Add a function to sort and then compare the output from the uniqueValues, deletes.Terms_Nunit().Keys and frozen set.


Chand2048:
Separating this to isolate the discussion and eventual acceptance.
Please let me know if this is an acceptable solution or not. Thanks!

Chand2048:
Excellent thanks!

PR #108, opened 2015-02-22
Index test fixes

laimis:
TestIndexWriterExceptions was failing because deletes were not reflected when the index was opened with the reader. The reason for that was that the documents were being sent with leading 0 in the id (e.g. "id:01" instead of "id:1") but deletes were executed against id without leading zero. Simple oversight of .NET port, Lucene adds the base and i first before converting to string.

TestParallelCompositeReader needed to be adjusted to match ported code that uses anonymous inner class helper so the type name is different.

### DIFF
--- a/src/Lucene.Net.Tests/core/Index/TestIndexWriterExceptions.cs
+++ b/src/Lucene.Net.Tests/core/Index/TestIndexWriterExceptions.cs
@@ -2169,7 +2169,7 @@ namespace Lucene.Net.Index
                 for (int i = 0; i < numDocs; i++)
                 {
                     Document doc = new Document();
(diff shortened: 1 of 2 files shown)


laimis:
Looks like parallel composite reader fix is addressed by @Chand2048 in https://github.com/apache/lucenenet/pull/102, I will close this and reopen only with TestIndexWriterExceptions fix.

synhershko:
@laimis what do you make of @Chand2048 's comment about this test not existing in the original Java code?

laimis:
oh yeah, thanks for pointing that out @synhershko. I see the test in Lucene, it is this: https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/test/org/apache/lucene/index/TestParallelCompositeReader.java#L392

PR #111, opened 2015-02-25
Fix for TestThreadInterruptDeadlock

Chand2048:
Use the C# ThreadInterruptedException so the catches will properly control the flow during interruption.
This fixes the two unit tests Lucene.Next.Index.TestIndexWriter.TestThreadInterruptDeadlock and Lucene.Next.Index.TestIndexWriter.TestTwoThreadsInterruptDeadlock
Lucene uses the ThreadInterruptedException for code flow control so it is important to use the C# ThreadInterruptedException to avoid duplicating the catch everywhere and preserve the flow.


Chand2048:
This is the squashed and rebased version.

PR #112, opened 2015-02-25
Fix for TestDirectoryReaderReopen.TestThreadSafety

Chand2048:
The java version uses a concurrent hash map so I updated VirtualMethod to use that.
After that change, we no longer needed the lock on the Reap() function.
Next the Get() needed to use a tryGet() and return the default if there was nothing in the collection.

### DIFF
--- a/src/Lucene.Net.Core/Util/VirtualMethod.cs
+++ b/src/Lucene.Net.Core/Util/VirtualMethod.cs
@@ -64,7 +64,7 @@ namespace Lucene.Net.Util
         private readonly Type BaseClass;
         private readonly string Method;
         private readonly Type[] Parameters;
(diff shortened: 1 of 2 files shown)


synhershko:
I'm holding off with merging this because I don't really like the way `WeakIdentityMap` is implemented and I'd like to review it a bit before deciding what to do with it. Hopefully I'll get to it in the next few days.

synhershko:
Ok just reviewed it - this class has only 2 usages and the fix looks legit enough...

I moved class using it to the TestFramework project; the other only usage warrant making `WeakIdentityMap` internal but we'll get to it when we are done by setting up `InternalsVisibleTo` properly for the test projects

PR #113, opened 2015-02-25
Fix for TestIndexReaderClose.TestCloseUnderException

Chand2048:
The IndexReader.NotifyReaderClosedListeners() did not properly suppress the exceptions thrown from the call to listener.OnClose(). Unfortunately there is no place to track the suppressed exception so it is just dropped.
Changing to return AggregateException will change the type and cause problems later for flow control.


Chand2048:
Rebased and removed the attempt to use AggregateException. This caused problems.

synhershko:
Merged. We should probably add a LUCENE TODO to tackle this later since we are swallowing the original exception

PR #114, opened 2015-02-26
Fix for TestPersistentSnapshotDeletionPolicy

Chand2048:
The PrepareIndexAndSnapshots function used a local Snapshots IList to store snapshots
instead of the SnapshotDeletionPolicy class storing the snapshots. This storage persists
between tests and causes failures. Specifically the TestBasicSnapshots() test would
fail when it was run after TestExistingSnapshots().

### DIFF
--- a/src/Lucene.Net.Tests/core/Index/TestPersistentSnapshotDeletionPolicy.cs
+++ b/src/Lucene.Net.Tests/core/Index/TestPersistentSnapshotDeletionPolicy.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.Diagnostics;
 
 namespace Lucene.Net.Index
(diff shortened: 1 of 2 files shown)


synhershko:
@Chand2048 this is exactly how this is being done in Apache Lucene - see https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/test/org/apache/lucene/index/TestSnapshotDeletionPolicy.java#L66

So we either missed a clean-up procedure that's being done there, or there's something else that we do differently. Either way, I'd like to avoid fixing the test - we may be covering up another problem.

Chand2048:
You are correct. Unfortunately my local copy of the java source does not have the global list. It actually puts the snapshots into the SnapshotDeletionPolicy instance passed into the function. I will get my source to the correct version and take a look again. Sorry for the confusion.

PR #117, opened 2015-03-01
match Lucene logic of not creating the directory

laimis:
There is still discussion going on about what to do with FSDirectory's GetCanonicalPath: https://github.com/apache/lucenenet/pull/70 @synhershko  has assigned this to himself to investigate. While he is doing that, this PR removes the offending code so that we match Lucene's behavior in that we don't create the directory if it does not exist. The failing tests related to this, TestIndexWriter.TestNoDocs and TestMultiPhraseQuery now pass.


synhershko:
I'll handle this in a few days, won't merge for now

PR #118, opened 2015-03-03
Index fixes

laimis:
TestSnapshotDeletionPolicy and TestPersistentSnapshotDeletionPolicy were failing because Snapshots list was not initialized on each test run. Results from previous test run would interfere and the test would fail. To reproduce, run the whole fixture as running a test by itself passes.

Marking TestAddIndexes.TestAddIndexMissingCodec with ignore because we are missing some codecs for that test to match what Lucene is doing.

Marking TestAtomicUpdate.TestAtomicUpdates with LongRunning attribute so that it runs on nightly only.

### DIFF
--- a/src/Lucene.Net.Tests/core/Index/TestAddIndexes.cs
+++ b/src/Lucene.Net.Tests/core/Index/TestAddIndexes.cs
@@ -1207,8 +1207,8 @@ namespace Lucene.Net.Index
         /*
          * simple test that ensures we getting expected exceptions
          */
-
         [Test]
+        [Ignore("We don't have all the codecs in place to run this.")]
         public virtual void TestAddIndexMissingCodec()
         {
             BaseDirectoryWrapper toAdd = NewDirectory();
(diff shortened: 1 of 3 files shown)


src/Lucene.Net.Tests/core/Index/TestAddIndexes.cs
```diff
@@ -1207,8 +1207,8 @@ public UnRegisteredCodec()
         /*
          * simple test that ensures we getting expected exceptions
          */
-
         [Test]
+        [Ignore("We don't have all the codecs in place to run this.")]
```
synhershko: Please don't put this as Ignore, the Codecs subproject is compiling IIRC and needs some work on tests, need to sync with @geobmx540 on status

laimis:
Sounds good, @synhershko. I figured it was safe to ignore as it had Lucene Net TODO comment in it to go back and fix it once all the required codecs are ported. I will open another PR with just "snapshots" changes which you can pull in.

laimis:
Never mind, I see that you merged it into the master.

@synhershko question for you about that test marked as ignored. It looks like Pulsing41PostingsFormat has been ported and compiles. Are we OK to reference Codecs project from core tests at this stage? Or should we wait until all the tests for codecs have been included and compiling. I see the Codec test files are checked in but they are not included in the project.

synhershko:
Yes, but if you can do some work on the Codec tests that will be great

synhershko:
Yes, but if you can do some work on the Codec tests that will be great

PR #120, opened 2015-03-04
don't use instance variables, clean up test

laimis:
Several tests in the fixture were using instance variables for directory and reader references and that was breaking the tests as they were not initialized to clean state. Refactored to get rid of instance variables and any shared state issues.

### DIFF
--- a/src/Lucene.Net.Tests/core/Index/TestTermsEnum.cs
+++ b/src/Lucene.Net.Tests/core/Index/TestTermsEnum.cs
@@ -431,41 +431,28 @@ namespace Lucene.Net.Index
             dir.Dispose();
         }
 
-        private Directory d;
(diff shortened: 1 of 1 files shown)


synhershko:
Merged, but can you please confirm those tests are now in-line with the Java one?

laimis:
@synhershko I can confirm that they are. What I was doing was run Java version side by side to make sure the tests did not rely on the instance variables in any way. In Lucene, again they benefit from the test runner initializing instance variables on each test run. In .NET that does not happen and we end up getting failures like these instead: http://teamcity.codebetter.com/viewLog.html?buildId=184897&tab=buildResultsDiv&buildTypeId=LuceneNet_Core#testNameId-4229341108728890812

PR #127, opened 2015-04-11
init bytesref on each iteration

laimis:
Fixes a bug where bytes were written to the index incorrectly if Lucene40Codec was used. The issue can be observed by following the data flow starting here:

https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.TestFramework/Codecs/lucene40/Lucene40DocValuesWriter.cs#L176

values.ToArray() enumeration instead of returning an array of different values, it returns an array with all members pointing to the same bytesref instance. This broke logic to detect min / max Lengths down the stream and as a result incorrect bytes were written to an index.

This bug did not show up until the Lucene40DocValuesWriter was rewritten to call ToArray(). Here is a gist illustrating the difference between the two usages with bug in place: https://gist.github.com/laimis/d49291d2f02a7f4830bf

The fix removes  BytesRef value = new BytesRef(); initialization and instead initializes new bytes ref instance anytime one should be returned.

There might be more places in code base that suffer from this, will need to review the iterators that use similar approach.

With the fix in place all Lucene40Codec tests pass.

### DIFF
--- a/src/Lucene.Net.Core/Index/BinaryDocValuesWriter.cs
+++ b/src/Lucene.Net.Core/Index/BinaryDocValuesWriter.cs
@@ -128,27 +128,7 @@ namespace Lucene.Net.Index
             int maxDoc = state.SegmentInfo.DocCount;
             Bytes.Freeze(false);
             dvConsumer.AddBinaryField(FieldInfo, GetBytesIterator(maxDoc));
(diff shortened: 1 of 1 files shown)


synhershko:
Thanks @laimis -- can you do said review?

laimis:
@synhershko yup, found several places and will get them in through the next commits.

PR #130, opened 2015-04-22
change how TopTermsRewrite instance of is checked

laimis:
Changing the logic that checks if rewrite method is an instance of some sort of TopTermsRewrite to match Lucene. Key change is here:

https://github.com/apache/lucenenet/compare/master...laimis:fuzzyquery_fix?expand=1#diff-605c6558db33049a7e4d93482e5432aeL67

Instead of checking for TopTermsRewrite< Q >, it seems like Lucene cares only if it is TopTermsRewrite based class but does not care what parameter type is used and it does not need to match SpanMultiTermQueryWrapper parameter type:

https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/java/org/apache/lucene/search/spans/SpanMultiTermQueryWrapper.java#L68

At least that's how interpreted it. The tests seem to pass and this also fixes failing test in TestSpanMultiTermQueryWrapper. TestFuzzy2

### DIFF
--- a/src/Lucene.Net.Core/Lucene.Net.csproj
+++ b/src/Lucene.Net.Core/Lucene.Net.csproj
@@ -438,6 +438,7 @@
     <Compile Include="Search\FuzzyTermsEnum.cs" />
     <Compile Include="Search\HitQueue.cs" />
     <Compile Include="Search\IndexSearcher.cs" />
+    <Compile Include="Search\ITopTermsRewrite.cs" />
     <Compile Include="Search\LiveFieldValues.cs" />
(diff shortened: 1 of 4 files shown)


synhershko:
Is there a reason to have that marker interface public and not kept private?

While I'd assume there are more places like this, I wonder if we can assume they are all internal and avoid exposing this interface?

laimis:
@synhershko great point, I will adjust, there is no need for it to be public.

laimis:
Done.

PR #131, opened 2015-04-22
Patch for QueryParser to avoid throwing lots of exceptions that slows down the debugger

franckspike:
Patch for QueryParser to avoid throwing lots of exceptions as part of main control flow.
When called very often, it slowed down a lot the
execution time when the visual studio debugger was attached.
This patch enhance the performance of the QueryParser under a debugger.


synhershko:
Thanks. We are not actively maintaining the 3.x branch as we put all our efforts in having a 4.x release soon. I'll be keeping this around for a reference.

paulirwin:
Closing this as we are preparing a 4.8 release and not planning on any changes to 3.0.x at this time.

PR #132, opened 2015-04-26
make sure to invoke appropriate CompareValues via FieldComparator

laimis:
.NET port introduced FieldComparator that generic FieldComparator<T> inherits from. The issue is that the CompareValues it introduced does not properly propagate CompareValues calls to subclasses of FieldComparator<T>.

To see the bug in action, look at the failing tests in TestTopDocsMerge.TestSort_1 or TestTopDocsMerge. TestSort_2 (does not fail always, but can be repro with a few runs). Here is the link on TC: http://teamcity.codebetter.com/viewLog.html?buildId=189195&tab=buildResultsDiv&buildTypeId=LuceneNet_Core#testNameId-8365680837810961892

The call invoked here:

https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Core/Search/TopDocs.cs#L207

Results in execution of:

https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Core/Search/FieldComparator.cs#L318

Instead of let's say specific FieldComparator implementation such as this:

https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Core/Search/FieldComparator.cs#L1002

FieldDoc.Fields array was put back to contain object types to match what Lucene has since IComparable interface becomes not used with the changes.

### DIFF
--- a/src/Lucene.Net.Core/Search/FieldComparator.cs
+++ b/src/Lucene.Net.Core/Search/FieldComparator.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 
 namespace Lucene.Net.Search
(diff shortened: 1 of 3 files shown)


synhershko:
Good catch, will merge now

Can you evaluate if we still need that non-generic class?

laimis:
@synhershko took a look, it seems like we need it because there are places using this without explicit specification of generic type T to match what Lucene is doing (where it declares it as FieldComparator<?>).

PR #135, opened 2015-05-02
Fix "The process cannot access the file '...' because it is being used by another process." issues

laimis:
A good number of tests fail with the error message stating that certain file cannot be accessed. Side by side comparison with what Lucene is doing while using Filemon showed that Lucene.Net version was opening the filestreams with correct FileAccess flag but incorrect FileShare flag. Instead of read it should have been ReadWrite.

The failure would occur when writer was disposed and it called IOUtils fsync. Fsync opens the stream again in FileAccess ReadWrite mode while at the same time there were readers that have the file opened with FileShare set to Read only.

Here is the screenshot of the file handle opened when Lucene is running:

javaversion

And here is the Lucene.Net:

dotnetversion

(hopefully these images come through).

Also was using the TestBufferedIndexInput to confirm the fix and noticed that it was rewritten to use using(..)s but during the rewrite some code was omitted so I rewrote it back to match exactly what Lucene is doing so it is easier to compare.

### DIFF
--- a/src/Lucene.Net.Core/Store/FSDirectory.cs
+++ b/src/Lucene.Net.Core/Store/FSDirectory.cs
@@ -477,7 +477,7 @@ namespace Lucene.Net.Store
             {
                 this.Parent = parent;
                 this.Name = name;
-                File = new FileStream(Path.Combine(parent.directory.FullName, name), FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.Read);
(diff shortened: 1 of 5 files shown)


laimis:
The same problem might exist in MMapDirectory and NIOFSDirectory but I haven't been able to verify them yet. Figured to push this out first, get confirmation on TC in addition to my local tests that look good, and then verify if those two are misbehaving the same way.

laimis:
Ran tests for MMapDirectory / NIOFSDirectory and looks like those open with sharemode set to read / write / delete. Adjusted the code in Lucene.Net to match the same behavior.

synhershko:
Awesome work, thanks. Pushing this now and let's see if TC confirms.

PR #138, opened 2015-05-03
Make sure offsetGap is nullable int

laimis:
When MockAnalyzer is used, in certain cases the index could have incorrect values written to it. GetOffsetGap does a comparison of OffsetGap_Renamed to null which is always false if OffsetGap_Renamed is not nullable int. Corrected the logic.

To see the bug in action, refer to the following failing test:

http://teamcity.codebetter.com/viewLog.html?buildId=190073&tab=buildResultsDiv&buildTypeId=LuceneNet_Core#testNameId778192460553334183

### DIFF
--- a/src/Lucene.Net.TestFramework/Analysis/MockAnalyzer.cs
+++ b/src/Lucene.Net.TestFramework/Analysis/MockAnalyzer.cs
@@ -46,7 +46,7 @@ namespace Lucene.Net.Analysis
         private readonly bool LowerCase;
         private readonly CharacterRunAutomaton Filter;
         private int PositionIncrementGap_Renamed;
-        private int OffsetGap_Renamed;
+        private int? OffsetGap_Renamed;
         private readonly Random Random;
         private IDictionary<string, int?> PreviousMappings = new Dictionary<string, int?>();
         private bool EnableChecks_Renamed = true;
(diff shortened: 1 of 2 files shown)


laimis:
Hold off on merging this one just yet, the failing test got fixed with this, but now new failures are appearing in the related code. Investigating...

laimis:
It looks like those other tests were failing because they were modified in Lucene.Net to work around the bug, ie tests were adjusted to pass while leaving the bug in place.

Corrected the tests to match Lucene. Here is the Lucene tests, starting with the first line that was made to match:

https://github.com/apache/lucene-solr/blob/lucene_solr_4_8_0/lucene/core/src/test/org/apache/lucene/index/TestTermVectorsWriter.java#L129

synhershko:
Good catch

2 things come to mind:

1. `OffsetGap_Renamed` and such other `_Renamed` instances need to be, well, renamed and have that suffix dropped. It's the converter doings, and it just makes the code base looks bad

2. I'm sure there are more such places, e.g. nullables that were ported as value types and somewhere in the code are compared to null, indicating they should indeed be nullables. R# should highlight those comparisons, but we need to search for them and inspect one by one.

laimis:
@synhershko agreed on _Renamed. I have been reluctant to touch those so as to keep the changes focused on the bug fix bug can start fixing those as I go.

Re: nullables, I looked at the compiler warnings for more of these spots and there were 3-4 more. Trying to nail down the remaining failing tests and then can grab those too if no one gets to it first. They did seem to be in the spots that are not very likely to be triggered.

PR #142, opened 2015-05-08
Debug.Assert with side effects fix

laimis:
TestDocumentsWriterStallControl fails in Release mode but not in Debug builds. Looks like Lucene is performing logic when calling assert, which translated to Debug.Assert in Lucene.Net. Debug.Assert is removed at the compile time if compiled in release mode.

Took out the logic from the Debug.Assert.

Here is the failing test that this fixes:
http://teamcity.codebetter.com/viewLog.html?buildId=190748&tab=buildResultsDiv&buildTypeId=LuceneNet_Core#testNameId-5834889155160949386

There are more places in the code base that suffer from this, I believe. Looking for them.

### DIFF
--- a/src/Lucene.Net.Core/Analysis/TokenStream.cs
+++ b/src/Lucene.Net.Core/Analysis/TokenStream.cs
@@ -91,7 +91,7 @@ namespace Lucene.Net.Analysis
         /// </summary>
         protected TokenStream()
         {
-            Debug.Assert(AssertFinal());
+            AssertFinal();
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


laimis:
@synhershko I am pushing the debug.asserts with side effects into this PR and then will wait for you to discuss what to do here (merge the changes, come up with some other "assert" method, etc).

laimis:
What I have here should be all that show up in unit test failures. Will do another pass in couple days to see if there are more spots like these in the code base.

synhershko:
All looks good, I don't see anything to discuss here really. Will merge now.

PR #145, opened 2015-05-16
use Environment.TickCount instead DateTime.Now.Millisecond

laimis:
During the port System.currentTimeMillis was translated to System.DateTime.Now.Millisecond which are not equivalent. This introduced long running tests that can be seen in several places in Lucene.Net (e.g. TestTransactions.TestTransactions_Mem). When used to determine how long to run a certain operation, in certain cases this bug was equivalent of running the test infinitely (or a very very long time). See changes to TestTransactions.cs that illustrate the problem well.


laimis:
Also removed LongRunningTest or Ignore attributes on the tests that used to fail because of this before.

synhershko:
Can you rebase please? :)

laimis:
@synhershko can you check now? If it is too messy I can reopen a different PR, ot sure if I did this right

src/Lucene.Net.Core/Util/OfflineSorter.cs
```diff
@@ -34,486 +34,6 @@ namespace Lucene.Net.Util
     /// </summary>
     public sealed class OfflineSorter
     {
-//        private bool InstanceFieldsInitialized = false;
```
synhershko: This is being used by various sub-projects, we can't just delete it (we need to finish porting it, basically)

laimis: Gotcha, added the code back

synhershko:
Good catch, I'll merge now.

For the record, `DateTime.UtcNow.Ticks` is a closer equivalent to `System.currentTimeMillis` but for the purpose of this change your change is good enough. Thanks.

laimis:
@synhershko oh yeah, you are right, I will keep that in mind.

PR #153, opened 2015-08-23
Complete port of Lucene.Net.Join

jpsullivan:
The port of the lucene-solr Join project is now complete. Tests are also implemented and passing.


src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
```diff
@@ -607,7 +607,7 @@ public virtual void TearDown()
         /// </summary>
         public static Random Random()
         {
-            return _random ?? (_random = new Random( /* LUCENENET TODO seed */));
+            return _random ?? (_random = new Random(1));
```
synhershko: remove the 1 seed

jpsullivan: Whoops! Missed that one. Will fix.

PR #154, opened 2015-10-06
Fix OpenBitSet.Union and .Xor methods.

uppi:
EnsureCapacityWords method changes this.Wlen to newLen,
so we have to save this.Wlen value before calling it.

Bugged version never copied the tail of the bigger array.
It worked correctly, but inefficiently.


src/Lucene.Net.Tests/core/Util/TestOpenBitSet.cs
```diff
@@ -514,5 +514,24 @@ public virtual void TestEnsureCapacity()
             bits.FastSet(bit - 1);
             Assert.IsTrue(bits.FastGet(bit - 1));
         }
+
+        [Test]
+        public virtual void TestXorWithDifferentCapacity()
```
synhershko: can you please mark this / comment that this is Lucene.NET specific (so we don't accidently delete it in future ports)?

PR #155, opened 2015-12-08
Analysis.Common tests

laimis:
This is a PR for just a straight up converted tests from https://github.com/apache/lucene-solr/tree/lucene_solr_4_8/lucene/analysis/common/src/test/org/apache/lucene. The idea is to include the converted source files but keep the project file empty. Then going forward we take tests one by one, include it, get it to compile, do any fixes needed to get them to pass and open PR for each such pass.

This gives a good way for anyone to be able to open the project file and see how many tests have been converted and how many still need work just by looking at the include / exclude shading in Visual Studio.


synhershko:
Looks good - let's do this!

PR #159, opened 2016-01-08
Port of CharTokenizer, other code it depends on and corresponding tests

laimis:


src/Lucene.Net.Analysis.Common/Analysis/Core/LetterTokenizer.cs
```diff
@@ -75,10 +75,10 @@ public LetterTokenizer(LuceneVersion matchVersion, AttributeSource.AttributeFact
         /// <summary>
         /// Collects only characters which satisfy
         /// <seealso cref="Character#isLetter(int)"/>.
-        /// </summary>	  
-        protected override bool IsTokenChar(char c)
+        /// </summary>
+        protected override bool IsTokenChar(int c)
```
synhershko: IIRC I tried keeping it a char for .NET-ification reasons, but I can see how UTF32 can force us back to int

PR #160, opened 2016-01-27
Adding two Grouping classes and supporting Arrays.cs methods.

WakeflyCBass:
Added AbstractDistinctValuesCollector and TermDistinctValuesCollector, and modified Arrays object to have some helper methods Java has in Arrays. 
I added an interface to allow AbstractDistinctValuesCollector to only have one type parameter, without it we'd have had to say "AbstractDistinctValuesCollector<T,U> : Collector where T : AbstractDistinctValuesCollector.GroupCount<U>", which is redundant but necessary in C#.


synhershko:
Looks good, thanks!

PR #161, opened 2016-03-10
Use a single Any CPU build configuration

jozefizso:
I've updated the solution and Lucene.Net project to use only the `Any CPU` build configuration.

The x86 build configuration was set only in the Lucene.Net project and produced the AnyCPU binary anyway.

This removes the x86 and Mixed solution configurations that are no longer needed.


synhershko:
Thanks!

PR #162, opened 2016-03-11
Fixed exception messages in NamedSPILoader and Codec classes

jozefizso:
NamedSPILoader and Codec classes used original Java based texts in exception messages that referenced JAR files and Java method names. This commit rewords these messages to be more .NET friendly.

It also fixes bug in the exception message in NamedSPILoader.Lookup() method that did not list the names of available services properly.

I refactored unit tests for the NamedSPILoader class to be more reliable. Original tests were moved to TestCodec class.


synhershko:
Looks good, thanks!

PR #165, opened 2016-04-05
Using CS#5 features to not break CI build

conniey:
Adding language version to use C# 5 features.


conniey:
@EMaher

conniey:
Sorry. I did this on the wrong branch.

PR #166, opened 2016-04-07
Remove breaking projects from Portable solution

conniey:
@EMaher


conniey:
Created PR on wrong branch.

PR #167, opened 2016-04-20
replace InstanceNotFoundException to allow building on Mono

devheroo:
It will be great to keep LuceneNet cross platform and compatible with Mono.
This is just a simple fix to make the build pass on Mono\Linux.

### DIFF
--- a/src/Lucene.Net.Facet/Taxonomy/WriterCache/CollisionMap.cs
+++ b/src/Lucene.Net.Facet/Taxonomy/WriterCache/CollisionMap.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Management.Instrumentation;
+
 
 namespace Lucene.Net.Facet.Taxonomy.WriterCache
 {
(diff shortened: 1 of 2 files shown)


src/Lucene.Net.Facet/Taxonomy/WriterCache/CollisionMap.cs
```diff
@@ -261,7 +261,7 @@ public Entry Next()
                 Entry e = this.next_Renamed;
                 if (e == null)
                 {
-                    throw new InstanceNotFoundException();
+					throw new Exception("InstanceNotFoundException");
```
synhershko: Use `InvalidOperationException` or whatever Exception .NET IEnumerator throws on `Next` that fails

synhershko:
Thanks! I've made a comment on the Exception type used

devheroo:
Thanks for the tip , I made the change regarding the InvalidOperationException and also I had to rename a file path inside Lucene.Net.Codecs.Tests.csproj because files names are case sensitive in Linux and build wont pass without making this change.

synhershko:
Thanks, merged. On your next PR, please work on a branch, it is an ill practice in git to merge from someone else's master.

devheroo:
Thanks, that make sense, I will follow your suggestion next time :)

PR #168, opened 2016-04-22
Removing Apache.NMS dependency.

EMaher:
Removing dependency on Apache.NMS.  This is needed to build Lucene.Net on .NET Core.
-Replaced AtomicReference with use of Interlocked.Exchange.
-Replaced CountDownLatch with CountdownEvent.

### DIFF
--- a/src/Lucene.Net.Core/Support/AtomicObject.cs
+++ b/src/Lucene.Net.Core/Support/AtomicObject.cs
@@ -1,0 +1,39 @@
+﻿using System;
+using System.Threading;
+
+namespace Lucene.Net.Support
+{
+    public class AtomicObject<T> where T : class
+    {
+        private T _value;
+        private ReaderWriterLockSlim _lock = new ReaderWriterLockSlim();
+
+        public AtomicObject() : this(default(T))
+        { }
+
+        public AtomicObject(T initial)
+        {
+            _value = initial;
+        }
+
+        public T Value
+        {
+            get
+            {
+                try
+                {
+                    _lock.EnterReadLock();
+                    return _value;
+                }
+                finally
+                {
+                    _lock.ExitReadLock();
+                }
+            }
+            set
+            {
+                Interlocked.Exchange(ref _value, value);
+            }
+        }
+    }
+}
--- a/src/Lucene.Net.Tests/core/Index/TestIndexWriterWithThreads.cs
+++ b/src/Lucene.Net.Tests/core/Index/TestIndexWriterWithThreads.cs
@@ -1,4 +1,3 @@
-using Apache.NMS.Util;
 using System;
 using System.Diagnostics;
 using System.Threading;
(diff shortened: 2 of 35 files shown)


src/Lucene.Net.Tests/core/Index/TestIndexWriterWithThreads.cs
```diff
@@ -657,7 +655,7 @@ public virtual void TestRollbackAndCommitWithThreads()
             }
 
             Assert.IsTrue(!failed.Get());
-            writerRef.Value.Dispose();
+            writerRef.Dispose();
```
synhershko: this will try to dispose of the local reference, not the reference that may have been updated within `ThreadAnonymousInnerClassHelper`. This is why writerRef was originally boxed - so if the reference changes we dispose of the correct instance. Can you compare with the Java version to verify correctness?

src/Lucene.Net.Tests/core/Index/TestIndexWriterWithThreads.cs
```diff
@@ -726,15 +724,15 @@ public override void Run()
...
-                                    WriterRef.Value.Commit();
+                                    WriterRef.Commit();
                                 }
-                                catch (AlreadyClosedException ace)
+                                catch (AlreadyClosedException)
                                 {
                                     // ok
                                 }
-                                catch (System.NullReferenceException npe)
+                                catch (NullReferenceException)
```
synhershko: why not just add a null check? why is this ok?

synhershko:
Looks great, thanks!

I add a few comments around `TestIndexWriterWithThreads`, it's not a trivial test and it's worth checking on our CI whether or not it fails before and after your changes. If it fails, it probably fails sporadically.

There is a high risk of introducing thread safety issues here, so I'd want us to cover our bases.

Let's compare the code in this test to the Java version and make sure we are preserving the same behavior. Once we know we do, let's run the tests many times, and strengthen them where possible, to make sure this is air tight.

WDYT?

src/Lucene.Net.Tests/core/Index/TestIndexWriterWithThreads.cs
conniey: @synhershko is correct. You can box this value with a helper class like:

```csharp
class Helper<T> where T : class
    {
        private T _value;

        public Helper() : this(default(T))
        { }

        public Helper(T initial)
        {
            _value = initial;
        }

        public T Value
        {
            get
            {
                lock (_value)
                {
                    return _value;
                }
            }
            set
            {
                Interlocked.Exchange(ref _value, value);
            }
        }
    }
```

eladmarg: @conniey great approach,
I would even use the ReaderWriterLockSlim in order to reduce the performance penalty of locking. (in this way for reads only 1 thread can access, and I'm not sure this is a must)

https://msdn.microsoft.com/en-us/library/system.threading.readerwriterlockslim.aspx

EMaher: Thanks @conniey and @eladmarg.  I'll make that change and update the PR on Monday.

src/Lucene.Net.Tests/core/Index/TestIndexWriterWithThreads.cs
EMaher: I just removed the unnecessary 'System' and removed the unused variable declaration.  I'm not sure why it is okay, you'd have to ask the original test writer.

EMaher:
OK.  Suggested changes have been made.  I also verified that passing tests continue to pass, especially the ones that used to use AtomicReference.

synhershko:
Thanks!

src/Lucene.Net.Core/Support/AtomicObject.cs
```diff
@@ -0,0 +1,39 @@
...
+                    return _value;
+                }
+                finally
+                {
+                    _lock.ExitReadLock();
+                }
+            }
+            set
+            {
+                Interlocked.Exchange(ref _value, value);
```
eladmarg: Great job :+1: 
just for safety, I would add 
`
            try
            {
                _lock.EnterWriteLock();
                Interlocked.Exchange(ref _value, value);
            }
            finally
            {
                _lock.ExitWriteLock();
            }`

PR #169, opened 2016-04-26
Fixes the Nuget Nunit reference for Lucene.Net.Tests.Expressions

Shazwazza:
Super simple PR just to fix the nuget package reference for  Lucene.Net.Tests.Expressions for Nunit since it wasn't using a real Nuget reference.


synhershko:
Thanks!

PR #171, opened 2016-05-03
Adding a MergeScheduler that uses Tasks

conniey:
- Adds a MergeScheduler that uses Tasks (default is still ConcurrentMergeScheduler)
- Adds tests to test TaskMergeScheduler

- Added the NUnit Project Guid that was continuously added by Visual Studio

### DIFF
--- a/src/Lucene.Net.Core/Index/IConcurrentMergeScheduler.cs
+++ b/src/Lucene.Net.Core/Index/IConcurrentMergeScheduler.cs
@@ -1,0 +1,15 @@
+﻿namespace Lucene.Net.Index
+{
+    public interface IConcurrentMergeScheduler : IMergeScheduler
+    {
+        int MaxThreadCount { get; }
+        int MaxMergeCount { get; }
+        int MergeThreadPriority { get; set; }
+
+        void Sync();
+        void SetMaxMergesAndThreads(int maxMergeCount, int maxThreadCount);
+
+        void SetSuppressExceptions();
+        void ClearSuppressExceptions();
+    }
+}
--- a/src/Lucene.Net.Core/Index/IConcurrentMergeScheduler.cs
+++ b/src/Lucene.Net.Core/Index/IConcurrentMergeScheduler.cs
@@ -1,5 +1,6 @@
 ﻿namespace Lucene.Net.Index
 {
+    // LUCENENET specific
     public interface IConcurrentMergeScheduler : IMergeScheduler
     {
         int MaxThreadCount { get; }
--- a/src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
+++ b/src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
@@ -75,7 +75,7 @@ namespace Lucene.Net.Index
 
         /// <summary>
         /// <seealso cref="MergeScheduler"/> to use for running merges. </summary>
-        protected internal volatile MergeScheduler mergeScheduler;
+        protected internal volatile IMergeScheduler mergeScheduler;
 
         /// <summary>
         /// Timeout when trying to obtain the write lock on init. </summary>
@@ -149,7 +149,9 @@ namespace Lucene.Net.Index
             useCompoundFile = IndexWriterConfig.DEFAULT_USE_COMPOUND_FILE_SYSTEM;
             openMode = OpenMode_e.CREATE_OR_APPEND;
             similarity = IndexSearcher.DefaultSimilarity;
+
             mergeScheduler = new ConcurrentMergeScheduler();
+
             writeLockTimeout = IndexWriterConfig.WRITE_LOCK_TIMEOUT;
             indexingChain = DocumentsWriterPerThread.defaultIndexingChain;
             codec = Codec.Default;
@@ -547,7 +549,7 @@ namespace Lucene.Net.Index
         /// Returns the <seealso cref="MergeScheduler"/> that was set by
         /// <seealso cref="IndexWriterConfig#setMergeScheduler(MergeScheduler)"/>.
         /// </summary>
-        public virtual MergeScheduler MergeScheduler
+        public virtual IMergeScheduler MergeScheduler
         {
             get
             {
--- a/src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
+++ b/src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
@@ -149,9 +149,11 @@ namespace Lucene.Net.Index
             useCompoundFile = IndexWriterConfig.DEFAULT_USE_COMPOUND_FILE_SYSTEM;
             openMode = OpenMode_e.CREATE_OR_APPEND;
             similarity = IndexSearcher.DefaultSimilarity;
-
+#if FEATURE_TASKMERGESCHEDULER
+            mergeScheduler = new TaskMergeScheduler();
+#else
             mergeScheduler = new ConcurrentMergeScheduler();
-
+#endif
             writeLockTimeout = IndexWriterConfig.WRITE_LOCK_TIMEOUT;
             indexingChain = DocumentsWriterPerThread.defaultIndexingChain;
             codec = Codec.Default;
--- a/src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
+++ b/src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
@@ -149,11 +149,7 @@ namespace Lucene.Net.Index
             useCompoundFile = IndexWriterConfig.DEFAULT_USE_COMPOUND_FILE_SYSTEM;
             openMode = OpenMode_e.CREATE_OR_APPEND;
             similarity = IndexSearcher.DefaultSimilarity;
-#if FEATURE_TASKMERGESCHEDULER
-            mergeScheduler = new TaskMergeScheduler();
-#else
             mergeScheduler = new ConcurrentMergeScheduler();
-#endif
             writeLockTimeout = IndexWriterConfig.WRITE_LOCK_TIMEOUT;
             indexingChain = DocumentsWriterPerThread.defaultIndexingChain;
             codec = Codec.Default;
--- a/src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
+++ b/src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
@@ -1,0 +1,665 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+
+namespace Lucene.Net.Index
+{
+    using Lucene.Net.Support;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Util;
+
+    /*
+         * Licensed to the Apache Software Foundation (ASF) under one or more
+         * contributor license agreements.  See the NOTICE file distributed with
+         * this work for additional information regarding copyright ownership.
+         * The ASF licenses this file to You under the Apache License, Version 2.0
+         * (the "License"); you may not use this file except in compliance with
+         * the License.  You may obtain a copy of the License at
+         *
+         *     http://www.apache.org/licenses/LICENSE-2.0
+         *
+         * Unless required by applicable law or agreed to in writing, software
+         * distributed under the License is distributed on an "AS IS" BASIS,
+         * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+         * See the License for the specific language governing permissions and
+         * limitations under the License.
+         */
+
+    using Directory = Lucene.Net.Store.Directory;
+
+    /// <summary>
+    ///  A <seealso cref="MergeScheduler"/> that runs each merge using
+    ///  Tasks on the default TaskScheduler.
+    /// 
+    ///  <p>If more than <seealso cref="#GetMaxMergeCount"/> merges are
+    ///  requested then this class will forcefully throttle the
+    ///  incoming threads by pausing until one more more merges
+    ///  complete.</p>
+    /// </summary>
+    public class TaskMergeScheduler : MergeScheduler, IConcurrentMergeScheduler
+    {
+        public const string COMPONENT_NAME = "CMS";
+
+        private readonly TaskScheduler _taskScheduler = TaskScheduler.Default;
+        private readonly ReaderWriterLockSlim _lock = new ReaderWriterLockSlim();
+        private readonly ManualResetEventSlim _manualResetEvent = new ManualResetEventSlim();
+        /// <summary>
+        /// List of currently active <seealso cref="MergeThread"/>s.</summary>
+        private readonly IList<MergeThread> _mergeThreads = new List<MergeThread>();
+
+        /// <summary>
+        /// How many <seealso cref="MergeThread"/>s have kicked off (this is use
+        ///  to name them).
+        /// </summary>
+        private int _mergeThreadCount;
+
+        /// <summary>
+        /// <seealso cref="Directory"/> that holds the index. </summary>
+        private Directory _directory;
+
+        /// <summary>
+        /// <seealso cref="IndexWriter"/> that owns this instance.
+        /// </summary>
+        private IndexWriter _writer;
+
+        /// <summary>
+        /// Sole constructor, with all settings set to default
+        ///  values.
+        /// </summary>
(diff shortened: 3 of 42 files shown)


synhershko:
Thanks! give me a day or two to review this properly :)

src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
```diff
@@ -149,7 +149,11 @@ internal LiveIndexWriterConfig(Analyzer analyzer, LuceneVersion matchVersion)
             useCompoundFile = IndexWriterConfig.DEFAULT_USE_COMPOUND_FILE_SYSTEM;
             openMode = OpenMode_e.CREATE_OR_APPEND;
             similarity = IndexSearcher.DefaultSimilarity;
+#if FEATURE_TASKMERGESCHEDULER
+            mergeScheduler = new TaskMergeScheduler();
+#else
             mergeScheduler = new ConcurrentMergeScheduler();
+#endif
```
synhershko: -1 for using compilation flags, can we decide which to initialize based on ctor param (with default value) or keep the Java default but allow to override it via the object initializer?

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
```diff
@@ -0,0 +1,665 @@
...
+    public class TaskMergeScheduler : MergeScheduler, IConcurrentMergeScheduler
+    {
+        public const string COMPONENT_NAME = "CMS";
+
+        private readonly TaskScheduler _taskScheduler = TaskScheduler.Default;
+        private readonly ReaderWriterLockSlim _lock = new ReaderWriterLockSlim();
+        private readonly ManualResetEventSlim _manualResetEvent = new ManualResetEventSlim();
+        /// <summary>
+        /// List of currently active <seealso cref="MergeThread"/>s.</summary>
+        private readonly IList<MergeThread> _mergeThreads = new List<MergeThread>();
```
synhershko: Will initializing the list size to _taskScheduler.MaximumConcurrencyLevel help with performance in any way?

src/Lucene.Net.Tests/core/Index/TestTransactions.cs
```diff
@@ -128,13 +134,21 @@ public IndexerThread(TestTransactions outerInstance, object @lock, Directory dir
...
                 // Intentionally use different params so flush/merge
                 // happen @ different times
-                IndexWriter writer2 = new IndexWriter(Dir2, ((IndexWriterConfig)NewIndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random())).SetMaxBufferedDocs(2)).SetMergeScheduler(new ConcurrentMergeScheduler()).SetMergePolicy(NewLogMergePolicy(3)));
-                ((ConcurrentMergeScheduler)writer2.Config.MergeScheduler).SetSuppressExceptions();
+                var config2 = NewIndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random()))
+                                .SetMaxBufferedDocs(2)
+                                .SetMergeScheduler(_scheduler2)
+                                .SetMergePolicy(NewLogMergePolicy(3));
+                IndexWriter writer2 = new IndexWriter(Dir2, config2);
+                ((IConcurrentMergeScheduler)writer2.Config.MergeScheduler).SetSuppressExceptions();
```
synhershko: why not do a randomized shceduler selection?

src/Lucene.Net.Tests/core/TestMergeSchedulerExternal.cs
```diff
@@ -76,7 +76,9 @@ protected override MergeThread GetMergeThread(IndexWriter writer, MergePolicy.On
             {
                 MergeThread thread = new MyMergeThread(this, writer, merge);
                 thread.ThreadPriority = MergeThreadPriority;
+#if !FEATURE_TASKMERGESCHEDULER
                 thread.SetDaemon(true);
+#endif
```
synhershko: why?

src/Lucene.Net.Core/Index/IConcurrentMergeScheduler.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Lucene.Net.Index
+{
+    public interface IConcurrentMergeScheduler : IMergeScheduler
```
synhershko: mark as LuceneNET specific (also other pieces of code that were added / changed not in line with the Java implementation)

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
```diff
@@ -0,0 +1,665 @@
...
+        /// </summary>
+        private void UpdateMergeThreads()
+        {
+            foreach (var merge in _mergeThreads.ToArray())
+            {
+                // Prune any dead threads
+                if (!merge.IsAlive)
+                {
+                    _mergeThreads.Remove(merge);
+                    merge.Dispose();
```
synhershko: You are cloning `_mergeThreads` to iterate a stable copy, but at this point the `merger` element may have been already disposed by another thread. Protective coding via adding a check for the return value of the Remove call or whether `merge` has been disposed already should help guarantee safety.

synhershko:
Looks great, thanks! left a few minor comments

src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
conniey: I agree that adding it as a ctor parameter is better. How would you decide whether to use ConcurrentMergeScheduler or TaskMergeScheduler? Would it be a config value... or?

src/Lucene.Net.Tests/core/Index/TestTransactions.cs
conniey: I wanted to ensure that both schedulers were being tested so I can compare in case there are bugs in the Task scheduler.

src/Lucene.Net.Tests/core/TestMergeSchedulerExternal.cs
conniey: Because Tasks are always run in the background so this method only applies to `ThreadClass`

src/Lucene.Net.Core/Index/IConcurrentMergeScheduler.cs
conniey: How do I mark it as that? Do you have an example? Thanks!

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
conniey: In the case that the thread was already removed, the call would return false, but would not throw an exception. This would still maintain a good state in the collection though.

src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
synhershko: The C# way is to use the object initializer on LiveIndexWriterConfig to set it, so:

new LiveIndexWriterConfig() { MergeScheduler = new TaskMergeScheduler() }

while keeping ConcurrentMergeScheduler the default.

WDYT?

src/Lucene.Net.Core/Index/IConcurrentMergeScheduler.cs
synhershko: A comment `// LUCENENET specific` should do

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
synhershko: yes, but wouldn't merge.Dispose() throw in some cases, if the same method is called concurrently by different threads? shouldn't you be checking if disposed before calling Dispose?

src/Lucene.Net.Tests/core/Index/TestTransactions.cs
synhershko: sounds like it's better to do using a ValueSource in the calling method?

src/Lucene.Net.Tests/core/TestMergeSchedulerExternal.cs
synhershko: Sounds legit. Need to think how to still do that once we remove the compilation condition as per above.

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
conniey: I don't think so. We we are using the [Default TaskScheduler](https://github.com/conniey/lucenenet/blob/conniey/addTaskScheduler/src/Lucene.Net.Core/Index/TaskMergeScheduler.cs#L45) which has its maximum level set to int.MaxValue. (we'd never hit this value though because it has a stair-stepping algorithm that schdules another task when one is taking too long so we'd never spin up to this level.)

src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
conniey: I remember trying to do this before `LiveIndexConfigWriter` is internal and is only instantiated here [IndexWriter.ctor](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Core/Index/IndexWriter.cs#L828)... but there is no setter for MergeScheduler on IndexWriter.

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
conniey: Sorry, it was late when I responded to this. All the usages of UpdateMergeThread() have acquired a write lock before calling this method.
* [TaskMergeScheduler: L217](https://github.com/conniey/lucenenet/blob/conniey/addTaskScheduler/src/Lucene.Net.Core/Index/TaskMergeScheduler.cs#L217)
* [TaskMergeScheduler: L331](https://github.com/conniey/lucenenet/blob/conniey/addTaskScheduler/src/Lucene.Net.Core/Index/TaskMergeScheduler.cs#L331)

src/Lucene.Net.Tests/core/Index/TestTransactions.cs
conniey: Unfortunately, [ValueSource](http://nunit.org/index.php?p=valueSource&r=2.6.4) works at the test method level.

conniey:
@synhershko I updated the PR with comments about Lucene.NET specific code. Is there anything else I could modify? I am still unsure about how to tackle your comment about the compilation directive in LiveIndexWriterConfig.

conniey:
Hey, we're about to finish the migration to .NET Core and I was hoping someone could take a look at this PR. :) We are blocked on getting this in so we can run the unit tests and obtain accurate results for .NETCore.

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
```diff
@@ -0,0 +1,667 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
```
geobmx540: No bearing on this merge in particular - but why are the `usings` split here? Also placement of the License should usually be at the top (I know here you are just following other files - trying to start a discussion about what's the best way to go)

synhershko: we can deal with the license later, there's too many files with that issue

src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
synhershko: I am not sure I am following - you only need a setter for MergeScheduler in this class (`LiveIndexWriterConfig`) and then override it when initialized if wanted. Since IndexWriter is initialized with a LiveIndexWriterConfig, you are all set.

src/Lucene.Net.Tests/core/Index/TestIndexWriterConfig.cs
```diff
@@ -307,7 +315,11 @@ public virtual void TestInvalidValues()
             }
 
             // Test MergeScheduler
+#if FEATURE_TASKMERGESCHEDULER
+            Assert.AreEqual(typeof(TaskMergeScheduler), conf.MergeScheduler.GetType());
+#else
             Assert.AreEqual(typeof(ConcurrentMergeScheduler), conf.MergeScheduler.GetType());
+#endif
```
synhershko: remove compilation flags

src/Lucene.Net.Core/Index/LiveIndexWriterConfig.cs
conniey: I removed all of the compiler preprocessors. I see that there is a setter in [IndexWriterConfig L325](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Core/Index/IndexWriterConfig.cs#L325).

src/Lucene.Net.Tests/core/Index/TestIndexWriterConfig.cs
conniey: Fixed

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
conniey: I moved all the usings into one location and the license at the top for this file.

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
```diff
@@ -0,0 +1,664 @@
...
+                _exceptionHandler = exceptionHandler;
+            }
+
+            private bool IsLoggingEnabled
+            {
+                get
+                {
+                    return _logger != null && _logger.IsEnabled(COMPONENT_NAME);
+                }
+            }
```
synhershko: what's the difference between this and `Verbose`?

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
```diff
@@ -0,0 +1,664 @@
...
+        ///
+        /// <pre class="prettyprint">
+        /// if (verbose()) {
+        ///   message(&quot;your message&quot;);
+        /// }
+        /// </pre>
+        /// </summary>
+        protected internal bool Verbose()
+        {
+            return _writer != null && _writer.infoStream.IsEnabled(COMPONENT_NAME);
```
synhershko: cache return value (or calculate on init / assignment of `_writer`)

synhershko:
Looks great, thanks for amazing work here.

I left a few more minor comments, other than that LGTM and I'll merge soon

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
conniey: IsLoggingEnabled is only seen in the MergeThread class and Verbose() lives in TaskMergeScheduler. 

I was trying to decrease the amount of entanglement between the two classes by only passing in what was needed by MergeThread, rather than passing in a reference to the outer instance.  I can pass in a `bool isLoggerEnabled` in the constructor instead.

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
conniey: The reason I kept it the same is because inside [the Merge method](https://github.com/conniey/lucenenet/blob/2c8e4a9e889e94cc133eeaba7a63765d13e96c99/src/Lucene.Net.Core/Index/TaskMergeScheduler.cs#L218) sets the _writer, so it could have changed in between. I duplicated this behaviour from [ConcurrentMergeScheduler.cs:L405](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Core/Index/ConcurrentMergeScheduler.cs#L405)

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
synhershko: passing a config var like you suggest sound more elegant, yes

src/Lucene.Net.Core/Index/TaskMergeScheduler.cs
synhershko: sounds legit

conniey:
Hey @synhershko , I fixed the config issue you mentioned. Please let me know if there is anything else that needs to be changed. :)

synhershko:
Merged, thanks!

PR #172, opened 2016-07-15
Removing/limiting uses of static variables and methods in Testcases

conniey:
This PR would allow an easier transition to using xUnit.  xUnit by default, runs its tests in parallel and asynchronously.  The current use of mutable static fields limits in test cases the ability to utilise this.

* Changing existing static variables to static readonly so we are explicit that these variables are set once and do not change.
* Change `OLD_FORMAT_IMPERSONATION_IS_ACTIVE` to non-static field since other tests share and use this variable
* Change `ClassEnvRule` to non-static since every test class generates its own rule during test
* Change dependent static methods in LuceneTestCase to non-static because their state depended on the state of these non-static variables

### DIFF
--- a/src/Lucene.Net.TestFramework/Index/RandomIndexWriter.cs
+++ b/src/Lucene.Net.TestFramework/Index/RandomIndexWriter.cs
@@ -85,24 +85,24 @@ namespace Lucene.Net.Index
 
         /// <summary>
         /// create a RandomIndexWriter with a random config: Uses TEST_VERSION_CURRENT and MockAnalyzer </summary>
-        public RandomIndexWriter(Random r, Directory dir)
-            : this(r, dir, LuceneTestCase.NewIndexWriterConfig(r, LuceneTestCase.TEST_VERSION_CURRENT, new MockAnalyzer(r)))
-        {
-        }
+        //public RandomIndexWriter(Random r, Directory dir)
+        //    : this(r, dir, LuceneTestCase.NewIndexWriterConfig(r, LuceneTestCase.TEST_VERSION_CURRENT, new MockAnalyzer(r)))
+        //{
+        //}
 
         /// <summary>
         /// create a RandomIndexWriter with a random config: Uses TEST_VERSION_CURRENT </summary>
-        public RandomIndexWriter(Random r, Directory dir, Analyzer a)
-            : this(r, dir, LuceneTestCase.NewIndexWriterConfig(r, LuceneTestCase.TEST_VERSION_CURRENT, a))
-        {
-        }
+        //public RandomIndexWriter(Random r, Directory dir, Analyzer a)
+        //    : this(r, dir, LuceneTestCase.NewIndexWriterConfig(r, LuceneTestCase.TEST_VERSION_CURRENT, a))
+        //{
+        //}
 
         /// <summary>
         /// create a RandomIndexWriter with a random config </summary>
-        public RandomIndexWriter(Random r, Directory dir, LuceneVersion v, Analyzer a)
-            : this(r, dir, LuceneTestCase.NewIndexWriterConfig(r, v, a))
-        {
-        }
+        //public RandomIndexWriter(Random r, Directory dir, LuceneVersion v, Analyzer a)
+        //    : this(r, dir, LuceneTestCase.NewIndexWriterConfig(r, v, a))
+        //{
+        //}
 
         /// <summary>
         /// create a RandomIndexWriter with the provided config </summary>
--- a/src/Lucene.Net.TestFramework/Search/QueryUtils.cs
+++ b/src/Lucene.Net.TestFramework/Search/QueryUtils.cs
@@ -122,25 +122,25 @@ namespace Lucene.Net.Search
         /// <seealso cref= #checkSkipTo </seealso>
         /// <seealso cref= #checkExplanations </seealso>
         /// <seealso cref= #checkEqual </seealso>
-        public static void Check(Random random, Query q1, IndexSearcher s)
+        public static void Check(Random random, Query q1, IndexSearcher s, Func<IndexReader, bool, IndexSearcher> NewSearcher)
         {
-            Check(random, q1, s, true);
+            Check(random, q1, s, true, NewSearcher);
         }
 
-        public static void Check(Random random, Query q1, IndexSearcher s, bool wrap)
+        public static void Check(Random random, Query q1, IndexSearcher s, bool wrap, Func<IndexReader, bool, IndexSearcher> NewSearcher)
         {
             try
             {
                 Check(q1);
                 if (s != null)
                 {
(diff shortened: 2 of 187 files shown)


src/Lucene.Net.TestFramework/Search/QueryUtils.cs
```diff
@@ -442,15 +445,17 @@ private class CollectorAnonymousInnerClassHelper2 : Collector
...
-            public CollectorAnonymousInnerClassHelper2(Query q, IndexSearcher s, float maxDiff, int[] lastDoc, AtomicReader[] lastReader, IList<AtomicReaderContext> context)
+            public CollectorAnonymousInnerClassHelper2(Query q, IndexSearcher s, float maxDiff, int[] lastDoc, AtomicReader[] lastReader, IList<AtomicReaderContext> context, Func<IndexReader, bool, IndexSearcher> newSearcher)
             {
                 this.q = q;
                 this.s = s;
                 this.MaxDiff = maxDiff;
                 this.LastDoc = lastDoc;
                 this.LastReader = lastReader;
                 this.Context = context;
+                NewSearcher = newSearcher;
```
synhershko: NewSearcherFunc would be a more appropriate name I think?

src/Lucene.Net.Tests.Join/Lucene.Net.Tests.Join.csproj
```diff
@@ -31,7 +31,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="nunit.framework, Version=2.6.3.13283, Culture=neutral, PublicKeyToken=96d09a1eb7f44a77, processorArchitecture=MSIL">
-      <HintPath>..\packages\NUnit.2.6.3\lib\nunit.framework.dll</HintPath>
+      <HintPath>..\..\packages\NUnit.2.6.3\lib\nunit.framework.dll</HintPath>
```
synhershko: can you confirm this was intentional?

src/Lucene.Net.Tests.Queries/TestCustomScoreQuery.cs
```diff
@@ -334,11 +334,11 @@ private void VerifyResults(float boost, IndexSearcher s, IDictionary<int, float>
             assertEquals("queries should have same #hits", h1.Count, h4CustomAdd.Count);
             assertEquals("queries should have same #hits", h1.Count, h5CustomMulAdd.Count);
 
-            QueryUtils.Check(Random(), q1, s, Rarely());
-            QueryUtils.Check(Random(), q2, s, Rarely());
-            QueryUtils.Check(Random(), q3, s, Rarely());
-            QueryUtils.Check(Random(), q4, s, Rarely());
-            QueryUtils.Check(Random(), q5, s, Rarely());
+            QueryUtils.Check(Random(), q1, s, Rarely(), NewSearcher);
+            QueryUtils.Check(Random(), q1, s, Rarely(), NewSearcher);
```
synhershko: q2

src/Lucene.Net.Tests/core/Index/TestBinaryDocValuesUpdates.cs
```diff
@@ -1109,25 +1109,55 @@ public virtual void TestUpdateBinaryDVFieldWithSameNameAsPostingField()
...
-            OLD_FORMAT_IMPERSONATION_IS_ACTIVE = true;
             IndexWriter writer = new IndexWriter(dir, conf);
             Document doc = new Document();
             doc.Add(new StringField("id", "doc", Store.NO));
             doc.Add(new BinaryDocValuesField("f", ToBytes(5L)));
             writer.AddDocument(doc);
             writer.Dispose();
+            dir.Dispose();
+
+            OLD_FORMAT_IMPERSONATION_IS_ACTIVE = false;
```
synhershko: why do you set this here? If I read the original code correctly, this isn't needed

src/Lucene.Net.TestFramework/Index/RandomIndexWriter.cs
```diff
@@ -85,24 +85,24 @@ public static IndexWriter MockIndexWriter(Directory dir, IndexWriterConfig conf,
...
         /// <summary>
         /// create a RandomIndexWriter with a random config </summary>
-        public RandomIndexWriter(Random r, Directory dir, LuceneVersion v, Analyzer a)
-            : this(r, dir, LuceneTestCase.NewIndexWriterConfig(r, v, a))
-        {
-        }
+        //public RandomIndexWriter(Random r, Directory dir, LuceneVersion v, Analyzer a)
+        //    : this(r, dir, LuceneTestCase.NewIndexWriterConfig(r, v, a))
+        //{
+        //}
```
synhershko: Why was removing these necessary? can't you just replace the calls to `LuceneTestCase.NewIndexWriterConfig` with just `new IndexWriterConfig` - or make that method threadsafe?

Since this PR makes us diverge from the original codebase, we want to try and minimize the amount of changes this will cause - and need to be maintained or commented properly so future updates / ports don't override it. By leaving the methods here the way they originally were (and fixing the thread safety issues they I assume pose) we can reduce the amount of changes in the PR  significantlly. WDYT?

synhershko:
Thanks @conniey , looks good! I left a few comments - the main idea is to minimize the amount of non-Java compatible changes we make and to document well the ones we have to make.

src/Lucene.Net.TestFramework/Search/QueryUtils.cs
conniey: Fixed

src/Lucene.Net.Tests.Join/Lucene.Net.Tests.Join.csproj
conniey: Yes. It was trying to find the nunit package in the `src` folder before instead of the repository root.

src/Lucene.Net.Tests.Queries/TestCustomScoreQuery.cs
conniey: Thanks! Good catch.

src/Lucene.Net.Tests/core/Index/TestBinaryDocValuesUpdates.cs
conniey: I had forgotten to remove it when I was originally testing this test failure. I removed it now.

src/Lucene.Net.TestFramework/Index/RandomIndexWriter.cs
conniey: Is that sufficient to replace it with a `new IndexWriterConfig`? The original static method does a lot more set-up.... and since I made LuceneTestCase non-static (since it depends on the non-static variable ClassEnvRule), I wasn't sure how to incorporate this call.

conniey:
Closing this PR because I created an improved one.

PR #175, opened 2016-07-28
IndexableField API Consistency Fix

NightOwl888:
Made `Name()`, `FieldType()`, `GetBoost()` and `BinaryValue()` into property getters for consistency with the other getters in the `IndexableField` type. So we have all properties instead of a mixed bag of properties and methods. For example...

```c#
public SerializableField(IndexableField field)
{
	name = field.Name();
	if (field.NumericValue != null)
	{
		fieldsData = field.NumericValue;
	}
	else if (field.StringValue != null)
	{
		fieldsData = field.StringValue;
	}
	else if (field.BinaryValue() != null)
	{
		fieldsData = field.BinaryValue().Bytes;
	}
	else
	{
		throw new NotSupportedException("Doesn't support this field type so far");
	}
}
```

now looks like...

```c#
public SerializableField(IndexableField field)
{
	name = field.Name;
	if (field.NumericValue != null)
	{
		fieldsData = field.NumericValue;
	}
	else if (field.StringValue != null)
	{
		fieldsData = field.StringValue;
	}
	else if (field.BinaryValue != null)
	{
		fieldsData = field.BinaryValue.Bytes;
	}
	else
	{
		throw new NotSupportedException("Doesn't support this field type so far");
	}
}
```

### DIFF
--- a/src/Lucene.Net.Classification/Utils/DatasetSplitter.cs
+++ b/src/Lucene.Net.Classification/Utils/DatasetSplitter.cs
@@ -97,19 +97,19 @@ namespace Lucene.Net.Classification.Utils
                         {
                             if (storableField.ReaderValue != null)
                             {
-                                doc.Add(new Field(storableField.Name(), storableField.ReaderValue, ft));
+                                doc.Add(new Field(storableField.Name, storableField.ReaderValue, ft));
                             }
                             else if (storableField.BinaryValue() != null)
                             {
-                                doc.Add(new Field(storableField.Name(), storableField.BinaryValue(), ft));
+                                doc.Add(new Field(storableField.Name, storableField.BinaryValue(), ft));
                             }
                             else if (storableField.StringValue != null)
                             {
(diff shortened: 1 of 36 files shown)


synhershko:
I *think* this makes sense, let me play with this a bit and see. Hopefully other community members can contribute to the discussion too.

synhershko:
Looks good, let's work on getting this in. This PR currently breaks sub-projects, as the API change hasn't been propogated there and to their tests. Can you run the entire solution and make sure this API change is brought all over?

NightOwl888:
I thought I got everything, but I will take another look.

NightOwl888:
Ok, I did a search of all of the files in this project for these method names. I only found one other file that referenced any of these fields, and it was in a project that (apparently) doesn't compile. But, I updated the references anyway.

If there is something else you are referring to when you say "sub-projects", please help me out by giving me an example. Are there other repositories that depend on this one?

synhershko:
Looks good, thanks!

PR #176, opened 2016-08-02
Ported QueryParser

NightOwl888:
I have ported much of the QueryParser + the tests.

Current Status
====

Most of (about 2/3) of the tests are passing. The good news is that the Classic.QueryParser is passing on all tests (except for one that I was unable to translate).

Not Yet Ported
----

1. The entire QueryParser.Flexible namespace.
2. The [QueryParser.Surround.Query.SrndTruncQuery.VisitMatchingTerms](https://github.com/NightOwl888/lucenenet/blob/queryparser/src/Lucene.Net.QueryParser/Surround/Query/SrndTruncQuery.cs#L87) method.
3. The [TestDateRange](https://github.com/NightOwl888/lucenenet/blob/queryparser/src/Lucene.Net.Tests.QueryParser/Util/QueryParserTestBase.cs#L751) test for the classic QueryParser.

Points of Interest
---

1. The [Locale and Timezone](https://github.com/NightOwl888/lucenenet/blob/queryparser/src/Lucene.Net.QueryParser/Flexible/Standard/CommonQueryParserConfiguration.cs#L77-L85) features of the QueryParser seem a bit incompatible with .NET. The TimeZone property is not actually hooked to anything since there doesn't seem to be a sensible way to use it in .NET. And in .NET we normally set the culture on the current thread or pass it to a method for a one-off. It doesn't make a lot of sense to make it a member of QueryParser, since a single parser instance could be used on different cultures just by switching the culture of the current thread. Thoughts?
2. I refactored the [QueryParserConstants](https://github.com/NightOwl888/lucenenet/blob/queryparser/src/Lucene.Net.QueryParser/Classic/QueryParserConstants.cs#L20) a bit since it is not possible to add constants to an interface as was done in Java. I weighed the options and ended up splitting them into separate static classes, since there doesn't seem to be a way to reference them in .NET from multiple classes (at least not without a base class) without putting an identifier before the member (`Something.Constant`). I went with `RegexpToken` and `LexicalToken` based on what was in the comments, but feel free to rename them or suggest a different approach.
3. I think that it would be best to use a `[Flags]` enum in .NET for the [`SimpleQueryParser` constants](https://github.com/NightOwl888/lucenenet/blob/queryparser/src/Lucene.Net.QueryParser/Simple/SimpleQueryParser.cs#L98-L121), since passing these flags to the API as an `int` is not very intuitive. However, currently there are 9 tests failing that all directly involve these constants and I don't want to throw another variable in to the debugging effort. I would be happy to change this as soon as the tests are passing if you agree with this assertion.
4. I ended up adding a couple of flags to `SimpleQueryParser` to [indicate a "not set" state](https://github.com/NightOwl888/lucenenet/blob/queryparser/src/Lucene.Net.QueryParser/Simple/SimpleQueryParser.cs#L529), since it is not possible to set an enum to null in .NET and IMO a nullable variable type should not be part of a public API.
5. I have run into a wall with debugging the 9 failing tests on the `SimpleQueryParser`. The [documentation](https://lucene.apache.org/core/4_7_0/queryparser/org/apache/lucene/queryparser/simple/SimpleQueryParser.html) and [comments](https://github.com/apache/lucene-solr/blob/8fdf89690404c0e65784b2c5477552b9dec58591/lucene/queryparser/src/java/org/apache/lucene/queryparser/simple/SimpleQueryParser.java#L102) both indicate that passing a flag is to *enable* the option. However, the [test names](https://github.com/NightOwl888/lucenenet/blob/queryparser/src/Lucene.Net.Tests.QueryParser/Simple/TestSimpleQueryParser.cs#L603-L696) indicate they are testing to *disable* the feature. As a result the wrong type of query is being returned to pass the tests.
6. I have also run into a wall with the failing tests (which are all of the tests) for the Surround.Parser.QueryParser. For example [this test](https://github.com/NightOwl888/lucenenet/blob/queryparser/src/Lucene.Net.Tests.QueryParser/Surround/Query/Test02Boolean.cs#L128) first correctly identifies the AND token, it then immediately tries to look for a `(` and then blows up because it doesn't exist. I triple checked the code and is doing nothing more than running the initialization code and then doing this check. A little assistance would be appreciated, as I don't have the Java version setup to run to check what it *should* do.


synhershko:
Looks good! will review shortly, thanks!

src/Lucene.Net.Core/Util/ToStringUtils.cs
```diff
@@ -36,12 +37,13 @@ public static string Boost(float boost)
         {
             if (boost != 1.0f)
             {
-                return "^" + Convert.ToString(boost);
+                float boostAsLong = (long)boost;
+                if (boostAsLong == boost)
+                    return "^" + boost.ToString(".0").Replace(CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator, ".");
+                return "^" + boost.ToString().Replace(CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator, ".");
```
synhershko: why do you need the Replace part? won't .ToString(CultureInfo) be enough?

synhershko:
Left one minor comments, looks good to merge otherwise. Thanks for your help!

src/Lucene.Net.Core/Util/ToStringUtils.cs
NightOwl888: I copied it over verbatim from Lucene.Net 3.0.3. The previous code was chopping off the .0 on the end, which was different behavior than what was expected by the tests. This was necessary to make some of the tests pass.

For example, 

```
Test Name:	TestSimple
Test FullName:	Lucene.Net.QueryParser.Classic.TestMultiFieldQueryParser.TestSimple
Test Source:	F:\Projects\lucenenet\src\Lucene.Net.Tests.QueryParser\Classic\TestMultiFieldQueryParser.cs : line 70
Test Outcome:	Failed
Test Duration:	0:00:00.003

Result StackTrace:	
at Lucene.Net.Util.LuceneTestCase.assertEquals(Object expected, Object actual) in F:\Projects\lucenenet\src\Lucene.Net.TestFramework\JavaCompatibility\LuceneTestCase.cs:line 31
at Lucene.Net.QueryParser.Classic.TestMultiFieldQueryParser.TestSimple() in F:\Projects\lucenenet\src\Lucene.Net.Tests.QueryParser\Classic\TestMultiFieldQueryParser.cs:line 87
Result Message:	
Expected string length 33 but was 31. Strings differ at index 16.
  Expected: "((b:one t:one)^2.0) (b:two t:two)"
  But was:  "((b:one t:one)^2) (b:two t:two)"
  ---------------------------^
  ```
And
```
Test Name:	TestRange
Test FullName:	Lucene.Net.QueryParser.Util.QueryParserTestBase.TestRange
Test Source:	F:\Projects\lucenenet\src\Lucene.Net.Tests.QueryParser\Util\QueryParserTestBase.cs : line 657
Test Outcome:	Failed
Test Duration:	0:00:00.003

Result StackTrace:	
at Lucene.Net.Util.LuceneTestCase.fail(String message) in F:\Projects\lucenenet\src\Lucene.Net.TestFramework\JavaCompatibility\LuceneTestCase.cs:line 120
at Lucene.Net.QueryParser.Util.QueryParserTestBase.AssertQueryEquals(String query, Analyzer a, String result) in F:\Projects\lucenenet\src\Lucene.Net.Tests.QueryParser\Util\QueryParserTestBase.cs:line 204
at Lucene.Net.QueryParser.Util.QueryParserTestBase.TestRange() in F:\Projects\lucenenet\src\Lucene.Net.Tests.QueryParser\Util\QueryParserTestBase.cs:line 683
Result Message:	Query /{ a TO z }^2.0/ yielded /{a TO z}^2/, expecting /{a TO z}^2.0/
```

Admittedly the code looks bad, but it worked in the prior version and was exactly what was needed to make these (and other) tests green.

NightOwl888:
Another correction - the Lucene.Net 3.0.3 code had a bug. When passing `0`, it would format the boost as `^.0`, but if you passed `0.123` you would get `^0.123`.

I checked through all of the 4.8 files and there are several unit tests that input `^0.xxx` but nothing boost-related indicates that `^.0` is valid. I think it is safe to assume it should be `^0.0`.

synhershko:
This is usually where we put all .NET specific stuff that is required by Core / subprojects: https://github.com/apache/lucenenet/tree/master/src/Lucene.Net.Core/Support/Compatibility

With test stuff here: https://github.com/apache/lucenenet/tree/master/src/Lucene.Net.TestFramework/JavaCompatibility

NightOwl888:
> This is usually where we put all .NET specific stuff that is required by Core / subprojects: 
> https://github.com/apache/lucenenet/tree/master/src/Lucene.Net.Core/Support/Compatibility

> With test stuff here: > https://github.com/apache/lucenenet/tree/master/src/Lucene.Net.TestFramework/JavaCompatibility

Hmm... I get that there are extension methods and such in there to prevent having to rewrite too much test code, but if I put the test in `Lucene.Net.TestFramework/JavaCompatibility` it will literally be the *only* executable test in `Lucene.Net.TestFramework`. Seems odd, so just double-checking that is where you want it. But then, there are also no tests currently under `Lucene.Net.Tests/core/Support/`.

NightOwl888:
Ok, I have moved the test to Lucene.Net.Tests/core/Support. Apparently there *was* a directory with this name, but it had been removed from the project. Not sure what the reasoning was for this, I suppose this one doesn't belong in the project either if none of those do. But, at least it is now with the rest of the .NET compatibility tests, rather than buried in the Java-ported code.

Feel free to remove it from the project if that is your prerogative. To me it seems like those tests to ensure the parts of Java framework that are ported over are working are a good idea. For example, projects that derive from this one (such as BoboBrowse.Net) might need them, too. But if the implementation is only complete/correct enough to make Lucene.Net work, but not work right for derived libraries, it could mean a lot of copy-paste-edit for those derived libraries.

synhershko:
Hi, sorry it took me so long to review. I wanted to review properly and give proper feedback to complement your hard (and very professional!) work here.

So,

1. This makes sense. This is in-line with our .NETification spirit, and as far as we don't remove important functionality I'd say this makes sense and can be removed altogether - or changed to work in a way that makes sense.

2. This makes sense as well. We took this approach in several other places in the code base. I agree the guideline here should be code readability / usability / fluidity.

3. Using an Enum does make sense, and indeed better to wait for tests to stabilize.

4. This might diverge from other places in our codebase. Why is it public API anyway? I can't really see anyone using it, except from people trying to customize the SimpleQueryParser - which is simplistic in it's nature. I'd say let's make it nullable and either accept it's a public API with not much expected usage, or hide it as non-public API. WDYT?

5. From my familiarity with the Java code base, you pass a flag to enable a feature. I think it's just confusing test names, and then a bug somewhere. @uschindler can you confirm? I vaguely recall hitting this one before, will try to look into it too later this week.

6. I will help here. Give me a few (more!) days.

Awesome work dude, thanks very much. I'll leave the decision up to you whether to get this PR merged or still work on it a bit more until it stabilizes. I will look at items 5 and 6 very soon.

NightOwl888:
Well, everything was going good on this until I got to the Surround parser. If you exclude it like you did with the unfinished parts of Analysis, there are only 12 failing tests. If you also exclude SimpleQueryParser there are only 3 failures. You might consider doing that and rolling those parts of QueryParser and the stable Analysis branch (or perhaps the more complete one) into a pre-release. At least then people will have their old StandardAnalyzer and QueryParser back.

I'll take a look at what can be done about the culture and timezone once it is merged to master.

NightOwl888:
> This might diverge from other places in our codebase. Why is it public API anyway? I can't really see anyone using it, except from people trying to customize the SimpleQueryParser - which is simplistic in it's nature. I'd say let's make it nullable and either accept it's a public API with not much expected usage, or hide it as non-public API. WDYT?

What I meant was, it needs to pass the value through to the public API of another class (I don't recall which offhand). The other class (or classes?) doesn't accept a nullable as part of its public API, so to make it compatible (and do what the Java version does) it would also either need to be nullable, or there would need to be a "guess" at that point in the event it is null. I suppose I could try to figure out what the other class does in the event of a null and move that logic into this class, but that would mean moving business logic from one class into another.

Anyway, can't rule out the possibility that those new flags are part of the problem causing those 9 failing tests. If you have the setup to run in Java to figure out where the execution paths diverge it would be a big help.

uschindler:
Hi @synhershko,

> From my familiarity with the Java code base, you pass a flag to enable a feature. I think it's just confusing test names, and then a bug somewhere. @uschindler can you confirm? I vaguely recall hitting this one before, will try to look into it too later this week.

I can check Lucene Java's code base later - but I have no idea what's going on. YES - we should also pass a EnumSet in Lucene, but that is old code :-)

synhershko:
Merging this before this branch goes too stale. Can you email the dev@ list with all the details here and the TODOs left - and me / others will take it from there?

PR #177, opened 2016-08-04
Limiting uses of static variables and methods in Testcases

conniey:
This is an improvement from PR #172.  The difference is that all the existing constructors are maintained with some additions. 

This PR would allow an easier transition to using xUnit. xUnit by default, runs its tests in parallel and asynchronously. The current use of mutable static fields limits in test cases the ability to utilise this.

* Changing existing static variables to static readonly so we are explicit that these variables are set once and do not change.
* Change OLD_FORMAT_IMPERSONATION_IS_ACTIVE to non-static field since other tests share and use this variable
* Change ClassEnvRule to non-static since every test class generates its own rule during test
* Change dependent static methods in LuceneTestCase to non-static because their state depended on the state of these non-static variables

### DIFF
--- a/src/Lucene.Net.TestFramework/Index/RandomIndexWriter.cs
+++ b/src/Lucene.Net.TestFramework/Index/RandomIndexWriter.cs
@@ -8,21 +8,21 @@ using Lucene.Net.Util;
 namespace Lucene.Net.Index
 {
     /*
-     * Licensed to the Apache Software Foundation (ASF) under one or more
-     * contributor license agreements.  See the NOTICE file distributed with
-     * this work for additional information regarding copyright ownership.
-     * The ASF licenses this file to You under the Apache License, Version 2.0
-     * (the "License"); you may not use this file except in compliance with
-     * the License.  You may obtain a copy of the License at
-     *
-     *     http://www.apache.org/licenses/LICENSE-2.0
-     *
-     * Unless required by applicable law or agreed to in writing, software
-     * distributed under the License is distributed on an "AS IS" BASIS,
-     * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-     * See the License for the specific language governing permissions and
-     * limitations under the License.
-     */
+* Licensed to the Apache Software Foundation (ASF) under one or more
+* contributor license agreements.  See the NOTICE file distributed with
+* this work for additional information regarding copyright ownership.
+* The ASF licenses this file to You under the Apache License, Version 2.0
+* (the "License"); you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
 
     using Analyzer = Lucene.Net.Analysis.Analyzer;
     using BytesRef = Lucene.Net.Util.BytesRef;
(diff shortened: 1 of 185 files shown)


synhershko:
👍

lots of files to review :) will be on it asap. Thanks!

src/Lucene.Net.TestFramework/Index/RandomIndexWriter.cs
```diff
@@ -85,22 +86,22 @@ public static IndexWriter MockIndexWriter(Directory dir, IndexWriterConfig conf,
...
         }
 
         /// <summary>
         /// create a RandomIndexWriter with a random config </summary>
-        public RandomIndexWriter(Random r, Directory dir, LuceneVersion v, Analyzer a)
-            : this(r, dir, LuceneTestCase.NewIndexWriterConfig(r, v, a))
+        public RandomIndexWriter(Random r, Directory dir, LuceneVersion v, Analyzer a, Similarity similarity, TimeZone timezone)
+            : this(r, dir, LuceneTestCase.NewIndexWriterConfig(r, v, a, similarity, timezone))
         {
         }
```
synhershko: can you add a LUCENENET comment saying this is specific to the .NET port + reasoning, so we don't override this change in the future?

src/Lucene.Net.Tests/core/Codecs/Lucene3x/TestLucene3xStoredFieldsFormat.cs
```diff
@@ -35,7 +35,8 @@ protected override Codec Codec
         {
             get
             {
-                return new PreFlexRWCodec();
+                Assert.IsTrue(OLD_FORMAT_IMPERSONATION_IS_ACTIVE, "This should have been set up in the test fixture");
```
synhershko: why are you assuming this to be true here?

src/Lucene.Net.Tests/core/Index/TestIndexWriterCommit.cs
```diff
@@ -732,5 +736,21 @@ public virtual void TestCommitUserData()
 
             dir.Dispose();
         }
+
+        private void AddDoc(IndexWriter writer)
```
synhershko: please put a comment on the original methods so if they change we know to reach back here to update accordingly

src/Lucene.Net.Tests/core/Search/TestFilteredQuery.cs
```diff
@@ -280,7 +280,7 @@ private void TBooleanMUST(bool useRandomAccess)
             bq.Add(query, BooleanClause.Occur.MUST);
             ScoreDoc[] hits = Searcher.Search(bq, null, 1000).ScoreDocs;
             Assert.AreEqual(0, hits.Length);
-            QueryUtils.Check(Random(), query, Searcher);
+            QueryUtils.Check(Random(), Query, Searcher, Similarity);
```
synhershko: `query`?

src/Lucene.Net.Tests/core/Search/TestPhraseQuery.cs
```diff
@@ -255,7 +255,7 @@ public virtual void TestPhraseQueryWithStopAnalyzer()
             query.Add(new Term("field", "words"));
             ScoreDoc[] hits = searcher.Search(query, null, 1000).ScoreDocs;
             Assert.AreEqual(1, hits.Length);
-            QueryUtils.Check(Random(), query, searcher);
+            QueryUtils.Check(Random(), Query, Searcher, Similarity);
```
synhershko: `query, searcher`?

synhershko:
Looks great! Added a few comments

(1) might have found a bug - you didn't change local vars in some occasions so I had commented in some of them, (2) please document all test API changes that diverge from the Java code in the appropriate places, so we don't chase after them in the future or override them by mistake.

conniey:
@synhershko Thank you for looking over the PR!

>(1) might have found a bug
I went through those classes again and fixed all the instances that I found where `query` and `searcher` were incorrectly referenced

>(2) please document all test API changes
I've added `LUCENENET specific` and a reason to all of the APIs that have changed.

Is there anything else that needs to be modified?

synhershko:
Awesome work, thanks!

PR #178, opened 2016-08-07
Ported Core.Codecs.Perfield.TestPerFieldPostingsFormat2

NightOwl888:
Ported the Core.Codecs.Perfield.TestPerFieldPostingsFormat2 tests and removed [Test] attribute from a couple of methods where it didn't belong (methods that are called from other tests).


synhershko:
Looks great! I'm not sure about the dependency of TestFramework in the Codecs subproject - is this like that in the original Java code?

synhershko:
Merged, if that dependency will cause issues we will handle it then

NightOwl888:
Not sure if that dependency was there or not in the original, but in a more recent commit I have removed the JavaCompatibility namespace entirely so that line has also been removed. Just the namespace though, left the directory and all files in place. Its just easier if all of the extension methods "just work" without having to reference another namespace from the test environment.

PR #180, opened 2016-08-21
Misc bugs

NightOwl888:
This removes 15 failing tests and in some other cases fixes the test failure message to show the real logic error instead of an exception.


synhershko:
Awesome, thanks!

PR #183, opened 2016-08-29
Fix test failures and making SPIClassIterator more performant

conniey:
### SPIClassIterator
* Improve loading performance:
   * Filtering out Microsoft/System assemblies
   * Only iterating through distinct assemblies
   * Only adding public classes to the iterator:
       * This fixed test failures where [MockCodec](https://github.com/apache/lucenenet/blob/8ce8e81fb06d43a44974436d79dadd33197bf0e0/src/Lucene.Net.Tests/core/Codecs/Perfield/TestPerFieldPostingsFormat2.cs#L200) (which had a codec name "Lucene46") was being loaded instead of [Lucene46Codec](https://github.com/apache/lucenenet/blob/c6d7bec47383919793721ecf9ca5f6966fb86033/src/Lucene.Net.Core/Codecs/Lucene46/Lucene46Codec.cs#L83) because it was being found first in the iterator
* Consolidated assembly loading code

### Test Failure Fixes
*  Adding a default ctor to the Codecs that were not being found because SPIClassIterator was skipping over them

I tested these changes in Visual Studio 2013 so they should not fail again

### DIFF
--- a/src/Lucene.Net.Core/Util/SPIClassIterator.cs
+++ b/src/Lucene.Net.Core/Util/SPIClassIterator.cs
@@ -79,7 +79,30 @@ namespace Lucene.Net.Util
                     {
                         try
                         {
-                            if (IsInvokableSubclassOf<S>(type))
+                            if (!IsInvokableSubclassOf<S>(type))
+                            {
+                                continue;
+                            }
+
+                            // We are looking for types with a default ctor
+                            // (which is used in NamedSPILoader) or has a single parameter
+                            // of type IDictionary<string, string> (for AnalysisSPILoader)
+                            var matchingCtors = type.GetConstructors().Where(ctor =>
+                            {
+                                var parameters = ctor.GetParameters();
+
+                                switch (parameters.Length)
+                                {
+                                    case 0: // default ctor
+                                        return true;
+                                    case 1:
+                                        return typeof(IDictionary<string, string>).IsAssignableFrom(parameters[0].ParameterType);
+                                    default:
+                                        return false;
+                                }
+                            });
+
+                            if (matchingCtors.Any())
                             {
                                 types.Add(type);
                             }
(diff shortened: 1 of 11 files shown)


NightOwl888:
Thanks. Do note that Analysis.Common also depends on SPIClassIterator, and I had to fix it so it would load classes that don't have a default constructor (which is what the Analysis.Util.AnalysisSPILoader requires). I also modified the NamedSPIClassLoader to filter out the classes without a default constructor.

However, the changes still haven't been merged into master from the [analysis-work branch](https://github.com/apache/lucenenet/tree/analysis-work). Could you please verify these changes work with Analysis.Common and also that they won't cause a merge conflict when it is merged into master?

conniey:
@NightOwl888 I took a look at your branch and merged your `NamedSPILoader` and `SPIClassIterator<S>` changes to avoid you having to do any fix-ups with those classes.

NightOwl888:
@conniey 

Thanks.

It occurred to me that the AnalysisSPILoader only uses classes with a constructor that take a single parameter of type `IDictionary<string, string>`. So, as long as we're optimizing, perhaps the SPIClassIterator could be limited to default constructors and constructors with one parameter of `IDictionary<string, string>`. That should eliminate a lot of types from being considered.

I did a quick search and the NamedSPILoader and AnalysisSPILoader are the only 2 classes that use the SPIClassIterator in lucene, so it should be safe to do this.

conniey:
@NightOwl888 I updated the [iterator to filter](https://github.com/conniey/lucenenet/blob/5611f70455199a97e3b7180300d356612f07a97e/src/Lucene.Net.Core/Util/SPIClassIterator.cs#L90-L108) for types with default constructors or `IDictionary<string,string>`. Is this OK now?

synhershko:
Looks good. Pending @NightOwl888 approval before merging this in. Thanks @conniey !

synhershko:
After merging #173 this branch has merge conflicts, @conniey can you resolve please?

NightOwl888:
@conniey - I ran this against Analysis.Common and it doesn't break any tests. As @synhershko mentioned though, it has some merge conflicts.

The latest commit [here](https://github.com/NightOwl888/lucenenet/tree/analysis-bugz-2) resolves them.

conniey:
@synhershko The merge conflicts are resolved now. :)

PR #185, opened 2016-09-08
Ported Core.Util.Fst tests + Core.Util.Packed.TestPackedInts test

NightOwl888:
This includes 3 missing test files from Core.Util, several bug fixes to make the Core.Util.Fst tests pass (all green), and API refactoring to .NETify the API and make the FST generics act more like they do in Java Lucene.

6 of the Core.Util.Packed.TestPackedInts tests are failing, and 1 additional test needs some additional work to make complete (which I have marked with an Ignore attribute). I looked into the failing tests briefly - they are all failing due to incorrect results from the RAMUsageEstimator, not due to problems with the PackedInts class.

There is [1 FST test](https://github.com/NightOwl888/lucenenet/blob/ef889cebdeaa180d0bb232afd8e5ab3865cd41bc/src/Lucene.Net.Tests/core/Util/Fst/TestFSTs.cs#L312) that I estimate will take around 10-14 hours to run. It was marked with a "Nightly" attribute in Lucene. Perhaps there should be such an attribute in Lucene.Net?


DivenLee:
Thank you for everything you do, and you are looking forward to seeing a new version appear. Not only repair, more progress. Thank you

synhershko:
A `Nightly` attribute would be nice, but our current infrastructure doesn't support it

src/Lucene.Net.Core/AssemblyInfo.cs
```diff
@@ -84,6 +90,6 @@
 //   (*) Delay Signing is an advanced option - see the Microsoft .NET Framework
 //       documentation for more information on this.
 //
-[assembly: AssemblyDelaySign(false)]
-[assembly: AssemblyKeyFile("")]
-[assembly: AssemblyKeyName("")]
+//[assembly: AssemblyDelaySign(false)]
+//[assembly: AssemblyKeyFile("")]
+//[assembly: AssemblyKeyName("")]
```
synhershko: we should be removing strong naming and assembly signing anyway

src/Lucene.Net.TestFramework/Util/fst/FSTTester.cs
```diff
@@ -395,6 +407,24 @@ internal virtual FST<T> DoTest(int prune1, int prune2, bool allowRandomSuffixSha
...
+                foreach (object value in a as IEnumerable)
+                {
+                    iter.MoveNext();
+                    if (!object.Equals(value, iter.Current))
+                    {
+                        return false;
+                    }
+                }
+                return true;
+            }
```
synhershko: looks like code dup?

src/Lucene.Net.TestFramework/Util/fst/FSTTester.cs
```diff
@@ -913,6 +945,73 @@ private void VerifyPruned(int inputMode, FST<T> fst, int prune1, int prune2)
...
+            //            if (cmo2 != null)
+            //            {
+            //                //System.out.println("    clear isLeaf " + inputToString(inputMode, scratch));
+            //                cmo2.IsLeaf = false;
+            //            }
+            //            scratch.Length--;
+            //        }
+            //    }
+            //}
+
```
synhershko: ?

src/Lucene.Net.Core/AssemblyInfo.cs
NightOwl888: I have :). I couldn't get the InternalsVisibleTo to work right without commenting these.

However, it would be better to ultimately have strong named libraries (at least for .NET 4.5) since it is not possible to reference a library that is not strong named from one that is.  If we didn't strong name, anyone who currently is strong naming their assembly won't be able to use Lucene.Net without compiling it themselves with a strong name.

src/Lucene.Net.Tests/core/Util/TestWeakIdentityMap.cs
```diff
@@ -174,12 +188,15 @@ public virtual void TestSimpleHashMap()
             Assert.IsTrue(map.Empty);
         }
 
+        /* LUCENENET TODO: Compile issues
```
synhershko: we probably need to fix those

synhershko:
Very good refactoring, thanks! left a few minor comments, can merge now but will give you a chance to fix them here and not on separate PRs before doing so

src/Lucene.Net.Core/AssemblyInfo.cs
synhershko: That's a whole discussion that we can have on it's own thread, but I'm against strong-naming in general. See http://code972.com/blog/2014/04/68-ditching-strong-naming-for-lucene-net.

src/Lucene.Net.Tests/core/Util/TestWeakIdentityMap.cs
NightOwl888: I realize that. I noticed a few other test files that have comments such as these - most of them seem to be around using threading and thread pools, which are somewhat difficult to port.

This wasn't intended to be a thorough completion of the tests, but it will reduce the amount of effort to be thorough in another pass. Frankly, I was trying to see if there were any unfinished tests that could help me locate a Lucene.Net.Misc test failure due to a bug that is somewhere in the core. I chased after it for a couple of days before trying the shotgun approach - which seemed more productive since these tests need to be completed anyway. Unfortunately, it didn't help me locate the bug. I will point it out and go into details about it when I make the pull request for Suggest/Misc.

src/Lucene.Net.TestFramework/Util/fst/FSTTester.cs
NightOwl888: Code dup from where?

synhershko: `AssertRootArcs()`. Also a similar fix duped in NodeHash. Just nit-picking here, but worth noting.

NightOwl888:
Okay, I added some commits that take care of these issues.

I ported the test and worked out how to make the iterator logic work in the WeakIdentityMap. The idea about mimicking the iterator by using yield return was pretty clever, but unfortunately won't work because it won't stop the outer loop when a stale key is hit. We need an iterator for it to work.

Unfortunately, although the logic works on one thread (you can verify that by commenting [these lines](https://github.com/apache/lucenenet/blob/c67366fd4b383c31971a8ae68b458dec7aee2d7f/src/Lucene.Net.Tests/core/Util/TestWeakIdentityMap.cs#L286-L289)), when you hit it with multiple threads it blows up. Perhaps there is a way to make it work but then again, it is only used in 3 classes in the original Lucene:

1. Core.Util.AttributeSource
2. Core.Util.VirtualMethod
3. Core.Util.ByteBufferIndexInput

In AttributeSource it was replaced with WeakDictionary. Perhaps it could just be replaced in all of them and eliminated rather than spending more time trying to fix it.

synhershko:
At some point I looked at WeakDictionary as a replacement but didn't go forward with it because of some behavior mismatch. Can't really remember now exactly what it was. If you are certain that's the way to go I'm okay with it, otherwise let's keep a close eye on the beavior of that code...

PR #186, opened 2016-09-19
Ported Misc (mostly), Suggest, and Memory

NightOwl888:
This pull request contains Lucene.Net.Misc, Lucene.Net.Suggest, and Lucene.Net.Memory. There is also some cleanup of the compiler warnings in Analysis/QueryParser as well as renaming the QueryParser namespace to QueryParsers to avoid naming collisions with classes named QueryParser (similar to the Document namespace being Documents).

Note that this pull request contains the commits from #185 (which has already been reviewed), so you should pull that one first so you don't have to review the same thing twice. 

Current Status
==========

Misc
---

40 tests passing/42

#### 2 Failing Tests

1. Misc.Document.TestLazy() - The document is returning exactly (values * fields). The test [expects (values * fields) + 1](https://github.com/apache/lucene-solr/blob/8fdf89690404c0e65784b2c5477552b9dec58591/lucene/misc/src/test/org/apache/lucene/document/TestLazyDocument.java#L122). Without knowing what the +1 is supposed to represent, this is a wild goose chase. Any ideas?
2. Misc.Index.Sorter.TestEarlyTerminationDifferentSorter() - This test fails before it gets to the class under test due to a bug somewhere in the core. It fails about 15-20% of the time, but since the test is setup to run at least 5 times, the test fails about 90% of the time. I am unable to make it fail with manual input, so it is hard to figure out which random iteration will fail. The cause is a null reference exception due to the priority Queue being all nulls (its initial state). The UpdateBottom() method sets the Bottom member variable to null, and then the [code tries to access Bottom.Slot](https://github.com/NightOwl888/lucenenet/blob/misc-suggest/src/Lucene.Net.Core/Search/TopFieldCollector.cs#L397-L398). So, apparently some call is missing to update the priority queue and/or something is misbehaving and clearing it out. If I catch the exception, or put a null check before accessing the value (defaulting to 0), then the test passes. It seems like that null safety check should be there - or was the null check left out intentionally in Java to make it blow up when the PQ is empty?

#### Not Yet Ported

The Misc.Store directory. There is some C++ code in there for native Windows, Unix and Posix directories. I am not sure it makes any sense to port it.

Suggest
----

141 tests passing/141

Memory
----

10 tests passing/10


synhershko:
All 3 were already ported - can you please elaborate on the work done here?

Also some conflicts due to recent merges...

NightOwl888:
The projects were "ported", but they were not compiling and were not part of the solution. The fix for FST was to get the bugs out of it because both Misc and Suggest depend on it. Although Suggest only depends on 3 types in Misc, I ported nearly all of it.

The work done here was to finish up the unfinished implementation of Suggest, and to get its tests to pass (no easy feat, mind you).

Anyway, give me a few minutes to fix the merge conflict.

NightOwl888:
Okay, this one has rebased successfully without any issues.

PR #187, opened 2016-09-22
Fixed 76 QueryParser Failing Tests

NightOwl888:
This fixes 76 of the 77 failing tests in QueryParser. The remaining one is a matter of completing the test implementation (or removing it entirely depending on exactly how to handle the TimeZone).


synhershko:
Let's take the discussion on TimeZone to the dev@ mailing list - can you please start a thread there explaining the issue / dilemma and I will follow up?

src/Lucene.Net.QueryParser/Surround/Parser/QueryParser.cs
```diff
@@ -883,7 +883,7 @@ private void Jj_rescan_token()
                         p = p.next;
                     } while (p != null);
                 }
-                catch (LookaheadSuccess ls) { }
+                catch (LookaheadSuccess /*ls*/) { }
             }
             jj_rescan = false;
         }
```
synhershko: this file is auto-generated in Java so I'm not sure whether or not it's a good idea to change it here...

synhershko:
Great job, merging this now

src/Lucene.Net.QueryParser/Surround/Parser/QueryParser.cs
NightOwl888: Well, the alternative would be to use `#pragma` to disable the warning messages for the unused exception variable. Would you prefer it that way?

synhershko: Yes :)

PR #188, opened 2016-09-26
Fixed 64 Failing Facet Tests and Finished Facet Implementation

NightOwl888:
Fixed the 64 failing Facet tests, refactored the public API to be more .NET friendly (changing methods to properties and vice versa, renaming Size to Count, MaxSize to Capacity, Close() to Dispose(), etc.), and updated the documentation comments.

There is just [one area](https://github.com/NightOwl888/lucenenet/blob/facet-bugz/src/Lucene.Net.Facet/Taxonomy/WriterCache/NameIntCacheLRU.cs#L70-L80) that may need attention - the original implementation used a LinkedHashMap as a cache and switched it out for HashMap when the size grew large. The LRUHashMap is the closest thing we currently have to LinkedHashMap, but using it doesn't work. A plain old generic Dictionary works fine, but may use more RAM than the designers intended.

### DIFF
--- a/src/Lucene.Net.Facet/Taxonomy/LRUHashMap.cs
+++ b/src/Lucene.Net.Facet/Taxonomy/LRUHashMap.cs
@@ -22,33 +22,17 @@ namespace Lucene.Net.Facet.Taxonomy
      */
 
     /// <summary>
-    /// LRUHashMap is an extension of Java's HashMap, which has a bounded size();
-    /// When it reaches that size, each time a new element is added, the least
+    /// <see cref="LRUHashMap{TKey, TValue}"/> is similar to of Java's HashMap, which has a bounded <see cref="Capacity"/>;
+    /// When it reaches that <see cref="Capacity"/>, each time a new element is added, the least
     /// recently used (LRU) entry is removed.
     /// <para>
-    /// Java makes it very easy to implement LRUHashMap - all its functionality is
-    /// already available from <seealso cref="java.util.LinkedHashMap"/>, and we just need to
-    /// configure that properly.
+    /// Unlike the Java Lucene implementation, this one is thread safe. Do note
+    /// that every time an element is read from <see cref="LRUHashMap{TKey, TValue}"/>,
+    /// a write operation also takes place to update the element's last access time.
+    /// This is because the LRU order needs to be remembered to determine which element
+    /// to evict when the <see cref="Capacity"/> is exceeded. 
     /// </para>
     /// <para>
-    /// Note that like HashMap, LRUHashMap is unsynchronized, and the user MUST
-    /// synchronize the access to it if used from several threads. Moreover, while
-    /// with HashMap this is only a concern if one of the threads is modifies the
-    /// map, with LURHashMap every read is a modification (because the LRU order
-    /// needs to be remembered) so proper synchronization is always necessary.
-    /// </para>
-    /// <para>
-    /// With the usual synchronization mechanisms available to the user, this
-    /// unfortunately means that LRUHashMap will probably perform sub-optimally under
-    /// heavy contention: while one thread uses the hash table (reads or writes), any
-    /// other thread will be blocked from using it - or even just starting to use it
-    /// (e.g., calculating the hash function). A more efficient approach would be not
-    /// to use LinkedHashMap at all, but rather to use a non-locking (as much as
-    /// possible) thread-safe solution, something along the lines of
-    /// java.util.concurrent.ConcurrentHashMap (though that particular class does not
-    /// support the additional LRU semantics, which will need to be added separately
-    /// using a concurrent linked list or additional storage of timestamps (in an
-    /// array or inside the entry objects), or whatever).
     /// 
     /// @lucene.experimental
     /// </para>
@@ -74,6 +58,17 @@ namespace Lucene.Net.Facet.Taxonomy
             this.cache = new Dictionary<TKey, CacheDataObject>(capacity);
         }
 
+        /// <summary>
+        /// allows changing the map's maximal number of elements
+        /// which was defined at construction time.
+        /// <para>
+        /// Note that if the map is already larger than maxSize, the current 
+        /// implementation does not shrink it (by removing the oldest elements);
+        /// Rather, the map remains in its current size as new elements are
+        /// added, and will only start shrinking (until settling again on the
+        /// given <see cref="Capacity"/>) if existing elements are explicitly deleted.
+        /// </para>
+        /// </summary>
         public virtual int Capacity
         {
             get { return capacity; }
--- a/src/Lucene.Net.Facet/Taxonomy/LRUHashMap.cs
+++ b/src/Lucene.Net.Facet/Taxonomy/LRUHashMap.cs
@@ -95,20 +95,20 @@ namespace Lucene.Net.Facet.Taxonomy
                 if (cache.TryGetValue(key, out cdo))
                 {
                     // Item already exists, update our last access time
-                    cdo.Timestamp = GetTimestamp();
+                    cdo.timestamp = GetTimestamp();
                 }
                 else
                 {
                     cache[key] = new CacheDataObject
                     {
-                        Value = value,
-                        Timestamp = GetTimestamp()
+                        value = value,
+                        timestamp = GetTimestamp()
                     };
                     // We have added a new item, so we may need to remove the eldest
                     if (cache.Count > MaxSize)
                     {
                         // Remove the eldest item (lowest timestamp) from the cache
-                        cache.Remove(cache.OrderBy(x => x.Value.Timestamp).First().Key);
+                        cache.Remove(cache.OrderBy(x => x.Value.timestamp).First().Key);
                     }
                 }
             }
@@ -123,9 +123,9 @@ namespace Lucene.Net.Facet.Taxonomy
                 if (cache.TryGetValue(key, out cdo))
                 {
                     // Write our last access time
-                    cdo.Timestamp = GetTimestamp();
+                    cdo.timestamp = GetTimestamp();
 
-                    return cdo.Value;
+                    return cdo.value;
                 }
             }
             return null;
@@ -139,8 +139,8 @@ namespace Lucene.Net.Facet.Taxonomy
                 if (cache.TryGetValue(key, out cdo))
                 {
                     // Write our last access time
-                    cdo.Timestamp = GetTimestamp();
-                    value = cdo.Value;
+                    cdo.timestamp = GetTimestamp();
+                    value = cdo.value;
 
                     return true;
                 }
@@ -184,12 +184,12 @@ namespace Lucene.Net.Facet.Taxonomy
         private class CacheDataObject
         {
             // Ticks representing the last access time
-            public long Timestamp;
-            public TValue Value;
+            public long timestamp;
+            public TValue value;
 
             public override string ToString()
             {
-                return "Last Access: " + Timestamp.ToString() + " - " + Value.ToString();
+                return "Last Access: " + timestamp.ToString() + " - " + value.ToString();
             }
         }
 
--- a/src/Lucene.Net.Facet/Taxonomy/LRUHashMap.cs
+++ b/src/Lucene.Net.Facet/Taxonomy/LRUHashMap.cs
@@ -62,28 +62,28 @@ namespace Lucene.Net.Facet.Taxonomy
         // Record last access so we can tie break if 2 calls make it in within
         // the same millisecond.
         private long lastAccess;
-        private int maxSize;
+        private int capacity;
 
-        public LRUHashMap(int maxSize)
+        public LRUHashMap(int capacity)
         {
-            if (maxSize < 1)
+            if (capacity < 1)
             {
-                throw new ArgumentOutOfRangeException("maxSize must be at least 1");
+                throw new ArgumentOutOfRangeException("capacity must be at least 1");
             }
-            this.maxSize = maxSize;
-            this.cache = new Dictionary<TKey, CacheDataObject>(maxSize);
+            this.capacity = capacity;
+            this.cache = new Dictionary<TKey, CacheDataObject>(capacity);
         }
 
-        public virtual int MaxSize
+        public virtual int Capacity
         {
-            get { return maxSize; }
+            get { return capacity; }
             set
             {
                 if (value < 1)
                 {
-                    throw new ArgumentOutOfRangeException("MaxSize must be at least 1");
+                    throw new ArgumentOutOfRangeException("Capacity must be at least 1");
                 }
-                maxSize = value;
+                capacity = value;
             }
         }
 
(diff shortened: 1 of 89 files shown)


eladmarg:
Fantastic! this is awesome, thank you so much!
@synhershko please merge
Again, thanks a lot!

src/Lucene.Net.Facet/Taxonomy/LRUHashMap.cs
```diff
@@ -100,19 +120,25 @@ public bool Put(TKey key, TValue value)
                         timestamp = GetTimestamp()
                     };
                     // We have added a new item, so we may need to remove the eldest
-                    if (cache.Count > Capacity)
+                    if (cache.Count > capacity)
                     {
                         // Remove the eldest item (lowest timestamp) from the cache
                         cache.Remove(cache.OrderBy(x => x.Value.timestamp).First().Key);
```
eladmarg: This is working great, however, I think this will cause performance penalty on every put.
sometimes there are many items in the cache so each Put operation will cause expensive sorting.
in my previous implementation (I removed the PR), there is CleanPercentage, (so each clean we remove 10-15 percent).  
so a simple Take and remove will do the job, so we'll do cleaning less frequently.
amazing job!

src/Lucene.Net.Facet/Taxonomy/LRUHashMap.cs
```diff
@@ -100,19 +120,25 @@ public bool Put(TKey key, TValue value)
...
+            {
+                syncLock.ExitWriteLock();
             }
-            return true;
         }
 
         public TValue Get(TKey key)
         {
-            lock (syncLock)
+            syncLock.EnterWriteLock();
```
eladmarg: I'm not sure if it's worth the pay of price of thread safe here,
Generally, the main purpose of LRUCache is to delete the old entries, we want to store ~10K items, we don't really care if we removed the ones that has been used 2 ticks later. 
so the order is nice to have but not a must, this is a cache so the main goal is performance.
here is mine implementation, you can take a look.
https://gist.github.com/eladmarg/8d4a7f7a43a36c35d488e99475a101d9

yes, there is a chance that the Lru tests will fail, but we shouldn't care.

this is of course just a recommendation, we should get the version out and then think about performance.
Thank You!

eladmarg: I can confirm that after replacing to ConcurrentDictionary implementation, same all tests pass.
(Except TestLRUHashMap, but we can ignore it)

NightOwl888:
Thanks.

I am all for making this code run as efficiently and fast as possible, and agree there is room for improvement in the LRUHasMap. However, we are not inventing something new here, this is a port of an existing application and therefore must exhibit the same behavior. If you require a cache with different behavior than what is built-in, you can [implement your own ITaxonomyWriterCache](https://github.com/NightOwl888/lucenenet/blob/facet-bugz/src/Lucene.Net.Tests.Facet/Taxonomy/Directory/TestDirectoryTaxonomyWriter.cs#L275-L296). But without passing tests we are missing our mark for Lucene.Net.

The [documentation comments](https://github.com/apache/lucene-solr/blob/8fdf89690404c0e65784b2c5477552b9dec58591/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/LRUHashMap.java#L24-L48) and supporting tests are very clear on how the LRUHashMap is supposed to behave. There is not supposed to be any bulk removal of elements, they are removed one by one as new elements are added. We only remove the *least recently removed* element. There are no expiration dates or number of accesses to contend with.

I took a look at your implementation and it looks very similar to the original code that I changed to the current implementation. I am not sure why there is "Usage" being tracked in there, but it is certainly not something that is required to determine which element was *used least recently*. We don't really even need a date or timestamp for that - just an `int` counter would be enough. Also, we don't seem to be gaining anything from using ConcurrentDictionary since the only multi-step actions it supports are inadequate for LRU, so we would need locking anyway.

But my current implementation could also use some work. I think that it might work best to break this up into 2 or more internal data structures - perhaps one SortedSet and one Dictionary. I suspect that adding an item to an existing SortedSet in the right position is far more efficient than sorting a *whole* set. Then when we come along to delete the *least recent* item, it is simply a matter of getting the key of the item from the top of the SortedSet and using that key in Dictionary.Delete.

Perhaps it could be smarter than that, too. There could be a queue and some internal worker thread that manages the updates as well as the deletes in the background, but I am not sure if that would pass our tests, since they need to see the changes immediately.

Anyway, this is worth another pass, which is why I noted it above.

BTW, if you *really* want to thank me for doing this, please spend a weekend porting one of the remaining sections that doesn't have an open pull request. There are still a few sections that don't have any large library dependencies that would also need to be ported. Many of them were already ported in Lucene 3.0.3, which is a big help if you are stuck on trying to figure out how to make the pieces fit. We are almost there, but that doesn't mean we still don't need help!

eladmarg:
@NightOwl888 , you're absolutely right. this should be done by using two data structures,
here is a direction that might work:

[LinkedHashMap.cs](https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Util/LinkedHashMap.cs) or even this one: [LinkedHashMap.cs](https://github.com/grokit/website_grokit_ca/blob/master/articles/web/CSharpMissingDataStructures/LinkedHashMap.cs)

as you can see by the original [java implementation](https://github.com/apache/lucene-solr/blob/master/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/LRUHashMap.java) its only a wrapper, so one of them should work and they're both seems more efficient.

in another subject, which of the sub-project not ported yet?
Thank You

eladmarg:
by the way, another option is to port the java equivalent [LinkedHashMap](http://developer.classpath.org/doc/java/util/LinkedHashMap-source.html) derived from HashSet and just implement the Access method on Gets.

NightOwl888:
> which of the sub-project not ported yet?

AFAIK, these are the pieces that are missing that are not currently being worked on, but you should send an email to `dev@lucenenet.apache.org` and ask for the all clear to prevent any duplicate work (and if you don't receive a reply, consider it a "go ahead").

```
AFAIK, No External Dependencies (may want to double check, though)
----
Analysis.Kuromoji
Analysis.SmartCN
Analysis.Stempel
Demo
Replicator
QueryParser.Flexible (QueryParser has open pull requests...better to wait until they are merged)

With External Dependencies 
----
Analysis.ICU (Depends on ICU4j)
Analysis.Morfologik (Depends on Morfologik)
Analysis.Phonetic (Depends on Apache Commons)
Analysis.UIMA (Depends on Tagger, uimaj-core, WhiteSpaceTokenizer)
Benchmark (many dependencies)
Highlighter (Depends on Collator (which is still being ported) and BreakIterator (which we don't have a good solution for yet))
Sandbox (Depends on Apache Jakarta)

```

The projects with external dependencies need some research to find out if there are any good ports of those dependencies already out there. Even if they don't have, much of the code doesn't depend on those external dependencies, so we can get a head start on it by porting the parts that don't.

It almost goes without saying, we need both the code and the tests ported so we can see where the bugs are.

Also, be sure you are looking at the [4.8.0 tag](https://github.com/apache/lucene-solr/tree/releases/lucene-solr/4.8.0/lucene) when porting over the code (its difficult enough to get the pieces to fit without having to contend with different versions, too).

eladmarg:
Hi, 
just found a great implementation which does exactly what is expected (truly same as java)

please take a look at : [CSharpTest.Net.Collections](https://github.com/csharptest/CSharpTest.Net.Collections), there is a [nuget](https://www.nuget.org/packages/CSharpTest.Net.Collections/) package as well.

at the constructor use:
`new LurchTable<TKey, TValue>(LurchTableOrder.Access, capacity, EqualityComparer<TKey>.Default)
`
the lucene get method expect for default(TValue) or null, but the LurchTable throw exception,
so just wrap the get method with:
` TValue result;
if (!lru.TryGetValue(key, out result))
{
return default(TValue);
}
return result;
`
all tests working, (include the LruHashMap), performance much faster.

NightOwl888:
@eladmarg 

That's great news. 

I took a look and it seems okay. [Here is my branch](https://github.com/NightOwl888/lucenenet/tree/facet-lurch-table) showing it functioning.

1. [NameIntCacheLRU was updated](https://github.com/NightOwl888/lucenenet/blob/facet-lurch-table/src/Lucene.Net.Facet/Taxonomy/WriterCache/NameIntCacheLRU.cs#L73) to use `LurchTable` for LRU, and it is now passing all of the concurrency tests.
2. [LRUHashMap](https://github.com/NightOwl888/lucenenet/blob/facet-lurch-table/src/Lucene.Net.Facet/Taxonomy/LRUHashMap.cs) is now just an adapter around `LurchTable` so we get the right `Get` and `Put` behavior to match Java.
3. [I added some unit tests](https://github.com/NightOwl888/lucenenet/blob/facet-lurch-table/src/Lucene.Net.Tests.Facet/Taxonomy/TestLRUHashMap.cs#L63-L67) to demonstrate that the `Put` behavior functions properly. Although it took some work to figure out out how to achieve it, it turned out to be [quite simple](https://github.com/NightOwl888/lucenenet/blob/facet-lurch-table/src/Lucene.Net.Facet/Taxonomy/LRUHashMap.cs#L80-L86) to mimic Java's behavior in a thread-safe way.

However, there are a couple of issues:

1. The `Limit` property is readonly, but [in Lucene it needs to be changed](https://github.com/apache/lucene-solr/blob/releases/lucene-solr/4.8.0/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java#L330-L347) (increased) dynamically. So, for now I have it [throwing a `NotImplementedException`](https://github.com/NightOwl888/lucenenet/blob/facet-lurch-table/src/Lucene.Net.Facet/Taxonomy/LRUHashMap.cs#L69-L74).
2. We are getting this Lucene.Net implementation to work on .NET core, and there is currently no support for it with `CSharpTest.Net.Collections`.

The first issue can be easily patched. We could ultimately go back to putting the locks into DirectoryTaxonomyReader to update that property safely.

However, the second issue isn't something we can ignore. If there aren't too many dependencies, it might make the most sense to bring the code over into our Support namespace. 

If not, our only option may be to fork it, patch it, rename it, and post it on NuGet since the code hasn't been touched in over 2 years, there was only 1 release, and it is not very likely we will be able to get the original owner to update it (although we should try that first). I don't imagine getting a set of data collections to work on .NET core will be too much of a challenge.

Still, despite these issues, this is looking like our best option, because it is thread-safe.

The alternative would be to port `LinkedHashMap` from Java. While it is only ~750 lines of code, half of which are comments, its base class is `HashMap` and I am not sure how well it will fit with our current .NETified `HashMap` implementation (which doesn't have `Put` or `Get` methods). But even the author of the Lucene LRUHashMap [said it would be better if it were thread-safe rather than going this route](https://github.com/apache/lucene-solr/blob/releases/lucene-solr/4.8.0/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/LRUHashMap.java#L42-L48).

@synhershko - what are your thoughts on this?

eladmarg:
there are few files of the LurchTable, so all we need is @csharptest permission, and then we can just copy them to the solution and fork the relevant parts.

I think if we'll contact @csharptest, he won't have any opposition.
The LurchTable is a tested and working solution already, so i think its better to use what's working.

@synhershko is it possible to email @csharptest ?

csharptest:
I have no problem with this. Feel free to extract it. You can probably leave the interfaces behind if you don't need them. 


-- Roger
roger@csharptest.com

synhershko:
> A plain old generic Dictionary works fine, but may use more RAM than the designers intended.

I am totally okay with that. We should aim to get something out the door asap, and then optimize later.  This shouldn't keep us from going ahead, and I'd consider this a 80/20 case.

I say let's first release, then see who uses this. Once we see a lot of usage, we may reconsider the implementation. WDYT?

> if you really want to thank me for doing this, please spend a weekend porting one of the remaining sections that doesn't have an open pull request.

Indeed @eladmarg :)

On that note, @NightOwl888 I have no idea where you are located, but if I happen to be in your neighborhood during my travels do let me know and I'll buy you beers.

> in another subject, which of the sub-project not ported yet?

Let's have this discussion in the dev@ mailing list please. The list Shad provided may be correct, but our priorities are different - the spatial module needs work, and there are still failing tests at core. We can also skip Analysis.Kuromoji and Analysis.SmartCNand completely now - they aren't worth our efforts now. So, Let's have that discussion in the right place so other people could chime in as well.

And thank you @csharptest :) -- @eladmarg @NightOwl888 let's see if we can wrap this PR soon and move on to finish up the rest of the core and more important stuff.

NightOwl888:
@synhershko 

> I say let's first release, then see who uses this. Once we see a lot of usage, we may reconsider the implementation. WDYT?

In that case this is ready to go. But if we can spend 2-3 hours copying over an existing implementation that is almost a drop-in replacement for what we need, why not do one better?

Personally, I have a stake in faceted search - heck I have a whole open source project [BoboBrowse.Net](https://github.com/NightOwl888/BoboBrowse.Net), and I would like to have alternatives.

Also, Facet and Suggest are both brand new things to Lucene.Net. If you want to sway people's attention to this project, give them features they don't already have in Lucene.Net 3.0.3 (and make them as good as they can be). I have seen a few questions on the user mailing list from people who were trying to get this to work. Why not give them something useful?

> Let's have this discussion in the dev@ mailing list please. The list Shad provided may be correct, but our priorities are different - the spatial module needs work, and there are still failing tests at core. We can also skip Analysis.Kuromoji and Analysis.SmartCNand completely now - they aren't worth our efforts now. So, Let's have that discussion in the right place so other people could chime in as well.

Actually, I am nearly done with Analysis.Stempel now. I know it is not that critical, but it looked to be less than a day of work to finish and I started it before I knew whether these pull requests were going to be merged.

Also, nearly all of the "failures" in the core are due to the test runners screwing up. Many tests are in abstract classes that are meant to be run for multiple test subclasses. Those subclasses provide the setup that is necessary for the tests to run successfully. But, depending on what test runner you use, they are either being run without the setup *and* are being skipped or are being run *one additional time* in the abstract class without any setup, causing them to fail in the out-of-context run.

I know @conniey is currently working on setting up xUnit and I am hoping that its test runner(s) and/or specialized attributes are also a fix for this issue. If not, we need to find another solution that works regardless of the test runner used. The only thing I have come up with so far is to make those tests virtual, override them in the sub classes and only put the `[Test]` attribute on the stub methods in the sub classes, not in the abstract class. In fact, I did that in the Suggest tests before I realized how widespread this issue is (and that it also affects the QueryParser tests). It works, but creates a lot of duplicate stub methods that do nothing but call their base class.

And yes, I will continue this on the mailing list...

> On that note, @NightOwl888 I have no idea where you are located, but if I happen to be in your neighborhood during my travels do let me know and I'll buy you beers.

I am in Jomtien Beach, Thailand. And I'll take them draft, thank you :).

NightOwl888:
Okay, I added the LurchTable to the Support namespace - turned out to be pretty straightforward. This one is now ready to pull (I still can't push...).

Thanks @csharptest for permission to use your creation. And thanks @eladmarg for tracking it down.

csharptest:
No problem, glad you like it. Should serve you well as its been in production for years now without issue. Do let me know if you need advice on usage. 


-- Roger

PR #189, opened 2016-10-01
Fixed 24 Failing Analaysis Tests + Completed CharArraySet/CharArrayMap Implementation

NightOwl888:
This fixes the 23 Analysis tests that were failing all of the time, and 1 test that was failing about 10% of the time.

> NOTE: This includes the commits from #185.

Also, the CharArraySet and CharArrayMap classes were refactored:

1. An `ICharArrayMap` interface was introduced to allow CharArraySet to hold a reference to CharArrayMap without knowing its generic type.
2. A `MapValue` class was introduced to emulate Java's ability to put an `object()` instance into a generic type, whose only purpose is to make the value non-null (since null has a special meaning and default(TValue) would break the null check anyway if we were dealing with value types).
3. A `ValueCollection` class and `ValueEnumerator` class were introduced to house the logic that converts the `MapValue` to a real value, and to override the `ToString()` method to output the correct representation of the data for the tests.
4. A similar `KeyCollection` and `KeyEnumerator` were introduced to break the infinite recursion between CharArraySet/CharArrayMap without consuming RAM.
5. Overloads were created for all main methods of CharArrayMap and CharArraySet for `char[]`, `ICharSequence`, `string` and `object`. 
6. Extension methods were created for the main methods for string-able value types, converting them to strings automatically using the invariant culture.
7. The generic closing type of CharArrayMap and CharArraySet was changed from object to string (similar to how it was done in Lucene.Net 3.0.3).
8. Added a `CharArrayMap` class to house the static methods and public classes of `CharArrayMap<TValue>` so they can be accessed without specifying a generic class type for CharArrayMap (`CharArrayMap.Copy(x)` rather than `CharArrayMap<object>.Copy(x)`).

> NOTE: I know this looks a lot like a .NET 1.0 type with all of the overloads, but I ran some experiments and determined that it is 11x - 17x faster to do have several overloads than to have a single overload (whether it is generic or object) that checks the type and takes the appropriate action.

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/Compound/hyphenation/PatternParser.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Compound/hyphenation/PatternParser.cs
@@ -156,11 +156,12 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
                     // Element start
                     uri = node.NamespaceURI;
                     name = node.Name;
+                    bool isEmptyElement = node.IsEmptyElement;
                     var attributes = GetAttributes(node);
                     raw = string.Empty; // node.ReadOuterXml(); - not used, but was messing with the node pointer
 
                     this.StartElement(uri, name, raw, attributes);
-                    if (node.IsEmptyElement)
+                    if (isEmptyElement)
                     {
                         this.EndElement(uri, name, raw);
                     }
--- a/src/Lucene.Net.Analysis.Common/Analysis/Compound/hyphenation/TernaryTree.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Compound/hyphenation/TernaryTree.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.IO;
 using System.Text;
 
 namespace Lucene.Net.Analysis.Compound.Hyphenation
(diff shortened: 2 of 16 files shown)


synhershko:
Looks good - I will review properly once merge conflicts are resolved (so hopefully the diff will be smaller)

NightOwl888:
Yep, definitely a smaller diff.

synhershko:
Merge conflict again :\

LGTM, so feel free to merge and push - you should have permissions to do that by now :)

NightOwl888:
I haven't tried pushing to the repo, but it doesn't look like I have permissions in the GitHub control panel to merge pull requests.

It should only take a couple of minutes to fix this conflict if you bear with me. I needed some commits on multiple branches so I cherry-picked (in order) in order to ensure the changes were in multiple places. Rebasing works without any conflicts.

synhershko:
The github interface doesn't reflect on your permissions - please try and let us know if it doesn't work.

NightOwl888:
Nope, I have permissions issues:

```
F:\Projects\lucenenet>git push upstream master
Enter passphrase for key '/c/Users/Shad/.ssh/id_rsa':
ERROR: Sorry, but @apache has blocked access to SSH keys created by some third-party applications. Your key was created before GitHub tracked keys created by applications, so we need your help.

If you personally created this key, you can approve it at:

  https://github.com/settings/ssh/audit/3046281/policy

Otherwise, please upload a new key:

  https://github.com/settings/ssh

Fingerprint: 87:d3:f4:6f:df:78:25:86:fa:cf:20:1a:bc:50:56:ab

[EPOLICYKEYAGE]

fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

```

So I went to the URL and approved the SSH key and then...

```
F:\Projects\lucenenet>git push upstream master
Enter passphrase for key '/c/Users/Shad/.ssh/id_rsa':
ERROR: Permission to apache/lucenenet.git denied to NightOwl888.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

```

I am pushing to the GitHub repo (mirror). When I tried to push to the Apache repo, I got:

```
F:\Projects\lucenenet>git push apache master
fatal: remote error: access denied or repository not exported: /lucenenet.git
```

synhershko:
Ok, email us at private@ and let's figure this out. Ready to merge now?

PR #190, opened 2016-10-03
Ported Analysis.Stempel

NightOwl888:
All 16 tests are passing.

I ended up porting over `DataInputStream` and `DataOutputStream` and their corresponding interfaces from Java to save time. There is probably a way to make `Egothor.Stemmer` work with `BinaryReader` and `BinaryReader`, but these classes write the bytes in reverse order from Java and also do not have the corresponding `WriteUTF` and `ReadUTF` methods that use a proprietary UTF-8 format. The upside is that it now writes files that are *probably* compatible with Java, however since these are temp files I doubt it matters very much.


Lucene.Net.sln
```diff
@@ -59,14 +60,14 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Lucene.Net.Suggest", "src\L
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Lucene.Net.Tests.Suggest", "src\Lucene.Net.Tests.Suggest\Lucene.Net.Tests.Suggest.csproj", "{A6511598-3008-4A3B-AE68-2D1DA792CA8A}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Lucene.Net.QueryParser", "src\Lucene.Net.QueryParser\Lucene.Net.QueryParser.csproj", "{949BA34B-6AE6-4CE3-B578-61E13E4D76BF}"
-EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Lucene.Net.Tests.QueryParser", "src\Lucene.Net.Tests.QueryParser\Lucene.Net.Tests.QueryParser.csproj", "{5719FB4F-BF80-40E5-BACC-37E8E18FCA2E}"
-EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Lucene.Net.Memory", "src\Lucene.Net.Memory\Lucene.Net.Memory.csproj", "{42ECF239-AFC1-427D-921E-B5A277809CF0}"
```
synhershko: please revert this change

NightOwl888: I merged master into this branch - it seems to have removed this.

src/Lucene.Net.Core/Support/DataInputStream.cs
```diff
@@ -0,0 +1,323 @@
...
+using System.IO;
+
+namespace Lucene.Net.Support
+{
+    /// <summary>
+    /// Java's DataInputStream is similar to .NET's BinaryReader. However, it reads
+    /// using a modified UTF-8 format that cannot be read using BinaryReader.
+    /// This is a port of DataInputStream that is fully compatible with Java's DataOutputStream.
+    /// </summary>
+    public class DataInputStream : IDataInput, IDisposable
```
NightOwl888: I added all of the tests I could find from Java for these 2 classes. Do note that the ReadDouble() method is not implemented because we don't have `Number.LongBitsToDouble()` implemented. Neither are currently required by Lucene.Net

PR #191, opened 2016-10-31
Migrating Lucene.Net to .NET Core

conniey:
## Prerequisites

* Visual Studio 2015 with Update 3
* [.NET Core 1.0.1 SDK](https://www.microsoft.com/net/core)

## Building Lucene.Net.Portable.sln

This will build Lucene.NET for .NET Standard and .NET 4.5.1.

### Instructions

There is a build script, `Build.ps1` that performs build steps. For more information, execute `Get-Help Build.ps1 -Full` in Powershell.

#### Building

1. Go to repository root
2. Execute: `Build.ps1 -CreatePackages`

#### Testing

* Execute all tests:  `Build.ps1 -RunTests`
* Execute specific tests: `Build.ps1 -RunTests -ProjectsToTest @("Lucene.Net.Tests", "Lucene.Net.Tests.Facet")`
* Execute Manually: 
    1. `cd` to any of the testing projects
    2. `dotnet test --where="Category!=LongRunningTest && Category != DtdProcessingTest && Category != HasTimeout"`

## Building Lucene.Net.sln

This will build Lucene.NET using the old projects for .NET 4.5.1. **Hopefully Lucene.Net.sln can be removed in the future because the .NET Core projects compile for .NET 4.5.1 already.**

* Visual Studio
   1. Open solution in Visual Studio 2015
   2. Restore NuGet packages (right-click on solution, restore NuGet packages)
   3. Build solution
* Command Line 
   1. Open a Visual Studio 2015 developer command prompt
   2. Restore packages using NuGet v3
   3. Execute `msbuild Lucene.Net.sln`

## Notes

These are some explicit things to highlight about changes that were made in order to get it to build for .NET Standard 1.5

### Compiler preprocessors

Since .NET Standard is a subset of the full .NET Framework, there are some features that are not supported, so they had to be compiled out until support for them was enabled.  The following preprocessors were added to the code:

* FEATURE_SERIALIZABLE - Binary serialization (ie. `[Serializable]`) is [not supported](https://github.com/dotnet/corefx/issues/6564) in .NET Core.
* FEATURE_CLONEABLE - `ICloneable` interface isn't being brought back until .NET Standard 2.0 (Spring 2017)
* FEATURE_DTD_PROCESSING - `DtdProcessing.Parse` and `XmlResolver` isn't being brought back until .NET Standard 2.0 (Spring 2017). Issues [#4376](https://github.com/dotnet/corefx/issues/4376), [#10322](https://github.com/dotnet/corefx/issues/10322)
[...]

PR #193, opened 2016-11-07
Completed Grouping Implementation

NightOwl888:
This adds the remaining Grouping classes, the Grouping tests, as well as the support classes (`TreeSet` and `LinkedHashMap`) required by Grouping. 12/12 test are passing.

There is also a sweeping bug fixed in the test framework - the `TestUtil.RandomRealisticUnicodeString()` method was returning only random series of numbers, not Unicode characters. Fortunately, the impact wasn't as bad as I thought it would be - there are ~6 additional tests solution-wide failing due to this fix.

I am posting this as a pull request rather than pushing it to master for 2 reasons:

1. I am hoping it helps get #191 merged to master sooner (@conniey, let me know if it would be more helpful or harmful to merge this before you finish the .NET core integration).
2. There are a few things about this that need more work and further discussion.

API Refactoring
-----

Due to the liberal use of Java wildcard generics, it was a challenge putting something together that would compile. One of the solutions I came up with was to use .NET covariance to fix some of the issues. Unfortunately, covariance in .NET is only supported on interfaces and the fact that the Lucene designers made everything into abstract classes didn't help. Basically, we now have abstract classes that are backed by another abstraction (interfaces).

Furthermore, since the Collector abstraction is also an abstract class rather than an interface, it means these Collector interfaces cannot also be Collectors (which means lots of casting). This is because interfaces cannot inherit abstract classes. So, there are 3 different ways to go from here to get rid of most of the casting:

1. Change the Collector abstract class into an interface
2. Create an interface to back the Collector, similar to what was done for Grouping
3. Use a different solution than covariant interfaces to mimic Java wildcard generics

I am hoping someone can present another option than covariant interfaces that doesn't result in lots of casting, but that may not be a possibility. IMO, it would make more sense to change Collector into an interface (since it has no code and only public members) than to create a backing interface, which would help to avoid confusion over the double-abstraction.

What I mean is that pretty much every place where Collector abstract class is used now would need to change to the Collector interface to ensure that covariant Collector interfaces can inherit Collector, thus making them into a Collector. Keeping a Collector abstract class around after it is essentially no longer called out anywhere (except in subclasses) might be cause for confusion.

Also, it would be helpful if someone would work with Grouping and let us know of any other issues with the API that could be improved.

C5's TreeSet & TreeMap
----

I brought over the TreeSet and TreeDictionary from C5, along with the applicable tests. Unfortunately, this drags most of the C5 library along with it. We could alternatively reference it via NuGet, but there is not yet .NET core support (just an open [pull request](https://github.com/sestoft/C5/pull/50)).

Also, despite being a "set", it does not implement `ISet`, which I have now done. It isn't a problem yet, but may eventually be needed.

LinkedHashMap
-----

Unfortunately, the `LurchTable` that worked well for an LRU cache doesn't keep track of insertion order (even when you specify that option) and therefore doesn't replace `LinkedHashMap` as was advertised. So, I ended up modifying [this solution](http://qiita.com/matarillo/items/c09e0f3e5a61f84a51e2) (the second option) to be backed by `HashMap` rather than `Dictionary` so it can support `null` keys.

`LinkedHashMap`'s main reason for being is that it keeps track of the insertion order of items. We have unwittingly replaced it with `Dictionary` in several places. The fact that the tests pass is happenstance - `Dictionary` doesn't guarantee insertion order, but if you add items only and don't delete and re-add, they will enumerate in insertion order. That said, there is nothing stopping someone from changing this behavior from one .NET version to another and we can't rely on it. We should go back and find all of the places where `LinkedHashMap` is used in Lucene where we are using Dictionary, replace, and test thoroughly.


conniey:
@NightOwl888 Feel free to merge your PR into master whenever its ready. I am just cleaning up code and rerunning tests on my PR.

src/Lucene.Net.Core/Support/HashMap.cs
```diff
@@ -125,9 +140,87 @@ public TValue AddIfAbsent(TKey key, TValue value)
...
+                        if (!value.Equals(m[key]))
+                            return false;
+                    }
+                }
+            }
+            catch (InvalidCastException)
+            {
+                return false;
+            }
+            catch (NullReferenceException)
```
synhershko: why are you catching NullReferenceException here?

NightOwl888: It is the same in the JDK - http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/util/AbstractMap.java#AbstractMap.equals%28java.lang.Object%29

Basically, these overloads ensure that if HashMap is used as a dictionary key that they will be compared by their contained values rather than by reference (which is the default in .NET).

NightOwl888:
I have now merged these changes to master.

Do note that there is a new commit where the test class override stubs can be reverted (if needed): https://github.com/apache/lucenenet/commit/4dbc3590361814d13fae64c8d030820eb4987489

NightOwl888:
Oops, looks like the first attempt to push was rejected and I posted the wrong commit before. The new test class override rollback point is: https://github.com/apache/lucenenet/commit/a209d7183138d3f2179b03113626ce5708f6e835

PR #194, opened 2016-11-12
Updating from NUnit 2.6.3 to NUnit 3.5.0

conniey:
Updates the current test projects to NUnit 3.5.0.

* Replaces obsolete `TestFixtureSetUp` and `TestFixtureTearDown` with `OneTimeSetUp` and `OneTimeTearDown`
* Replaces `[ExpectedException]` with `Assert.Throws<T>`
* Adds a reason for [Ignore] tests because it is required now

Is there anything I need to edit in make it work in TeamCity?

PR #191 depends on this because [dotnet-test-nunit ](https://github.com/nunit/dotnet-test-nunit) uses NUnit 3.5.0

### DIFF
--- a/src/Lucene.Net.TestFramework/Index/BasePostingsFormatTestCase.cs
+++ b/src/Lucene.Net.TestFramework/Index/BasePostingsFormatTestCase.cs
@@ -488,7 +488,7 @@ namespace Lucene.Net.Index
             }
         }
 
-        [TestFixtureTearDown]
+        [OneTimeTearDown]
         public static void AfterClass()
         {
             AllTerms = null;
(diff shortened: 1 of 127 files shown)


conniey:
@synhershko @NightOwl888 Is there anything I need to do to get this merged in?

synhershko:
Looks good to me, but seems like it's blocked by #191?

conniey:
@synhershko This doesn't depend on #191. I created this PR off of master.

synhershko:
Ok, cool, merging now

NightOwl888:
@synhershko 

I reverted these commits because we already have this change in #191 and it causes a conflict, not to mention it breaks the build (see #200).

synhershko:
Cool thanks - that's why I asked before merging...

conniey:
@synhershko @NightOwl888 When I had opened this PR, no one was working on my branch, so I was going to rebase my commits on top of this so there would be less changes to look at in #191. :) But since there is active development.

PR #195, opened 2016-11-23
Moving ICU dependent classes into a separate project

laimis:
Would this help to deal with ICU issues that people are encountering in using early versions of 4.8 port? Seems like a quick way to get past 64-bit issues while a long-term solution for ICU is found.

This PR moves Analysis.Common/Th/* and Util/SegmentingTokenizerBase into a separate project allowing to remove ICU4NET dependency from Lucene.Net.Analysis.Common.


synhershko:
👍 

WDYT @NightOwl888 ?

NightOwl888:
No offense, but since @conniey has already provided a [working solution to the ICU bitness issue](https://github.com/apache/lucenenet/pull/191#issuecomment-262379449), this seems like a step backward instead of a step forward. At the very least, this will create a lot of merge conflicts for #191 and an unneeded NuGet package to be created and then abandoned shortly afterward. Not to mention, naming the package `Lucene.Net.Analysis.Common.ICU` is a poor naming choice, since there is another package `Lucene.Net.Analysis.ICU`, which will cause lots of confusion if/when it is ported. We are getting very close to the point where #191 can be merged - we are probably just a few days away from that point, after which we will be able to get the pre-release on NuGet shortly afterward.

That said, it might pay off if some interested parties take a look at her [demo](https://github.com/conniey/lucenenet-demo) and create prototypes to verify icu-dotnet works for all environments. That is what will happen anyway when we get the pre-release on NuGet, but we might be able to head off any remaining issues earlier with earlier feedback.

In the meantime, if anyone needs an immediate solution to this, they should choose one of the following options:

1. [Manually build and pack #191](https://github.com/apache/lucenenet/pull/191#issue-186402248) (recommended)
2. Manually build this branch (will require some extra work to pack if you need NuGet packages)
3. Use the [MyGet packages](https://www.myget.org/gallery/lucene-net) and delete the ICU DLLs (assuming you don't need Thai support or to utilize the SegmentingTokenizerBase). This is not guaranteed to work, but has worked in at least one case.

laimis:
@NightOwl888 I wasn't sure how close #191 was, that's why offered this as an option so that Analysis package could be released without ICU dependencies. I will go ahead and close this if it's not necessary and #191 is about to be merged in. Thanks.

PR #196, opened 2016-11-28
Renamed hyphenation to Hyphenation

DevMk:
Renamed hyphenation to Hyphenation to fix build and run on case sensitive file systems


synhershko:
Thanks!

PR #197, opened 2016-12-06
Fixing build errors

conniey:
Currently compiling master results in build errors:

```
  core\Search\JustCompileSearch.cs(204,46): error CS0507: 'JustCompileSearch.JustCompileFilteredDocIdSet.Match(int)': cannot change access modifiers when overriding 'protected' inherited member 'FilteredDocIdSet.Match(int)' [D:\git\conniey\lucenenet\src\Lucene.Net.Tests\Lucene.Net.Tests.csproj]`
```

This is due to changes in [FilteredDocIdSet](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Core/Search/FilteredDocIdSet.cs#L95) from: 
`protected internal abstract bool Match(int docid);` to `protected abstract bool Match(int docid);`

This fixes those errors.  An alternative would be to add the `internal` modifier back to FilteredDocIdSet.

@NightOwl888 Do you know why it was removed in the first place?

### DIFF
--- a/src/Lucene.Net.Tests/core/Search/JustCompileSearch.cs
+++ b/src/Lucene.Net.Tests/core/Search/JustCompileSearch.cs
@@ -201,7 +201,7 @@ namespace Lucene.Net.Search
             {
             }
 
-            protected internal override bool Match(int docid)
+            protected override bool Match(int docid)
             {
                 throw new System.NotSupportedException(UNSUPPORTED_MSG);
             }
--- a/src/Lucene.Net.Tests/core/Search/TestDocIdSet.cs
+++ b/src/Lucene.Net.Tests/core/Search/TestDocIdSet.cs
@@ -135,7 +135,7 @@ namespace Lucene.Net.Search
                 this.OuterInstance = outerInstance;
             }
 
-            protected internal override bool Match(int docid)
+            protected override bool Match(int docid)
             {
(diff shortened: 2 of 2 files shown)


NightOwl888:
Not sure why this happened - probably a botched merge of some kind. I'll pull this into master.

For future reference, it would work smoother on my end if you just add `fixes #197` or `closes #197` to the commit message. I don't have access to be able to merge from GitHub, but if you add one of those to the commit message, the pull request will be closed automatically when merged.

conniey:
@NightOwl888 Ahh okay! Thanks for letting me know. Just to make sure I understand it correctly, when I am creating a commit, I'll preemptively add "fixes <issue number>" to my commit message.  And the <issue number> where it'll be the number my PR will be when I create it?

NightOwl888:
See https://github.com/apache/lucenenet/commit/e8735ed9622ab56eaf86bae6394f93f3fb058d30 for an example.

If it is something that needs to go in right away (such as this) you could add to the message preemptively (and hope nobody else beats you to it). If not, you can just add a commit at the end of the PR indicating to close. It really doesn't matter so much, you can close any PR or GitHub issue from any commit by using these commit message commands.

conniey:
@NightOwl888 Awesome! Thanks for the explanation. I updated my commit message.

PR #198, opened 2016-12-06
Removing C# 6 features

conniey:
There are C# 6 features in use (ie. [NumericConfig.cs](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.QueryParser/Flexible/Standard/Config/NumericConfig.cs#L3)) which will not build in Visual Studio 2013.  This removes those features.

@NightOwl888 @synhershko : Alternatively, I can replace the C# 6 features. What do you think?

### DIFF
--- a/src/Lucene.Net.QueryParser/Flexible/Standard/Config/NumericConfig.cs
+++ b/src/Lucene.Net.QueryParser/Flexible/Standard/Config/NumericConfig.cs
@@ -1,6 +1,6 @@
 ﻿using Lucene.Net.Support;
 using System;
-using static Lucene.Net.Documents.FieldType;
+using Lucene.Net.Documents;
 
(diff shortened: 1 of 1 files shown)


NightOwl888:
Thanks for informing - this was something that slipped in by accident. Ever since I first noticed that VS2012 can't compile these features, I have been trying to keep them out of the code base.

Is NumericConfig.cs the only culprit, or are there more?

conniey:
@NightOwl888 I replaced all the C# 6 features instead, so it'll continue to work in older VS versions.

jeme:
FYI: Project Properties -> Build -> Advanced -> Language Version.
e.g. <LangVersion>4</LangVersion>

I Would have gone in the other direction though and migrate up, unless there is a build server issue somewhere?

PR #200, opened 2016-12-19
Adding nunit 3 console package to enable TeamCity to run the tests.

wwb:
Adding Nunit 3 console package to enable TeamCity to run the nunit 3.5 tests.

### DIFF
--- a/src/Lucene.Net.Tests.Analysis.Common/packages.config
+++ b/src/Lucene.Net.Tests.Analysis.Common/packages.config
@@ -2,4 +2,11 @@
 <packages>
   <package id="ICU4NET-ICU4C55.1-bin32" version="1.0.0" targetFramework="net451" />
   <package id="NUnit" version="3.5.0" targetFramework="net451" />
+  <package id="NUnit.Console" version="3.5.0" targetFramework="net451" />
+  <package id="NUnit.ConsoleRunner" version="3.5.0" targetFramework="net451" />
+  <package id="NUnit.Extension.NUnitProjectLoader" version="3.5.0" targetFramework="net451" />
+  <package id="NUnit.Extension.NUnitV2Driver" version="3.5.0" targetFramework="net451" />
+  <package id="NUnit.Extension.NUnitV2ResultWriter" version="3.5.0" targetFramework="net451" />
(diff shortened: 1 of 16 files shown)


NightOwl888:
Hi, does this have anything to do with #194? We should probably revert that change, since it causes a conflict with #191 (which we are actively working on).

We really should be working off of #191 only at this point.

wwb:
Yes -- this is to keep TeamCity running with that in play. You can reject this one and revert that. 

It would be cool to get #191 merged so we can get TeamCity working with the new layout, not sure what is between us and that though.

NightOwl888:
> It would be cool to get #191 merged so we can get TeamCity working with the new layout, not sure what is between us and that though.

I know there is at least:

- Tests that are failing/running infinitely over there that are not here
- Working out what to do about the build/TeamCity. The build script is setup to run the build for the entire solution at once. But you seem to be running one project at a time (at least for testing). Is there any reason we can't just hook into a single build script that does the build? Or should we aim to make the build script work at the project level?

It looks like @conniey already setup a TeamCity task to run the build once we get over there, but I am still unsure it can accept our current versioning scheme (which I have confirmed will do the trick and I think we should keep).

wwb:
If you are talking about [this one] (https://teamcity.jetbrains.com/viewType.html?buildTypeId=LuceneNet_Vs2015LuceneNetPortable) -- I set that one up to wrap my head around what needed to change to run the new version of the project. It still isn't quite perfect but it is getting there. Currently I am using build.ps1 to build things. I am still working on getting the tests to run successfully. It can easily be made to accept the current versioning once we unify the codebase.

The individual project builds and tests are at @synhershko 's request. Mechanically, the [Lucene.net.core](https://teamcity.jetbrains.com/viewType.html?buildTypeId=LuceneNet_Core)  build is the real build and test with the nightly generating the myget packages. At the end of the day the build system has to take into account that we want individual packages for the sub-projects.

In terms of a path forward I suggest that we just get a clean build up with the core things in place and then fix the tests -- as well as build out nuget packaging. How we want to build it changes based on the new structure so we might as well get that in place and build on top of it.

PR #201, opened 2016-12-19
updating assemblyinfo files for nuget package metadata.

wwb:
Updating the AssemblyInfo files to provide better myget metadata for now. This should not conflict w/ #191.


synhershko:
lgtm, thanks!

PR #202, opened 2017-01-02
Fix TermText property accessibility

konopka:


NightOwl888:
Hello, and thanks for the PR. However, API accessibility has already been addressed on my personal branch (a fork of #191) and for now I am working on making it stable before merging it here.

Therefore, it would be better to close this PR than to merge it, since it will surely cause a merge conflict.

konopka:
Ok, I am closing it then

PR #204, opened 2017-03-25
Adding missing build configurations for highlighter and highlighter test

vvdb:


NightOwl888:
Cool - certainly will be better than sending files one by one :).

PR #207, opened 2017-05-16
API: Added ReferenceManager<G>.AcquireContext()

NightOwl888:
Opening this as a pull request so the team can discuss this approach vs the previous `ExecuteSearch` method (that was inadvertently removed from the API).

## AcquireContext Example

```c#
var searcherManager = new SearcherManager(indexWriter, true, null);
using (var context = searcherManager.AcquireContext())
{
	IndexSearcher searcher = context.Reference;
	var topDocs = searcher.Search(query, 10);

       // use results...
}
```

### Pros

1. Eliminates the try-finally block.
2. Doesn't swallow exceptions by default.
3. Implicitly cleans up.
4. Since the extension method is on `ReferenceManager<G>`, all of its subclasses automatically gain this functionality.

### Cons

1. If the user forgets the using block, cleanup is not automatic.

## ExecuteSearch Example
```c#
var searcherManager = new SearcherManager(indexWriter, true, null);
searcherManager.ExecuteSearch(searcher =>
{
	var topDocs = searcher.Search(query, 10);

       // use results...
}, exception => { Console.WriteLine(exception.ToString()); });
```

### Pros

1. Eliminates the try-finally block.
2. Implicitly cleans up.
3. `ExecuteSearch` name makes it easier to find on the API.
4. The user doesn't need to remember any special syntax to clean up.

### Cons

1. Swallows exceptions by default. To make the exception bubble, you have to re-throw it.
2. API only applies to SearcherManager, so similar code needs to be repeated for other `ReferenceManager<G>` subclasses.

### Discussion

Although the `ExecuteSearch` method is well named and thus easier to find on the API, it has the disadvantage of swallowing exceptions by default. Exceptions do not automatically bubble - you have to re-throw the exception (in which case performance suffers and you lose the stack trace). Its main advantage is that there is no way for the user to accidentally forget to clean up.

The pattern seems to lend itself better to a using block. If all code examples are shown with a using block the user is not likely to forget it. However, admittedly, `AcquireContext` and `context.Reference` are probably not the most intuitive names (I was torn between `Context` or `Holder` for the name, but there might be a better choice than either of those). `context.Reference` was chosen because `ReferenceContext` is the name of the generic class that holds the reference and it applies to all subclasses of `ReferenceManager<G>`. Of course, the original API was named `Acquire()`, so it seems logical to use `AcquireContext()` or `AcquireHolder()` or something along those lines. Ideas welcome.

### DIFF
--- a/src/Lucene.Net/Lucene.Net.csproj
+++ b/src/Lucene.Net/Lucene.Net.csproj
@@ -653,6 +653,8 @@
     <Compile Include="Support\ICallable.cs" />
     <Compile Include="Support\ICharSequence.cs" />
     <Compile Include="Support\RectangularArrays.cs" />
+    <Compile Include="Support\Search\ReferenceContext.cs" />
+    <Compile Include="Support\Search\ReferenceManagerExtensions.cs" />
     <Compile Include="Support\Threading\ICompletionService.cs" />
     <Compile Include="Support\IO\IDataInput.cs" />
     <Compile Include="Support\IO\IDataOutput.cs" />
--- a/src/Lucene.Net/Support/Search/ReferenceContext.cs
+++ b/src/Lucene.Net/Support/Search/ReferenceContext.cs
@@ -1,0 +1,64 @@
+﻿using System;
+
+namespace Lucene.Net.Search
+{
+    /*
+	 * Licensed to the Apache Software Foundation (ASF) under one or more
+	 * contributor license agreements.  See the NOTICE file distributed with
+	 * this work for additional information regarding copyright ownership.
+	 * The ASF licenses this file to You under the Apache License, Version 2.0
+	 * (the "License"); you may not use this file except in compliance with
+	 * the License.  You may obtain a copy of the License at
+	 *
+	 *     http://www.apache.org/licenses/LICENSE-2.0
+	 *
+	 * Unless required by applicable law or agreed to in writing, software
+	 * distributed under the License is distributed on an "AS IS" BASIS,
+	 * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	 * See the License for the specific language governing permissions and
+	 * limitations under the License.
+	 */
+
+    /// <summary>
+    /// <see cref="ReferenceContext{T}"/> holds a reference instance and
+    /// ensures it is properly de-referenced from its corresponding <see cref="ReferenceManager{G}"/>
+    /// when <see cref="Dispose()"/> is called. This class is primarily intended
+    /// to be used with a using block.
+    /// <para/>
+    /// LUCENENET specific
+    /// </summary>
+    /// <typeparam name="T">The reference type</typeparam>
+    public class ReferenceContext<T> : IDisposable
+        where T : class
+    {
+        private readonly ReferenceManager<T> referenceManager;
+        private T reference;
+
+        internal ReferenceContext(ReferenceManager<T> referenceManager)
+        {
+            this.referenceManager = referenceManager;
+            this.reference = referenceManager.Acquire();
+        }
+
+        /// <summary>
+        /// The reference acquired from the <see cref="ReferenceManager{G}"/>.
+        /// </summary>
+        public T Reference
+        {
+            get { return reference; }
+        }
+
+        /// <summary>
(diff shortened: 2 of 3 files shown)


geobmx540:
The first approach is much more intuitive and inline with the rest of the C# ecosystem.

NightOwl888:
What about naming? Do you think it would be more intuitive to call it `AcquireReference()` and make the class named `Reference` instead of `ReferenceContext`? Seems like that would be a bit more intuitive from a user point of view.

geobmx540:
Is our reference only an `IndexSearcher` and it's descendants? I'm partial to the `Context` language - I'm using EF as my implementation reference because I think it's widely known, accepted and pretty well architected. `DbContext` is well known, just as perhaps `LuceneContext` could be very intuitive (as the class) with the method 'Context()'.

We could also overload  `ReferenceContext.cs` (or `LuceneContext`) to have pass through methods directly to the context/reference IndexSearch to avoid the extra method call to get the Reference. This would require an additional constraint on `T`

```
using (var context = searcherManager.Context()) 
{
      // General use case
     var topDocs = context.Search(query, 10);

     // Cast if you're trying to reach a descendant class type with additional methods / properties
     BetterIndexSearcher searcher = (BetterIndexSearcher)context.Context;
     var betterTopDocs = searcher.BetterSearch(query,10);
}
```

or alternatively,

using (var context = (BetterIndexSearcher)searcherManager.Context()) 
{
      var topDocs = context.BetterSearch(query, 10);
}
```

**Comment by NightOwl888** [comment]
Looks nice. Although, I tend to favor the "all methods are verbs" and "all properties are nouns or adjectives"  methodology. For example, I would expect a method named `Buffer()` to buffer something, not return a buffer. In this case that argument is a bit weaker, but it still feels like `GetContext()` would be more clear than `Context()`.

I primarily named it `AcquireContext()` because it is essentially the same functionality as the existing `Acquire()` method (from Java), so the similar names seem natural. But `GetContext()` is also clear (and more concise).

Hmm, that wasn't what I meant by "other subclasses". Actually, I made the constructor of `ReferenceContext` internal so it is effectively sealed to the outside world. What I was referring to is the fact that SearcherManger is just one of a few (with many more possible) subclasses of `ReferenceManager<G>`. 
And since `ReferenceManager<G>` is generic, the `Reference` property is also generic and the cast is not necessary - its type is specific to the subclass of `ReferenceManager<G>`. 

However, we can't very well name the property `Searcher` (so we have `context.Searcher`) because it wouldn't be generic enough for the other subclasses of `ReferenceManager<G>`. Well, we could by making extension methods and a custom "holder" type for each subclass of `ReferenceManager<G>`. But the disadvantage there is repetitive similar code. Maybe it would be worth it for such an integral part of the API, though.

The purpose of the using block here isn't to dispose anything really - it is just to help ensure the variable isn't utilized after its reference is no longer tracked. The context is just a "holder" for the reference to implement `Dispose()` on, which tells the parent `ReferenceManager<G>` class that we are done using the reference. You can see an example of how it looks with the Java functions [here](https://github.com/synhershko/LuceneNetDemo/blob/master/LuceneNetDemo/GitHubIndex.cs#L157-L184) with a more verbose finally block. Same thing, but shaves a few lines of code off of the block. You don't gain much if you need to catch an exception, though because you still need to put a try catch inside of the using block.

The using block does do one nice thing. Since it is a block, the variable goes out of scope when it is done.

```c#
using (var context = searcherManager.AcquireContext)
{
    var foo = context.Reference;
    foo.Search();
} // foo goes out of scope
```
The only issue is that it is not 100% obvious that this is a problem:

```c#
Searcher foo;

using (var context = searcherManager.AcquireContext)
{
    foo = context.Reference;
} // foo doesn't go out of scope - BAD

foo.Search(); // Not allowed here
```

But this just seems to be a scenario where there is no real way to force the end user to follow the rules. You just have to document the right way and hope they follow the documentation.

geobmx540:
Ah, ok I misunderstood your use case regarding other subclasses. I agree, the amount of repetition in the code for that is probably unwieldy.
 
No argument with `GetContext()`, even though it doesn't match java, it matches .net style (i.e. `GetEnumerator()`

PR #210, opened 2017-08-01
Corrected bug in Lifetime reference count

vvdb:
The addFactory in ConcurrentDictionary's GetOrAdd isn't guaranteed to be called only once, yielding incorrect reference counts. Switched to a lazily evaluated singleton to avoid this

### DIFF
--- a/src/Lucene.Net/Search/SearcherLifetimeManager.cs
+++ b/src/Lucene.Net/Search/SearcherLifetimeManager.cs
@@ -145,7 +145,7 @@ namespace Lucene.Net.Search
         // TODO: we could get by w/ just a "set"; need to have
         // Tracker hash by its version and have compareTo(Long)
         // compare to its version
-        private readonly ConcurrentDictionary<long, SearcherTracker> _searchers = new ConcurrentDictionary<long, SearcherTracker>();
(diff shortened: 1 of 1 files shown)


NightOwl888:
Thanks.

I would have thought you would be more creative with the branch name, though.

I pushed this to master, but somehow GitHub didn't get the message and left it open. For whatever reason, I don't have the ability to close pull requests on GitHub from the UI, the only way I can do it is by pushing another commit with a special `#closes 210` commit message. So, if you don't mind (and you can see a close button here), please close this pull request, it is accepted.

vvdb:
Good morning Shad,

I followed your detailed instruction to the letter and was so happy it worked. And “vincents-amazing-patch” sounded creative enough <g>.
I closed the pull request, and got this as a message:

Closed with unmerged commits
This pull request is closed, but the vvdb:vincents-amazing-patch branch has unmerged commits.

… with a button  “Delete branch”. I did not click on that button. But can I? If so, next time I’ll have a more creative name, such as vincents-attempt-at-a-pull-requests-waiting-for-Shad-s-approval. Or something like that…

Vincent

NightOwl888:
As you can see your commit is here: https://github.com/apache/lucenenet/commits/master

What you do with the branch on your GitHub account is up to you. Personally, I usually just keep them in case I need to review them later - it is easier to find your commits when they are on top of your own branch than buried among everyone else's. And GitHub won't send you a bill for keeping them.

PR #212, opened 2017-08-14
LUCENENET-469 - enumerables

AndyPook:
See https://issues.apache.org/jira/projects/LUCENENET/issues/LUCENENET-469 for background.

This is an early stage proposal for implementing IEnumerable<T> on various "Enum" types. To enable `foreach` and linq style operators.

This prototype provides `EnumEnumerator<T>` as a helper. It can be used in various ways to make the implementation of `IEnumerable<T>` more straightforward.

I have modified `IBytesRefIterator` and `TermsEnum` to add `IEnumerable<BytesRef>` and therefore to all of it's descendants.

The test (such as it is) demonstrates retrieving a `TermsEnum` from a reader. Then using `foreach` and a simple `.Count()` linq operator.

If this approach is acceptable it ought to be adaptable to other Enum types

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+# Code Files
+[*.{cs,csx}]
+indent_style = space
+trim_trailing_whitespace = true
+csharp_style_var_for_built_in_types = true:suggestion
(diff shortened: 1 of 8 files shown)


NightOwl888:
Thanks for the PR, but we decided to take another approach that removes the Java-style iterators entirely rather than simply wrapping them.

PR #214, opened 2017-12-21
Update ConcurrentMergeScheduler.cs

phillipc:
Add a comment: Non-treatable exception

### DIFF
--- a/src/Lucene.Net/Index/ConcurrentMergeScheduler.cs
+++ b/src/Lucene.Net/Index/ConcurrentMergeScheduler.cs
@@ -732,6 +732,9 @@ namespace Lucene.Net.Index
             //{
             //    throw new ThreadInterruptedException("Thread Interrupted Exception", ie);
             //}
+            
+            //Note: Non-treatable exception
+            //In this case, the whole process dies probably because the exception isn't handled (except: AppDomain.UnhandledException)
             throw new MergePolicy.MergeException(exc, m_dir);
         }
 


laimis:
I spent some more time on this, see the comment on https://issues.apache.org/jira/projects/LUCENENET/issues/LUCENENET-603, but essentially our behavior is the same as in lucene, so we should not merge this without a more in-depth explained reason.

lmnash:
This pull request just adds a comment, so wouldn't actually resolve the issue.  I have created my own pull request duplicating the changes I have already made in our system in order to resolve the issue.
https://github.com/apache/lucenenet/pull/220

jeme:
Considering that the HandleMergeException method is virtual, it's possible to override the behavior in a custom MergeScheduler and pass that to the IndexWriterConfig...

Is there any reason that you can't do that?

lmnash:
Yes, I could create a custom MergeScheduler.  For the moment I am using slightly modified 3.0.3 code with the throw commented out (as of July 2018) and that is working fine in multiple Production environments.

The point is that this is a defect which will impact other users (and presumably did since this is not my pull request).  It is clear from the comments in the code that the expected behaviour is not for the application to crash but for the thread to end and another attempt to be made.  I presume the behaviour has changed due to the port from Java because of different behaviour in Java and .NET around unhandled exceptions in threads.

I'm not looking for a workaround, I'm suggesting introducing a permanent fix.

lmnash:
I found this about the Java behaviour: https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.UncaughtExceptionHandler.html

In contrast with .NET behaviour:
https://docs.microsoft.com/en-us/dotnet/standard/threading/exceptions-in-managed-threads

jeme:
Sorry, my example left out a vital piece...

```
    public class MyConcurrentMergeScheduler : ConcurrentMergeScheduler
    {
        protected override void HandleMergeException(Exception exc)
        {
        }
    }
```

I wouldn't consider inheriting from ConcurrentMergeScheduler and replacing the behavior as a workaround... There is a reason the the method is virtual...

sisve:
Would the SetSuppressExceptions() method be useful here? Calling it will basically turn off the call to HandleMergeException() which means the exception this PR comments out will never be thrown.

PR #215, opened 2018-02-07
remove redundant termAtt initialization in UpperCaseFilter constructor

igor-brylin:
Constructor of UpperCaseFilter has duplicated initalization code for termAttr field.
I have modified UpperCaseFilter constructor by removing duplicate.


NightOwl888:
Thanks!

PR #216, opened 2018-10-01
Add support for netstandard2.0 for Queries and SandBox

gabrieldelaparra:
I was trying to use Lucene.Net on a side project with Net Core 2 and Blazor for front end.
I needed the Queries and QueryParser (which requires SandBox). Those 2 libraries are were not built for netstandard2.0, now they work.

### DIFF
--- a/src/Lucene.Net.Queries/Lucene.Net.Queries.csproj
+++ b/src/Lucene.Net.Queries/Lucene.Net.Queries.csproj
@@ -24,7 +24,7 @@
   <Import Project="..\..\Version.proj" />
   
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.6;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard1.6;net45</TargetFrameworks>
     <PackageTargetFallback Condition=" '$(TargetFramework)' == 'netstandard1.6' ">$(PackageTargetFallback);dnxcore50</PackageTargetFallback>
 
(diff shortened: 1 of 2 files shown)


lahma:
A bit out of topic, more of question to maintainers. Should the support be only for netstandard2.0 and net452? Microsoft's recommendation is just to target netstandard20 and for the ones who need older than 4.7.2 full framework, .NET 4.5.2 the oldest still supported framework version. I see little value for 1.6 if it only gives extra maintenance burden.

NightOwl888:
@lahma 

.NET Standard 1.6 support is to support .NET Core 1.x, it has nothing to do with supporting legacy versions of .NET Framework. From what I understand (although I can't seem to find the doc), NuGet will always choose a native framework target over .NET standard if there is one.

Should we still support .NET Core 1.x? That is another question altogether. Frankly, I wish Microsoft hadn't released what was clearly an unfinished product and called it 1.0, but now that it is in the wild, those who can support it, probably should.

NightOwl888:
There is a tool that simulates the behavior of NuGet on which target version to select.

http://nugettoolsdev.azurewebsites.net/4.8.0/get-nearest-framework?project=net472&package=net45%0D%0Anetstandard1.6%0D%0Anetstandard2.0

If you install Lucene.Net into a project that targets .NET Framework 4.7.2, it will pick the `net45` target depsite the fact that .NET Standard 1.6 supports `net461`. NuGet has the final say unless you manually unpack and install the .NET Standard 1.6 version into your project.

PR #218, opened 2019-01-17
Fixed TestThreadInterruptDeadlock and TestTwoThreadsInterruptDeadlock

sisve:
This fixes the TestThreadInterruptDeadlock  and TestTwoThreadsInterruptDeadlock.

I've added a ThreadClass::interrupted() to match Java's Thread.interrupted(). Most of the issues seems related to the interrupt flag not being cleared, which will cause a ThreadInterruptedException next time the thread blocks.

Changing MaybeYield to actually yield seems to be needed to avoid some ThreadInterruptedException. It makes sense in my head, since a yield isn't a block and thus an exception isn't thrown, but I have no documentation or evidence that this argument actually holds true. Anyhow, changing to yield matches the Java behavior, so I think this is the correct behavior.

### DIFF
--- a/src/Lucene.Net.TestFramework/Store/MockDirectoryWrapper.cs
+++ b/src/Lucene.Net.TestFramework/Store/MockDirectoryWrapper.cs
@@ -580,7 +580,11 @@ namespace Lucene.Net.Store
         {
             if (RandomState.NextBoolean())
             {
+#if NETSTANDARD1_6
                 Thread.Sleep(0);
+#else
+                Thread.Yield();
+#endif
             }
         }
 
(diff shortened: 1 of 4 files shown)


laimis:
@sisve let me spend some more time on this, and also see if we can get the teamcity builds going again, before I merge. Any change to IndexWriter should be treated with care 😄

NightOwl888:
Looks better, but still not completely stable. Sometimes they pass, but often there is a failure. If you put a `[Repeat(15)]` attribute at the top of each test, they will fail consistently.

Also, be sure to only target Release mode. There seems to be more crashes in Debug mode that go away in Release.

Good work, hopefully we can fix this soon! For now, I won't merge.

NightOwl888:
I have merged this into master along with another patch to the `ThreadClass` to make it act like the `Thread` class in Lucene - that is, it catches any exception thrown by `Run` and re-throws it after calling `Join`. So, we no longer need the hard-coded failure because there is no longer any danger of an exception being thrown on a background thread.

Unfortunately, that wasn't all that was needed to fix these 2 tests (they still fail), but at least now we have failure messages instead of crashes.

PR #223, opened 2019-03-04
Website updates - DOAP file and copy changes

Shazwazza:
Commits some small pending changes that i forgot to previously push and adds the DOAP file to the website repo files so it's included in the build output.

If you want to test, run the powershell site build script: `/websites/site/site.ps1 -ServeDocs 0` then check the output directory `/websites/site/_site` and you'll see the `doap_Lucene_Net.rdf` there.


Shazwazza:
NOTE - the small pending changes are already live on the website, these are not new changes

PR #224, opened 2019-03-04
https://issues.apache.org/jira/browse/LUCENENET-607

hindikaynen:
StringHelper.GOOD_FAST_HASH_SEED property is not thread-safe and could return different values if called in severeal threads in one moment, so it could result in duplicate values in BytesRefHash (same values return different hashes because hashes were calcucated with different seeds).

There is another issue with GOOD_FAST_HASH_SEED. DateTime.Now.Millisecond is used to randomize the seed, but DateTime.Now.Millisecond could return 0 and this value is treated an "uninitialized" and the second GOOD_FAST_HASH_SEED call will return another value.

### DIFF
--- a/src/Lucene.Net/Util/StringHelper.cs
+++ b/src/Lucene.Net/Util/StringHelper.cs
@@ -31,16 +31,15 @@ namespace Lucene.Net.Util
         /// <summary>
         /// Pass this as the seed to <see cref="Murmurhash3_x86_32(byte[], int, int, int)"/>. </summary>
 
+        //Singleton-esque member. Only created once
+        public static readonly int GOOD_FAST_HASH_SEED = InitializeHashSeed();
+
         // Poached from Guava: set a different salt/seed
         // for each JVM instance, to frustrate hash key collision
         // denial of service attacks, and to catch any places that
         // somehow rely on hash function/order across JVM
         // instances:
-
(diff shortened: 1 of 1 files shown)


aaron-meyers:
This is great, we've run into this issue in production with Lucene.Net but hadn't been able to isolate a repro.

One minor comment: when possible, avoid using static constructors and instead initialize static fields directly (using private static helper methods if needed).

e.g. instead of

        public static readonly int GOOD_FAST_HASH_SEED;

        static StringHelper()
        {
            ...
            GOOD_FAST_HASH_SEED = x;
        }

use this pattern:

        public static readonly int GOOD_FAST_HASH_SEED = InitializeHashSeed();

        private static int InitializeHashSeed()
        {
            ...
            return x;
        }

It may look like these are equivalent but in fact they are not. Using the latter pattern gives the CLR more flexibility in how it initializes the class.

More details:
https://stackoverflow.com/questions/10465961/potential-pitfalls-with-static-constructors-in-c-sharp
http://csharpindepth.com/Articles/BeforeFieldInit

hindikaynen:
@aaron-meyers thank you for the review, fixed that.

laimis:
Will do a quick review just to make sure but this looks good at the first glance. Thanks for the PR!

PR #226, opened 2019-06-07
Support.IO.FileSupport::fileCanonPathCache needs to be threadsafe

sthmathew:
Lucene.Net.Support.IO.FileSupport is a static class with a private static readonly dictionary used as a cache. If accessed by multiple threads, when the dictionary is resized requests will start to fail.

Reported [here](https://issues.apache.org/jira/browse/LUCENENET-609)

### DIFF
--- a/src/Lucene.Net/Support/IO/FileSupport.cs
+++ b/src/Lucene.Net/Support/IO/FileSupport.cs
@@ -180,7 +180,7 @@ namespace Lucene.Net.Support.IO
             return Path.Combine(directory.FullName, string.Concat(prefix, randomFileName));
         }
 
-        private static readonly IDictionary<string, string> fileCanonPathCache = new ConcurrentDictionary<string, string>();
+        private static readonly ConcurrentDictionary<string, string> fileCanonPathCache = new ConcurrentDictionary<string, string>();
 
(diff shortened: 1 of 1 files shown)


aaron-meyers:
Minor: it would be a little cleaner to change the type of the fileCanonPathCache field to ConcurrentDictionary and use GetOrAdd rather than the indexer set when populating the fileCanonPathCache in the GetCanonicalPath method (returning the result of GetOrAdd instead of the canonPath value calculated in the method).

I believe the GetCanonicalPath method should always produce the same value anyway but this pattern ensures consistency in the event that a different value was produced for some reason.

sthmathew:
> Minor: it would be a little cleaner to change the type of the fileCanonPathCache field to ConcurrentDictionary and use GetOrAdd rather than the indexer set when populating the fileCanonPathCache in the GetCanonicalPath method (returning the result of GetOrAdd instead of the canonPath value calculated in the method).
> 
> I believe the GetCanonicalPath method should always produce the same value anyway but this pattern ensures consistency in the event that a different value was produced for some reason.

Yeah, I originally didn't want to change too much, so I made the smallest possible change. Updated to use GetOrAdd with the valuefactory now.

PR #230, opened 2019-11-06
Docs updates - inline examples, build to include website and more

Shazwazza:
* Added the ability to show the example code inline for an example code document, for example this one https://lucenenetdocs.azurewebsites.net/api/Lucene.Net.Demo/Lucene.Net.Demo.Facet.AssociationsFacetsExample.html which means the user can just see the example without having to click through to go to GitHub to see the actual code for this file (inception!)
  * TODO: the facet demos all work but not the Lucene.Net.Demo yet, will do this week
* Updatse to latest docfx and latest docfx nuget components
* Fixes `documenation` spelling mistake
* Fixes converted markdown for bold and italics since the previous format doesn't work with bullet points
* Adds the website build to the azure pipelines build, BTW both the docs and the website successfully build now, you can see my custom setup https://shazwazza.visualstudio.com/Lucene.Net%20Build%20Test/_build?definitionId=1&_a=summary
* Re-runs the docs migrator - which is why there's so many changes, these are due to the automation changes above

PR #232, opened 2019-11-13
Documentation: Update formatting of Lucene.Net and port code sample

stevetemple:
1. Currently the code sample won't compile in Lucene.NET it's still the java syntax. I have updated to and tested this in C#
2. Also some formatting had been lost in translation, I have corrected the formatting

### DIFF
--- a/src/Lucene.Net/overview.md
+++ b/src/Lucene.Net/overview.md
@@ -23,38 +23,41 @@ summary: *content
 Apache Lucene.NET is a high-performance, full-featured text search engine library. Here's a simple example how to use Lucene.NET for indexing and searching (Using NUnit for Asserts)
 
 ```
-   Analyzer analyzer = new StandardAnalyzer(LuceneVersion.LUCENE_48);
+Analyzer analyzer = new StandardAnalyzer(LuceneVersion.LUCENE_48);
 
-			// Store the index in memory:
-			Directory directory = new RAMDirectory();
-			// To store an index on disk, use this instead:
-			//Directory directory = FSDirectory.Open("/tmp/testindex");
-			IndexWriterConfig config = new IndexWriterConfig(LuceneVersion.LUCENE_48, analyzer);
-			using (IndexWriter writer = new IndexWriter(directory, config))
-			{
-				Document doc = new Document();
-				String text = "This is the text to be indexed.";
-				doc.Add(new Field("fieldname", text, TextField.TYPE_STORED));
-				writer.AddDocument(doc);
-			}
+// Store the index in memory:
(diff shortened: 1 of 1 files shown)


Shazwazza:
Hi @stevetemple this is great! Like I mentioned in the other PR https://github.com/apache/lucenenet/pull/231 we'll need to find a way specifically to deal with code snippets when running the tool to transform from the Java source to the .net docs. This is on my todo list to figure out the best approach for this. 

I think what it will come down to is being able to parse out their inline code snippets during transformation and then looking for replacement snippets that exist in another file that we manage in our source that replaces it. Or something along those lines. 

That part of the transform process isn't quite in place yet

stevetemple:
I started to look at this.

I'm thinking we could have a folder structure that mirrors the javadoc one, and put files in there let's say codesample-1.cs and codesample-2.cs. We could then swap the first code sample with codesample-1.cs and 2nd with codesample-2.cs etc.

The 2 problems I see with this. The currently Html2Markdown version encodes `pre` as just exact markup rather than a code sample. So there is no way of spotting code samples post conversion. It needs to be done pre/during conversion. The latest Html2Markdown has a way of overriding the behaviour, upgrading to that and having code that jumps in and replaces `<pre class="prettyprint"></pre>` samples should be possible. We could also then put in some sort of markup to indicate this is the original java code sample where a c# version can't be found.

How does that sound? I've seen a comment in the code somewhere saying the latest Html2Markdown breaks some conversion.

Shazwazza:
Hi @stevetemple I think if you would like to make a proof of concept/prototype to see how this could work it will be a good start. We will need to do something so any start on this would be good and then can iterate from there. IIRC the Html2Markdown latest version isn't an issue anymore ... but its been a while since i looked into it. 

I did recently update the docs converter to ensure that bold and italics were done correctly by implementing a custom `IScheme`, you can see the changes here https://github.com/apache/lucenenet/pull/230, the custom `IScheme` is here https://github.com/apache/lucenenet/pull/230/files#diff-45541bf79c42108aca6173ab1c872005

I need to get that PR merged soon... just got too busy recently but should prob just merge

stevetemple:
Well it seems like I've at least solved your TODO! 

I will try and get a prototype together today, I'll base it off the #230 branch.

Shazwazza:
Just posting this link here so its not lost, this was a writeup about an idea of how to possibly handle code blocks going forward https://github.com/apache/lucenenet/pull/229#issuecomment-520831453

I have some other ideas too but nothing concrete at the moment. 

As I make more changes (see current WIP https://github.com/apache/lucenenet/pull/235) i'll try to get my head back into this challenge.

PR #233, opened 2019-12-09
JavaDocToMarkdownConverter: Map Lucene.Net.Search.Join and Grouping correctly, upgrade Html2Markdown

stevetemple:
Also exit when no args passed to prevent exception showing

This should fix the following issues raised in https://github.com/apache/lucenenet/pull/231 :
* Issues with namespaces
* Lists rendering incorrectly


Shazwazza:
Looks good! I'll get a build running and see how it goes

PR #234, opened 2020-01-06
Moving the markdown converter solution to the src/docs folder

Shazwazza:
Since we have other utilities for managing docs that are totally separate from the actual source figured they should be together. Possibly will create a single sln for these two.

Its just moving

`src/dotnet/tools/JavaDocToMarkdownConverter` -> `src/docs/JavaDocToMarkdownConverter`


Shazwazza:
@NightOwl888 you ok with this? it's just moving a proj

NightOwl888:
I don't see an issue. It would be good to have all of the doc build stuff in the same folder.

Shazwazza:
great, will merge (trying to get some of this docs work done this week)

PR #236, opened 2020-01-13
Fix for part of LUCENENET-640 JIRA Issue

Matt2843:
In relation to [LUCENENET-640](https://issues.apache.org/jira/projects/LUCENENET/issues/LUCENENET-640?filter=allopenissues)
Switching IdentityWeakReference from a class inheriting from the WeakReference class to a struct wrapping the GCHandle directly to avoid garbage collection bottleneck.

### DIFF
--- a/src/Lucene.Net/Util/WeakIdentityMap.cs
+++ b/src/Lucene.Net/Util/WeakIdentityMap.cs
@@ -3,6 +3,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using System.Collections;
+using System.Runtime.Serialization;
 
 namespace Lucene.Net.Util
 {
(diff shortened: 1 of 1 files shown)


Matt2843:
Further testing shows that this fix may slow down single Lucene index writing. And the unmanaged resource gcHandle must be freed where neccessarry.

NightOwl888:
Replied @ https://jira.apache.org/jira/projects/LUCENENET/issues/LUCENENET-640

Matt2843:
This is my original code and I License it under Apache License, Version 2.0. Unfortunately I don't have time to follow-up on this issue, but hope that my contributions here have helped in some way nevertheless :)

NightOwl888:
Thanks. Identifying the problem was certainly helpful. We may also need to use this as an interim solution for < .NET Standard 2.1.

NightOwl888:
On second pass, the `ConditionalWeakTable` internally uses `RuntimeHelpers.GetHashCode()` to find the item in the table. Therefore, technically it is a direct replacement for `WeakIdentityMap`.

I have factored out `WeakIdentityMap` and replaced it with `ConditionalWeakTable`. The tests pass, and there doesn't appear to be any memory leaks. Please give it a try and let us know if you are still experiencing problems.

PR #238, opened 2020-02-21
Adds ability to disable caching in DirectoryTaxonomyReader

rohitkumbhar:
`DirectoryTaxonomyReader` depends on `LurchTable` to cache ordinal and
FacetLabel mapping. However, `LurchTable` causes a SIGSEGV on ARM64.
To work around this crash, I've added the ability to disable the use
of cache in DirectoryTaxonomyReader.

Related issue for changing LurchTable (if at all possible): https://github.com/NightOwl888/J2N/issues/3

### DIFF
--- a/src/Lucene.Net.Facet/Taxonomy/Directory/DirectoryTaxonomyReader.cs
+++ b/src/Lucene.Net.Facet/Taxonomy/Directory/DirectoryTaxonomyReader.cs
@@ -66,6 +66,9 @@ namespace Lucene.Net.Facet.Taxonomy.Directory
         // TODO: test DoubleBarrelLRUCache and consider using it instead
         private LRUHashMap<FacetLabel, Int32Class> ordinalCache;
         private LRUHashMap<int, FacetLabel> categoryCache;
+        
+        // Disable cache
+        public bool CacheDisabled { get; set; }
 
         private volatile TaxonomyIndexArrays taxoArrays;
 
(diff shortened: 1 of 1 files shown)


NightOwl888:
Thanks for the report and the PR.

I am not sure if disabling the cache is the right fix for everyone on `Xamarin.iOS`, as that will likely negatively affect performance. On the other hand, maybe such a feature isn't entirely a bad thing, so please leave this PR open for now.

However, since it seems that the fix from [LUCENENET-602](https://issues.apache.org/jira/projects/LUCENENET/issues/LUCENENET-602) isn't working in every case (even after we had confirmation it worked for others), perhaps it is time to take a different approach.

- **Option 1**: There was [a report](https://github.com/csharptest/CSharpTest.Net.Collections/issues/16#issuecomment-564505408) that changing the structs to classes in `LurchTable` will fix the problem. We could create a target for `Xamarin.iOS` on `J2N` and only apply the patch only to that target with conditional compilation.

- **Option 2**: There was [a report](https://issues.apache.org/jira/browse/LUCENENET-602?focusedCommentId=16902515&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16902515) that swapping in `LruCache` instead of `LurchTable` will also work. We could create a target for `Xamarin.iOS` on `Lucene.Net.Facet` and `Lucene.Net.QueryParser` and only apply the patch only to that target with conditional compilation.

If you are willing to do the testing and submit a PR with either of those solutions, it would be much appreciated. Ideally, we would have it setup to run tests on `Xamarin.iOS`, too, but we can probably get by without those tests if you don't have the time to get them working.

Also, if attempting #1, it would be helpful to know whether the fix works without having to target `Xamarin.iOS` directly in `Lucene.Net.Facet` and `Lucene.Net.QueryParser`.

rohitkumbhar:
Hi! Thank you for checking this. As-is, the PR will not disable cache and existing performance will not be affected. I set the `CacheDisabled` to `true` explicitly in my code to avoid the crash. I'll look at the linked options as well.

NightOwl888:
Thanks for the PR. If we have any additional reports of problems with `LurchTable`, I think a better option would be to add constructor overloads of `DirectoryTaxonomyWriter` that accept an abstract factory so the `LruDictionary` implementation can be swapped with a custom subclass. However, since we have not received any other reports of crashes on Xamarin.iOS.

I realize I already mentioned it in https://github.com/NightOwl888/J2N/issues/3, but just reporting it here so it is in the Lucene.NET mail archive for reference. I have also closed #242, which is related.

PR #239, opened 2020-03-31
Minor updates

bongohrtech:
Hi Shad,

I have completed a couple of the tasks that you set: 

1. Replace all private/internal nested IComparer<T> implementations (usually including the word "Anonymous" in the name) with the Comparer.Create() method (see: https://stackoverflow.com/a/28533921). This will eliminate a lot of unnecessary classes and make the code correspond to Java better, which uses anonymous classes to achieve similar code inlining. If you are unsure if one needs to be converted to inline code rather than a concrete class, check the Java implementation to see if it is inline (https://github.com/apache/lucene-solr/tree/releases/lucene-solr/4.8.0/lucen
e).

2. Change overloads of OpenStringBuilder.Append(charsequence, start, end) to OpenStringBuilder.Append(charsequence, startIndex, charCount) to match the conventions in .NET
(https://github.com/apache/lucenenet/blob/1274197c39b4b229af7c6734d35840ff21
d47e97/src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs#L98
). In Java, the convention is to use start index and end index to select text in a string, but in .NET, the corresponding convention is to use start index and length (count).

On building in Azure I noticed that the set BuildNumber step in the Build.ps1 file was failing. This is due to some changes to the Azure pipeline. These variables are now readonly. I commented this line to successfully run the build/test process but you may want to look at it in more detail.

Thanks
Michael

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -129,7 +129,7 @@ stages:
         if (!(Test-Path $dir)) { New-Item -ItemType Directory -Path "$dir" -Force }
         '$(PackageVersion)' | Out-File -FilePath "$dir/$(PackageVersionFileName)" -Force
         '$(FileVersion)' | Out-File -FilePath "$dir/$(FileVersionFileName)" -Force
-        '$(Build.BuildNumber)' | Out-File -FilePath "$dir/$(BuildNumberFileName)" -Force
+      # '$(Build.BuildNumber)' | Out-File -FilePath "$dir/$(BuildNumberFileName)" -Force
       displayName: 'Write Versions to Files'
 
     # If this is a release pipeline, copy the build.bat and Version.props files as version artifacts, which will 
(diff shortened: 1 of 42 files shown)


NightOwl888:
Thanks for the PR.

I have reviewed and will accept this PR, provided you please [submit an ICLA or CCLA to Apache](https://www.apache.org/licenses/contributor-agreements.html#clas). You will only need to submit this document once for this and all future contributions.

I will accept **without any additional changes on your part**, but for future reference please follow these guidelines when submitting pull requests:

1. Please limit PRs to a single focused task, when possible, and provide a more descriptive title on each PR. For example, for these changes I would recommend you submit 3 separate PRs, but given the fact that you were having trouble getting builds to function, what you did here is understandable.
2. Please provide more descriptive commit message headings.

Again, I will fix up the changes (including the failing build process) and merge them from here, no changes are required on your part. Thanks for finding and reporting the Azure DevOps build issue so we can take care of it.

PR #240, opened 2020-04-10
Implemented IAppendable as per J2N 

bongohrtech:
Implement IAppendable on OpenStringBuilder as per J2N

https://github.com/NightOwl888/J2N/blob/master/src/J2N/Text/StringBuffer.cs

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
@@ -26,12 +26,12 @@ namespace Lucene.Net.Analysis.Util
     /// <summary>
     /// A StringBuilder that allows one to access the array.
     /// </summary>
-    public class OpenStringBuilder : ICharSequence
+    public class OpenStringBuilder : IAppendable, ICharSequence
     {
         protected char[] m_buf;
         protected int m_len;
 
-        public OpenStringBuilder() 
+        public OpenStringBuilder()
             : this(32)
         {
         }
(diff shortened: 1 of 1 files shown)


src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
```diff
@@ -270,5 +270,50 @@ public override string ToString()
...
+        }
+
+        public IAppendable Append(char[] value)
+        {
+            return Append(value);
+        }
+
+        public IAppendable Append(char[] value, int startIndex, int count)
+        {
+            return Append(value, startIndex, count);
```
NightOwl888, line 306: This method calls itself. Please make this an explicit `IAppendable` declaration and add another method with the signature `public virtual OpenStringBuilder Append(char[] value, int startIndex, int count)`.

Note that for .NET, since there is no common interface for "character sequences" like there is in Java, we are using explicit overloads of `ICharSequence`, `string`, `StringBuilder`, and `char[]` in most members where just `CharSequence` is a parameter in Java.

src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
```diff
@@ -270,5 +270,50 @@ public override string ToString()
...
+        }
+
+        IAppendable IAppendable.Append(StringBuilder value, int startIndex, int count)
+        {
+            return Append(value, startIndex, count);
+        }
+
+        public IAppendable Append(char[] value)
+        {
+            return Append(value);
```
NightOwl888, line 301: This method calls itself. Please make this an explicit `IAppendable` declaration and add another method with the signature `public virtual OpenStringBuilder Append(char[] value)`.

Note that for .NET, since there is no common interface for "character sequences" like there is in Java, we are using explicit overloads of `ICharSequence`, `string`, `StringBuilder`, and `char[]` in most members where just `CharSequence` is a parameter in Java.

src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
```diff
@@ -270,5 +270,50 @@ public override string ToString()
...
+        }
+
+        IAppendable IAppendable.Append(string value)
+        {
+            return Append(value);
+        }
+
+        IAppendable IAppendable.Append(string value, int startIndex, int count)
```
NightOwl888, line 284: Please update all of the `Append` overloads in this class to consistently use `count` rather than `length`, similar to [`System.Text.StringBuilder`](https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.append?view=netframework-4.8#System_Text_StringBuilder_Append_System_String_System_Int32_System_Int32_)

bongohrtech:
Modified with your changes

PR #241, opened 2020-04-13
https://issues.apache.org/jira/browse/LUCENENET-607

bongohrtech:
modified to be a generic IList<char[]>. Created methods for readability.

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/Ar/ArabicStemmer.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Ar/ArabicStemmer.cs
@@ -1,4 +1,6 @@
 ﻿using Lucene.Net.Analysis.Util;
+using System;
+using System.Collections.Generic;
 
 namespace Lucene.Net.Analysis.Ar
 {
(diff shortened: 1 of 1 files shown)


src/Lucene.Net.Analysis.Common/Analysis/Ar/ArabicStemmer.cs
```diff
@@ -47,9 +47,9 @@ public class ArabicStemmer
         public const char YEH = '\u064A';
 
         // LUCENENET: Avoid static constructors (see https://github.com/apache/lucenenet/pull/224#issuecomment-469284006)
-        public static readonly IList<char[]> prefixes = InitializePrefix();
-        public static readonly IList<char[]> suffixes = InitializeSuffix();
-        
+        public static IList<char[]> Prefixes { get { return InitializePrefix(); } }
```
NightOwl888, line 50: This syntax does not cache the result in a backing field, which is the goal here. Please change to:

```c#
public static IList<char[]> Prefixes { get; } = InitializePrefix();
public static IList<char[]> Suffixes { get; } = InitializeSuffix();
```

bongohrtech: Right! I did wonder, and now I see. Didn't realise this could be done inline. Thanks.

Issue #288, opened 2020-05-15
Fix Random Seed Functionality in TestFramework


NightOwl888:
The random seed that is set to `StringHelper.GOOD_FAST_HASH_SEED` through the `tests.seed` system property in Java currently is the same as the seed that NUnit uses in `NUnit.Framework.TestContext.CurrentContext.Random`. If we continue using NUnit's `Random` property, more research is needed to figure out how to set that seed the way the NUnit designers intended so it can be documented for end users.

During testing, I noticed that NUnit does set the seed to a consistent pseudo-random value, so the same test can be run over and over with exactly the same random setup. However, the seed is re-generated when the code in the project under test is changed. I noted that NUnit writes the seed to a file named `nunit_random_seed.tmp` in the `bin` folder of the test project when run in Visual Studio.

### A Typical Testing Scenario in Lucene

1. Test fails
2. Test failure generates an error message that includes the random seed as a hexadecimal string (that represents a long)
3. Developer sets the `tests.seed` system property to the same hexadecimal string that caused the test to fail
4. Debugging can take place because the pseudo-randomly generated conditions are exactly the same as they were when the test failed
5. A fix is devised

 ### Testing in .NET

All we have been able to do is get the same test to fail multiple times, but that breaks as soon as any code is changed. We could probably revert back if we save a copy of the `nunit_random_seed.tmp`, but this is very complicated to do in addition to debugging, and not very intuitive.

We are missing:

1. The test framework reporting the seed that it was using when the test failed
2. The ability to set the random seed

I suspect that although setting the seed cannot be done publicly, we can hack it by using .NET Reflection to set an internal field. There was some [discussion about setting it through an attribute](https://github.com/nunit/nunit/issues/1461), but it doesn't sound like that has been implemented.

I haven't looked into what it will take to extend NUnit to attach the random seed to the test message that is reported. One option (but not a very good one) would be to change the `Lucene.Net.TestFramework.Assert` class to add the random seed to all of test messages. The main thing we need is to read the seed value. The only place it seems to be exposed is in the `bin/nunit_random_seed.tmp` file, but it would be preferable to read it through a property within the test session.

### Testing Lucene.NET Itself

For end users, the above would solve all of the issues. However, for Lucene.NET we often need to generate the same conditions as Java to determine where the execution paths diverge. This is a problem because the NUnit Random class (probably) isn't the same implementation as the Java Random class. Also, in Java the random seed is a `long`, but in .NET, it is an `int`.

[...]

PR #313, opened 2020-07-21
Fix/random seed simple

bongohrtech:
This addresses a fix for random failures in the TestRandomstrings and TestRandomHugeStrings. It seems that in the tight loop using the same Random variable creates some issues "randomly". This was not repeatable using a seed so my guess is it is due to the threads or some other concurrency within the subfunctions.

I did a lot of work to prove that the outputs from RandomAnalysisString where consistent. The issue was CheckAnalysisConsistency and I didnt see an obvious or even obscure reason why it would fail.

I havent run into any test fails so far but that doesnt mean there isnt one in the next run.

### DIFF
--- a/src/Lucene.Net.TestFramework/Analysis/BaseTokenStreamTestCase.cs
+++ b/src/Lucene.Net.TestFramework/Analysis/BaseTokenStreamTestCase.cs
@@ -883,6 +883,9 @@ namespace Lucene.Net.Analysis
 
             try
             {
+
+                long seed = random.Next();
+
                 for (int i = 0; i < iterations; i++)
                 {
                     string text;
@@ -913,12 +916,12 @@ namespace Lucene.Net.Analysis
                     else
                     {
                         // synthetic
-                        text = TestUtil.RandomAnalysisString(random, maxWordLength, simple);
+                        text = TestUtil.RandomAnalysisString(new Random((int)seed), maxWordLength, simple);
                     }
 
                     try
                     {
-                        CheckAnalysisConsistency(random, a, useCharFilter, text, offsetsAreCorrect, currentField);
+                        CheckAnalysisConsistency(new Random((int)seed), a, useCharFilter, text, offsetsAreCorrect, currentField);
                         if (iw != null)
                         {
                             if (random.Next(7) == 0)
--- a/src/Lucene.Net.TestFramework/Support/Attributes/FindFirstFailingSeedAttribute.cs
+++ b/src/Lucene.Net.TestFramework/Support/Attributes/FindFirstFailingSeedAttribute.cs
@@ -1,0 +1,97 @@
+﻿using System;
+using System.Diagnostics;
+using System.Reflection;
+using System.Threading;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+using NUnit.Framework.Internal.Commands;
+
+[AttributeUsage(AttributeTargets.Method)]
+public sealed class FindFirstFailingSeedAttribute : Attribute, IWrapSetUpTearDown
+{
+    public int StartingSeed { get; set; }
+    public int TimeoutMilliseconds { get; set; } = Timeout.Infinite;
+
+    public TestCommand Wrap(TestCommand command)
+    {
+        return new FindFirstFailingSeedCommand(command, StartingSeed, TimeoutMilliseconds);
+    }
+
+    private sealed class FindFirstFailingSeedCommand : DelegatingTestCommand
+    {
+        private readonly int startingSeed;
+        private readonly int timeoutMilliseconds;
+
+        public FindFirstFailingSeedCommand(TestCommand innerCommand, int startingSeed, int timeoutMilliseconds) : base(innerCommand)
+        {
+            this.startingSeed = startingSeed;
+            this.timeoutMilliseconds = timeoutMilliseconds;
+        }
+
+        public override TestResult Execute(TestExecutionContext context)
+        {
+            var stopwatch = timeoutMilliseconds == Timeout.Infinite ? null : Stopwatch.StartNew();
+
+            for (var seed = startingSeed; ;)
+            {
+                ResetRandomSeed(context, seed);
+                context.CurrentResult = context.CurrentTest.MakeTestResult();
+
+                try
+                {
+                    context.CurrentResult = innerCommand.Execute(context);
+
+                    TestContext.Write("TEST TestExecutionContext SEED: ");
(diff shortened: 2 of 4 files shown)


bongohrtech:
FindFirstFailing and Seed attribute files allow to specify a particular seed to run a test under

NightOwl888:
Its a bit strange, but although this seems to have fixed the `TestRandomStrings` tests for `TestICUFoldingFilter` and `TestThaiAnalyzer`, the `TestThaiAnalzyer::TestRandomHugeStrings()` test still fails. But digging into it, they both terminate in the same place, the only difference is the `maxWordLength` parameter is increased.

I suspect we may have a difference in behavior somewhere in `TestUtil.RandomAnalysisString(Random, int, bool)` that may be causing some rare weirdness. Sadly, `TestUtil` has no tests to verify the behavior is doing what it should be doing.

`TestUtil.RandomAnalysisString()` is also called by `Lucene.Net.Analysis.NGram.EdgeNGramTokenizerTest::TestFullUTF8Range()` and `Lucene.Net.Analysis.NGram.NGramTokenizerTest::TestFullUTF8Range()`, both which are also randomly failing. Perhaps one of the paths that `TestUtil.RandomSubString()` is going down is broken, which would explain the randomness. I suggest to divide and conquer - keep excluding the random paths until you find the one that causes the failure to stop happening. That would probably be a bit quicker than reviewing every one of those methods and comparing them against the Java implementation.

bongohrtech:
> Its a bit strange, but although this seems to have fixed the `TestRandomStrings` tests for `TestICUFoldingFilter` and `TestThaiAnalyzer`, the `TestThaiAnalzyer::TestRandomHugeStrings()` test still fails. But digging into it, they both terminate in the same place, the only difference is the `maxWordLength` parameter is increased.
> 
> I suspect we may have a difference in behavior somewhere in `TestUtil.RandomAnalysisString(Random, int, bool)` that may be causing some rare weirdness. Sadly, `TestUtil` has no tests to verify the behavior is doing what it should be doing.
> 
> `TestUtil.RandomAnalysisString()` is also called by `Lucene.Net.Analysis.NGram.EdgeNGramTokenizerTest::TestFullUTF8Range()` and `Lucene.Net.Analysis.NGram.NGramTokenizerTest::TestFullUTF8Range()`, both which are also randomly failing. Perhaps one of the paths that `TestUtil.RandomSubString()` is going down is broken, which would explain the randomness. I suggest to divide and conquer - keep excluding the random paths until you find the one that causes the failure to stop happening. That would probably be a bit quicker than reviewing every one of those methods and comparing them against the Java implementation.

I am not seeing the failing tests for TestRandomHugeStrings - can you rerun and perhaps send the seed failing? Ive run this several times and havent hit this issue. However, maybe this is because im using the FindFirstFailingSeed attribute - this sets NUnit.Framework.Internal.Randomizer.InitialSeed and NUnit.Framework.Internal.TestExecutionContext.CurrentContext.CurrentTest.Seed to the same seed which are not under normal conditions. 

RandomAnalysisString is producing the same result each time, I verified this by writing the hashed outputs to a file with the iteration and comparing them on subsequent runs. All identical. I can include this code in the pull request if you want to review?

I will take a look at those other tests now (Lucene.Net.Analysis.NGram.EdgeNGramTokenizerTest::TestFullUTF8Range()/Lucene.Net.Analysis.NGram.NGramTokenizerTest::TestFullUTF8Range())

bongohrtech:
Seems both these tests call RandomUnicodeString which is predictably failing on the same seed. This one is solvable i think.

NightOwl888:
> I will take a look at those other tests now (Lucene.Net.Analysis.NGram.EdgeNGramTokenizerTest::TestFullUTF8Range()/Lucene.Net.Analysis.NGram.NGramTokenizerTest::TestFullUTF8Range())

There's no need. This has been patched already in #316.

bongohrtech:
> > I will take a look at those other tests now (Lucene.Net.Analysis.NGram.EdgeNGramTokenizerTest::TestFullUTF8Range()/Lucene.Net.Analysis.NGram.NGramTokenizerTest::TestFullUTF8Range())
> 
> There's no need. This has been patched already in #316.

Ack I wish I had checked my email earlier, this does seem to fix the error. I had found a seed that failed but didnt understand why. I suspected IncrementToken() but clearly not the issue

bongohrtech:
Adding this before the test runs
            NUnit.Framework.Internal.Randomizer.InitialSeed = NUnit.Framework.Internal.TestExecutionContext.CurrentContext.CurrentTest.Seed;

seems to stop the tests erroring. Is this hiding something tho? If there is one random using a different seed, this shouldn't cause an issue.

NightOwl888:
> Adding this before the test runs
NUnit.Framework.Internal.Randomizer.InitialSeed = NUnit.Framework.Internal.TestExecutionContext.CurrentContext.CurrentTest.Seed;

> seems to stop the tests erroring. Is this hiding something tho? If there is one random using a different seed, this shouldn't cause an issue.

Before the test runs how? I tried adding it to the beginning of the test and it didn't seem to make a difference.

I took a deep dive into this and there were several contributors causing these tests to fail. 

- The culture sensitivity was broken by the change of default when we switched to `J2N.Character` in 4.8.0-beta00008 without specifying `CultureInfo.InvariantCulture`. Fixed in #321.
- The `ThaiWordBreaker` wasn't set up to handle surrogate pairs, so would fail when there were surrogate pairs in the random input. Fixed in #322.
- Something is causing `ThaiAnalyzer` to fail when using concurrency. This one I haven't found the cause, but being that the behavior changed when I added a `.Clone()` to each of the prototypes, I suspect the issue lies in `ICU4N.RuleBasedBreakIterator` where it loads the Thai break engine. Adding locking also seems to make it fail less, but it still does occasionally. We need to add some concurrency tests to ICU4N for `BreakIterator` when using Thai to see if we can make it fail - there were no tests for concurrency in ICU4J.

There doesn't appear to be anything wrong with `BaseTokenStreamTestCase`, but when I commented out the concurrency check inside `CheckAnalysisConsistency()`, I ran the `TestRandomHugeStrings()` and `TestRandomStrings() tests 10000 times in a row each without a failure. Putting the concurrency check back in makes the failure happen after a few hundred runs.

NightOwl888:
The concurrency issue with `ICUTokenizer` has now been temporarily addressed in #328.

As for changing the `Random` class instantiation, there doesn't seem to be any effect when doing it this way. I am closing this because it doesn't address the issues described in #288. Of course, we should have a `FindFirstFailingSeed` feature to assist with debugging, but depending on what random number generator implementation we use, it may need to be very different than the one here.

PR #332, opened 2020-08-26
Lucene.Net.Replicator: Fixed an issue in IndexInputStream.Read(...)

jeme:
Lucene.Net.Replicator: Fixed an issue in IndexInputStream.Read(...) that meant the method could return a number larger than what was passed in for read count or what the buffer could hold.

It should instead return the total number of bytes that was read into the buffer, which logically can't be bigger than the buffer it self.

This would make implementations like [Stream.CopyTo](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.copyto?view=netcore-3.1) fail as it relies on the returned number to tell how much to write into the target stream.

### DIFF
--- a/src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
+++ b/src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
@@ -1,0 +1,70 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Lucene.Net.Index;
+using Lucene.Net.Replicator;
+using Lucene.Net.Store;
+using NUnit.Framework;
+
+namespace Lucene.Net.Tests.Replicator
+{
+    /*
+     * Licensed to the Apache Software Foundation (ASF) under one or more
+     * contributor license agreements.  See the NOTICE file distributed with
+     * this work for additional information regarding copyright ownership.
+     * The ASF licenses this file to You under the Apache License, Version 2.0
+     * (the "License"); you may not use this file except in compliance with
+     * the License.  You may obtain a copy of the License at
+     *
+     *     http://www.apache.org/licenses/LICENSE-2.0
+     *
+     * Unless required by applicable law or agreed to in writing, software
+     * distributed under the License is distributed on an "AS IS" BASIS,
+     * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     * See the License for the specific language governing permissions and
+     * limitations under the License.
+     */
+
+    //Note: LUCENENET specific
+    public class IndexInputStreamTest
+    {
+        
+        [Test]
+        public void Read_RemainingIndexInputLargerThanReadCount_ReturnsReadCount()
+        {
+            byte[] buffer = new byte[8.KiloBytes()];
+            new Random().NextBytes(buffer);
+            IndexInputStream stream = new IndexInputStream(new MockIndexInput(buffer));
+
+            int readBytes = 2.KiloBytes();
+            byte[] readBuffer = new byte[readBytes];
+            Assert.That(stream.Read(readBuffer, 0, readBytes), Is.EqualTo(readBytes));
+        }
+
+        [Test]
+        public void Read_RemainingIndexInputLargerThanReadCount_ReturnsExpectedSection([Range(1,8)]int section)
+        {
+            byte[] buffer = new byte[8.KiloBytes()];
+            new Random().NextBytes(buffer);
+            IndexInputStream stream = new IndexInputStream(new MockIndexInput(buffer));
+
+            int readBytes = 1.KiloBytes();
+            byte[] readBuffer = new byte[readBytes];
+            for (int i = section; i > 0; i--)
+                stream.Read(readBuffer, 0, readBytes);
+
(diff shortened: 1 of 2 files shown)


src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
```diff
@@ -0,0 +1,70 @@
...
+     * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     * See the License for the specific language governing permissions and
+     * limitations under the License.
+     */
+
+    //Note: LUCENENET specific
+    public class IndexInputStreamTest
+    {
+        
+        [Test]
```
NightOwl888, line 33: Please add `[LuceneNetSpecific]` attribute from `Lucene.Net.Attributes`.

src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
```diff
@@ -0,0 +1,70 @@
...
+            byte[] buffer = new byte[8.KiloBytes()];
+            new Random().NextBytes(buffer);
+            IndexInputStream stream = new IndexInputStream(new MockIndexInput(buffer));
+
+            int readBytes = 2.KiloBytes();
+            byte[] readBuffer = new byte[readBytes];
+            Assert.That(stream.Read(readBuffer, 0, readBytes), Is.EqualTo(readBytes));
+        }
+
+        [Test]
```
NightOwl888, line 45: Please add `[LuceneNetSpecific]` attribute from `Lucene.Net.Attributes`.

src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
```diff
@@ -0,0 +1,70 @@
...
+            byte[] buffer = new byte[8.KiloBytes()];
+            new Random().NextBytes(buffer);
+            IndexInputStream stream = new IndexInputStream(new MockIndexInput(buffer));
+
+            int readBytes = 1.KiloBytes();
+            byte[] readBuffer = new byte[readBytes];
+            for (int i = section; i > 0; i--)
+                stream.Read(readBuffer, 0, readBytes);
+
+            Assert.That(readBuffer, Is.EqualTo(buffer.Skip((section-1) * readBytes).Take(readBytes).ToArray()));
```
NightOwl888, line 57: Please use `Lucene.Net.TestFramework.Assert.AreEqual()` to make comparisons. NUnit's asserts are extremely slow so we are avoiding them.

src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
```diff
@@ -0,0 +1,70 @@
...
+        [Test]
+        public void Read_RemainingIndexInputLargerThanReadCount_ReturnsReadCount()
+        {
+            byte[] buffer = new byte[8.KiloBytes()];
+            new Random().NextBytes(buffer);
+            IndexInputStream stream = new IndexInputStream(new MockIndexInput(buffer));
+
+            int readBytes = 2.KiloBytes();
+            byte[] readBuffer = new byte[readBytes];
+            Assert.That(stream.Read(readBuffer, 0, readBytes), Is.EqualTo(readBytes));
```
NightOwl888, line 42: Please use `Lucene.Net.TestFramework.Assert.AreEqual()` to make comparisons. NUnit's asserts are extremely slow so we are avoiding them.

NightOwl888 (changes requested):
Thanks for the PR, Jens. I would be happy to accept it after making the requested changes.

src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
jeme: Should this go on the class as well?

src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
jeme: Sounds like something that was worth taking to https://github.com/nunit/nunit as their assertion model is very powerful for making really readable and precise tests, it's a shame to have to obt out :S...

Changed to Lucene's Assert implementations.

src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
NightOwl888: Sure, it is a standard NUnit Category attribute, so that should work.

src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
jeme: **Fixed:** Added `LuceneNetSpecificAttibute`

src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
jeme: **Fixed:** Added `LuceneNetSpecificAttibute`

src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
jeme: **Fixed:** Changed to Lucene's Assert implementations.

NightOwl888 (approved):

Issue #346, opened 2020-09-22
Remove Debugging.AssertsEnabled


theolivenbaum:
Doing some benchmarking here, just saw this strange one on FixedBitSet.cs

@NightOwl888 do you mind if I replace this with a conditional compilation flag?

![image](https://user-images.githubusercontent.com/8791811/93892208-9164b880-fcec-11ea-8d24-a077b8170bfb.png)

theolivenbaum:
Changed here: https://github.com/apache/lucenenet/pull/345/commits/be2e7bdf01d335530c0a38a7ae230e307b383c3c
But happy to port it to a separate PR in case that one has too much for reviewing :)

theolivenbaum:
Changing to a Lazy variable improves it but still quite heavy for that - would be better to fully remove from there, any thoughts @NightOwl888 ?

![image](https://user-images.githubusercontent.com/8791811/93896926-cc1d1f80-fcf1-11ea-97e9-dd801d354053.png)

eladmarg:
this seems expensive, 
I would do this all assertion in conditional compilation debug flag.

@NightOwl888, any reason its not in compilation flag?

theolivenbaum:
The only place I saw this being set so far is under Lucene.Net.Index.CheckIndex, so I added a conditional compilation there too:

https://github.com/apache/lucenenet/pull/345/commits/5a14eaaf8190d81755f48981552ebe16ee046b80#diff-3f750f45c88050fc4fbd6a09ee4c638cR54

But I have unloaded some projects (VS was too slow with everything loaded), so might be missing something...
In any case, unless there is a need for this as a settable value in release mode, probably good to remove it

NightOwl888:
See #326 for an explanation of this. We need this feature for the tests to work right in Release mode and for other features, such as CheckIndex  to work properly.

NightOwl888:
I am open to discussing alternatives to `Lucene.Net.Diagnostics.Debugging.Assert`. However, this feature was added after much consideration about how to actually test Lucene.NET. The main issue is that up until 4.8.0-beta00012, we were skipping test conditions that were driven by asserts, and we discovered 3 new test failures (one of which is still failing) that were missed because we were not running all of the test conditions.

1. In Java asserts can be enabled in the release.
2. The test framework (a component that is available to end users) depends on asserts being enabled in order to run all test conditions.
3. The test framework also catches the `AssertionException` that is thrown by `Debugging.Assert()` in order to ignore it in certain cases.
[...]

PR #347, opened 2020-09-23
Move call to debugging flag out of loops calling FixedBitSet.Get/Set

theolivenbaum:


NightOwl888:
Great. This looks like something that could really reduce the impact of this feature. Although, it would be great to see some benchmarks to understand the impact of this change.

theolivenbaum:
@NightOwl888 did some initial benchmarks on a similar code here: https://github.com/apache/lucenenet/issues/346#issuecomment-697257440

jeme:
Maybe a bit discouraging but in a simple benchmark, this doesn't seem to have any positive impact.

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.1016 (1909/November2018Update/19H2)
Intel Core i7-6700 CPU 3.40GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
  [Host]     : .NET Framework 4.8 (4.8.4200.0), X86 LegacyJIT
  DefaultJob : .NET Framework 4.8 (4.8.4200.0), X86 LegacyJIT


```
|                     Method |       Mean |    Error |   StdDev |
|--------------------------- |-----------:|---------:|---------:|
|                 CacheValue |   320.7 ns |  6.34 ns |  7.30 ns |
|              NonCacheValue |   318.5 ns |  4.74 ns |  4.20 ns |
|    CacheValue_PassToMethod | 6,687.3 ns | 47.91 ns | 40.01 ns |
| NonCacheValue_PassToMethod | 6,640.5 ns | 32.99 ns | 30.86 ns |

----

```csharp
   public class Benchmark
    {


        [Benchmark]
        public void CacheValue()
        {
            int counter = 0;
            bool cached = ValueSource.Value;
            for (int i = 0; i < 1000; i++)
                if (cached)
                    counter++;
        }

        [Benchmark]
        public void NonCacheValue()
        {
            int counter = 0;
            for (int i = 0; i < 1000; i++)
                if (ValueSource.Value)
                    counter++;
        }

        [Benchmark]
        public void CacheValue_PassToMethod()
        {
            bool cached = ValueSource.Value;
            for (int i = 0; i < 100; i++)
                DooWork(cached);
        }

        [Benchmark]
        public void NonCacheValue_PassToMethod()
        {
            for (int i = 0; i < 100; i++)
                DooWork();
        }

        public void DooWork(bool workMore)
        {
            int counter = 0;
            for (int i = 0; i < 100; i++)
                if (workMore)
                    counter++;
        } 
        
        public void DooWork()
        {
            int counter = 0;
            for (int i = 0; i < 100; i++)
                if (ValueSource.Value)
                    counter++;
        }
    }
```

Obviously benchmarks of the actual code could prove to have different results.

Caching the boolean flag using a Lazy seems very reasonable.

src/Lucene.Net/Support/Diagnostics/Debugging.cs
```diff
@@ -26,6 +26,9 @@ namespace Lucene.Net.Diagnostics
     /// </summary>
     internal static class Debugging
     {
+        // LUCENENET specific - use a lazy wrapper around SystemProperties to avoid the repeated cost of getting the value
+        private static Lazy<bool> _assertsEnabled = new Lazy<bool>(() => SystemProperties.GetPropertyAsBoolean("assert", false));
```
jeme, line 30: I don't think this matches the naming convention that Lucene.NET uses, I haven't seen the Underscore prefix used anywhere in the codebase. I know Resharper (if your using that, perhaps Lucene.NET should come with a Resharper config file, it's a rather popular tool after all, although I can't recommend it for the Lucene project due to it's size >.<) suggests this and it's perhaps common in the MS codebases, but not here., @NightOwl888

NightOwl888: Actually, some people used `_assertsEnabled` and other people used `assertsEnabled` for private/internal member variables throughout the project. During the API sweep, they were not normalized as the public/protected variables were, primarily because it isn't as high of a priority. 

> The main focus was to eliminate all of the `PascalCase` member variables, which caused a lot of debugging grief - in some cases a field got called where a property should have or vice versa because of errors during porting. Not to mention all of the extra mental anguish trying to work out what is a property and what is a field while debugging. Not sure why that convention was used, as it isn't a common convention in either Java or .NET. But now all of those fields have been renamed back to `camelCase`.

Currently, there are [tests that use Reflection](https://github.com/apache/lucenenet/blob/ece6bea0a3c98961a77d7060b5615fccefabe725/src/Lucene.Net.Tests.TestFramework/Support/TestApiConsistency.cs) to locate naming and other coding violations, but we probably should aim to use a tool such as Resharper or CodeAnalysis as a replacement at some point. The tool we use should have the ability to make customizations so we can use it to catch Java to .NET conversion errors as well (i.e. `size()` should be converted to `Count` or `Length`, not `Size()` or `Size`).

One convention that is a bit odd and is a special case because of the tendency to make a public `getSomething()` method in Java along with a protected `something` member variable. If we converted this to public `Something` property and protected `something` member variable, we end up with a CLS compliance conflict. As a result, all protected members (whether there was a conflicting property or not) were prefixed with `m_`, since prefixing with `_` is also not CLS compliant. Of course, in .NET it really doesn't make much sense to expose the member variable as protected if there is already a property to access or set it, so we probably could change this at some point as well.

> I started using Resharper recently after I discovered that [JetBrains has a free license program for Apache](https://blog.jetbrains.com/blog/2019/05/30/jetbrains-supports-the-apache-software-foundation/). I used it to change all of the single-lined properties to use expression syntax, but haven't dug down into the settings files yet. No objections if you want to add one as long as the rules don't conflict with our API tests.

NightOwl888:
@jeme 

Based on the [benchmarks you provided](https://github.com/apache/lucenenet/pull/347#issuecomment-698815793) (thank you), I would say it is pretty clear there isn't a benefit to copying the boolean value from the static field into each function.

> Caching the boolean flag using a Lazy seems very reasonable.

Maybe you missed the other benchmarks because they were placed in another thread https://github.com/apache/lucenenet/issues/346#issuecomment-697715665, but the value is being [cached as a static field](https://github.com/apache/lucenenet/blob/ece6bea0a3c98961a77d7060b5615fccefabe725/src/Lucene.Net/Support/Diagnostics/Debugging.cs#L37) already. `SystemProperties` is being lazy-loaded to allow for the user to inject their own `IConfigurationFactory` before it loads. From there, the value is read from the dictionary *once* and cached in a static field the first time `Debugging` is accessed.

Is your recommendation to use `Lazy` also based on the incorrect assumption that `Debugging.AssertsEnabled` reads the value from the dictionary every time it is read, or is there actually some benefit to using `Lazy` in this instance?

### Memory Consumption

What is not clear to me is why adding this feature that is disabled in production produces a 23% bigger memory footprint while indexing than when the feature is compiled out using `Debug.Assert()`. 

In most cases adding this feature is a wash because we are now able to turn off test features in production, but the amount of additional RAM is concerning. I suspect this has something to do with using `Func<string>` as the second parameter of `Debugging.Assert()` - something that was done to prevent strings from being unnecessarily built in cases where the condition parameter is `true`.

There are at least 3 things that could be tried to reduce or eliminate this extra hit:

1. Use string interpolation rather than `+` to build the message string of `Debugging.Assert(bool, Func<string>)`.
2. Change all calls to use the `Debugging.Assert(bool, string)` overload except in 2-3 cases that throw exceptions when the condition is `true`. This will slow down testing for sure, but may improve the situation in production by eliminating the `Func<string>`.
2. Abandon the DRY principle and move the contents of `Debugging.Assert(bool, Func<string>)` inline into each method where it is called (thus eliminating the `Func<string>` variable).

Thoughts?

jeme:
> Is your recommendation to use `Lazy` also based on the incorrect assumption that `Debugging.AssertsEnabled` reads the value from the dictionary every time it is read, or is there actually some benefit to using `Lazy` in this instance?

It was because I miss read the change as it was changed a property, so I read it as a `=>` instead of a `=` - maybe a "brain autocorrect" because I would avoud public fields in all cases, so I was expected it to be a Property before and now it became a cached Property. Seeing as it was a Field before, The change has little value i guess... I would change it to a property in that case though. So:

```
public static bool AssertsEnabled { get; set; }= SystemProperties.GetPropertyAsBoolean("assert", false);
```

> ### Memory Consumption
> What is not clear to me is why adding this feature that is disabled in production produces a 23% bigger memory footprint while indexing than when the feature is compiled out using `Debug.Assert()`.
> 
> In most cases adding this feature is a wash because we are now able to turn off test features in production, but the amount of additional RAM is concerning. I suspect this has something to do with using `Func<string>` as the second parameter of `Debugging.Assert()` - something that was done to prevent strings from being unnecessarily built in cases where the condition parameter is `true`.
> 
> There are at least 3 things that could be tried to reduce or eliminate this extra hit:
> 
> 1. Use string interpolation rather than `+` to build the message string of `Debugging.Assert(bool, Func<string>)`.
> 2. Change all calls to use the `Debugging.Assert(bool, string)` overload except in 2-3 cases that throw exceptions when the condition is `true`. This will slow down testing for sure, but may improve the situation in production by eliminating the `Func<string>`.
> 3. Abandon the DRY principle and move the contents of `Debugging.Assert(bool, Func<string>)` inline into each method where it is called (thus eliminating the `Func<string>` variable).
> 
> Thoughts?

Immediately I would think that there was a certain expense in using a Func for this as compared to alternatives, but that is really just speculations at this point. But I will see if I can find some time to go poke at it a bit...

You said you had access to JetBrains resharper though an Apache free use program, does that include dotMemory? It''s a fairly decent tool for doing memory profiling and it can certainly tell you which objects you have hanging around. It will also allow you to do a run with the flag on, safe a snapshot, then do a run without the flag and compare the two snapshots etc...

(It will be most straight forward if you can make your testing into a executeable you can run from it)

NightOwl888:
> > Is your recommendation to use `Lazy` also based on the incorrect assumption that `Debugging.AssertsEnabled` reads the value from the dictionary every time it is read, or is there actually some benefit to using `Lazy` in this instance?

> It was because I miss read the change as if it was a changed property. As such, I read it as a `=>` instead of a `=` - maybe a "brain autocorrect" because I would avoid public fields in all cases, so I expected it to be a Property before and now it became a cached Property. Seeing as it was a Field before, The change has little value i guess... I would change it to a property in that case though. So:

> ```
> public static bool AssertsEnabled { get; set; }= SystemProperties.GetPropertyAsBoolean("assert", false);
> ```

Actually, the field was by design to make it as cheap as possible to check the value, ensuring that the compiler has no chance to get the optimizations wrong. 

The naming convention was such that it would not need to be changed if it were made into a public property someday.


> You said you had access to JetBrains resharper though an Apache free use program, does that include dotMemory? It''s a fairly decent tool for doing memory profiling and it can certainly tell you which objects you have hanging around. It will also allow you to do a run with the flag on, safe a snapshot, then do a run without the flag and compare the two snapshots etc...

> (It will be most straight forward if you can make your testing into a executeable you can run from it)

Thanks for the suggestion. I do have access to dotMemory, and I will give it a shot when I get a chance (unless you beat me to it).

jeme:
> Actually, the field was by design to make it as cheap as possible to check the value, ensuring that the compiler has no chance to get the optimizations wrong.

I am fairly sure it will always be inlined in this case, so I don't think that is a concern.

> Thanks for the suggestion. I do have access to dotMemory, and I will give it a shot when I get a chance (unless you beat me to it).

Since I don't have a starting point, there is probably no change I will beat you do it :D

theolivenbaum:
@NightOwl888 I think we can close this PR if you merge the other #372 PR that fixes the actual issue that was observed here, that was due to the capturing of a class variable in the lambda...

PR #369, opened 2020-10-16
Add fork me on GitHub

joneskj55:
This PR adds the Fork me on GitHub ribbon to the site. I was unable to to build and test the solution on a local port (I tried several routes) however, I was able to test in the browser. I know that is not the ideal way but I could not get a local port running. Please let me know if this does not suffice and I will continue to try and get a successful build.

### DIFF
--- a/websites/apidocs/index.md
+++ b/websites/apidocs/index.md
@@ -1,3 +1,4 @@
+<span id="forkongithub"><a href="https://github.com/apache/lucenenet">Fork me on GitHub</a></span>
 ---
 title: Lucene.Net Docs - The documentation website for Lucene.Net
 description: The documentation website for Lucene.Net
--- a/websites/site/contributing/community-links.md
+++ b/websites/site/contributing/community-links.md
@@ -1,3 +1,5 @@
+<span id="forkongithub"><a href="https://github.com/apache/lucenenet">Fork me on GitHub</a></span>
+
 ---
 uid: contributing/community-links
 ---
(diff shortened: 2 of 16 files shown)


NightOwl888:
Thanks. The color scheme looks great.

## Build

When running the website build script, it is no longer processing all of the tags correctly so the home page looks like this:

image

DocFx doesn't process markdown files that contain plain HTML (`index.md`, `issue-tracker.md`, etc). I confirmed this by removing the `span` from `index.md`. However, that only fixes the home page. When the `span` is removed from other pages, the ribbon is no longer visible.

I don't have a lot of experience with DocFx, but I believe that instead of putting the `span` tag in Markdown, you need to put it inside a partial file (for example,  `\websites\site\lucenenet\partials\head.tmpl.partial` and `\websites\site\lucenenet\partials\head.tmpl.partial`). I believe in this case you may need to generate an additional `.partial` file to put the HTML into the body of every page, rather than in the header. There is a [quick guide to customizing a docfx site](https://www.cazzulino.com/customize-docfx.html), or if you need more detailed info, I suggest looking at the [DocFx Manual](https://dotnet.github.io/docfx/tutorial/docfx.exe_user_manual.html).

@Shazwazza - Can you provide some additional insights?

## Layout

On the 2 largest resolution viewports, the ribbon overlaps the rightmost menu item when resizing the browser. Could we perhaps add some padding to the right side to keep all of the menu items from being hidden?

image
image

joneskj55:
Ok thanks! I am able to build now so I can actually see my changes. I'll get this fixed and add another commit.

NightOwl888:
Closing, as this was superseded by #374.

Issue #365, opened 2020-10-08
Website: Add "Fork me on GitHub" ribbon


NightOwl888:
To entice more contributions, we need a direct line of marketing from our website to GitHub. Many other open source projects have a "Fork me on GitHub" ribbon on their web site.

The [original "fork me on GitHub" ribbons](https://github.blog/2008-12-19-github-ribbons/) are a bit dated, but there is a more [modern, responsive script generator](https://codepo8.github.io/css-fork-on-github-ribbon/) we can use to add and style a ribbon that matches the theme of our website.

Ideally, this would be on every page of the web site and API documentation. See the [Documentation & Website](https://lucenenet.apache.org/contributing/documentation.html) page for an overview of the site structures and how to build both to test the changes locally.

advaith-unnikrishnan:
I would like to try it.

joneskj55:
I'd like to tackle this.

NightOwl888:
@joneskj55 - Great. I have assigned the issue to you.

PR #374, opened 2020-10-18
Add fork me on github2

joneskj55:
Had to create another branch. This took care of the issue with the markdown files not processing tags correctly. I had to reposition where the `span` was located. The pages are rendering correctly now. Please let me know if this change satisfies the requirement. Also, I have not done the API docs yet. If the solution on this branch works for you I will go ahead and do the same for the API docs. Thanks.


Shazwazza:
Hi @joneskj55 thanks for the effort so far!

We'll want to put the `<span>` within the docfx html template files instead of coded into the .md files so that its just in a single place.  DocFx has a template engine that overrides the default implementation. In the case of this website we don't have an override html page for the normal pages so we'll have to create one. You can export the built in template system to see all of the files used by doing this: https://dotnet.github.io/docfx/tutorial/walkthrough/advanced_walkthrough.html#export-the-default-template. 

I think the template file used to render all of the standard pages is `conceptual.html.primary.tmpl` so you can copy that file in from the exported templates into the `/websites/site/lucenetemplate` and this will 'override' the default implementation. From there you can customize the html however you want so you won't have to copy the `<span>` tag to each .md page (which should only contain markdown) and just have this in one single spot.

Let me know if you have any questions, thanks!

joneskj55:
@Shazwazza will do, thanks!

joneskj55:
Hi @Shazwazza I removed the `<span>` from all of the markdown files. I only needed to add the `<span>` to `index.html.tmpl` (which I did on the previous commit) and the `head.tmpl.partial` file. This took care of adding it to all pages so I didn't need to create a new docfx file. Please let me know if this works as a viable solution.

Shazwazza:
Great stuff, I'm just about to leave on holidays for 3 weeks but i will absolutely get back to this one I return. Thanks for the help so far and sorry for this delay.

NightOwl888 (approved):
Great work! I added a couple of tweaks to add padding to the right side of the menu and to move the `span` to the `navbar.tmpl.partial` file.

I will accept this PR, but you may open another one if you wish to add the link to the API docs to complete #365.

For future reference, please be careful not to change the whitespace. Whitespace doesn't have any functional effect on the site, but makes reviewing the changes in your PR more difficult.

joneskj55:
@NightOwl888 Thanks! Sorry about that whitespace. I noticed that as well and thought it might be an issue. It's from an auto formatter extension that I have in VS Code. I'll make sure to disable it when doing any future work. Thanks again!

NightOwl888:
No worries.

I finally discovered where we are really supposed to put the `<span>` tag - the DocFx default templates have already been exported to [/websites/apidocs/Templates/DefaultTemplateNoAssets/](https://github.com/apache/lucenenet/tree/1fcc714ea2f13e078ba0b32b13392729353e0754/websites/apidocs/Templates/DefaultTemplateNoAssets).

The `/layout/_master.tmpl` file (and its containing folder `layout`) should be copied into the DocFx folder(s). I will do this for the `site`, but for the `apidocs`, it should be copied to:

- `/websites/apidocs/Templates/LuceneApiDocs/layout/_master.tmpl`
- `/websites/apidocs/Templates/LuceneTemplate/layout/_master.tmpl`

From there, you just need to edit the `_master.tmpl` files to put the `<span>` just inside of the `<body>` tag.

joneskj55:
Ok nice, I'll try to get that knocked out today.

PR #376, opened 2020-10-19
Add extension method to avoid boxing IndexOptions, closes #375

### LINKED ISSUE #375: Unnecessary boxing of IndexOptions enum
Found another easy improvement:

There are plenty of calls to IndexOptions.CompareTo(IndexOptions.DOCS_AND_FREQS_AND_POSITIONS) in the code, such as [this one](https://github.com/apache/lucenenet/blob/fc2da940da3ca32c2fe6ae9caf69f36b69f3de7f/src/Lucene.Net/Codecs/BlockTreeTermsReader.cs#L1158):

`````csharp
if (outerInstance.fieldInfo.IndexOptions.CompareTo(IndexOptions.DOCS_AND_FREQS_AND_POSITIONS) < 0)
{
    // Positions were not indexed:
    return null;
}
`````

The problem is that CompareTo receives an object, which leads to boxing the enum:

image

### PR DESCRIPTION
**Pull request by theolivenbaum**

### CODE REVIEW

**Comment by theolivenbaum** [comment]
Fix for #375

**Review by NightOwl888 — CHANGES_REQUESTED** [review]
Good find with discovering the boxing issue. This slipped through because in Java enum types are reference types.

However, let's try to follow .NET conventions rather than using an oddly named `CompareAgainst` extension method. Unfortunately, Microsoft left the boxing in place for the built in `Comparer<T>.Default` when using enums, so the next best thing we can do is build a custom comparer for `IndexOptions`.

```c#
namespace Lucene.Net.Index
{
    internal class IndexOptionsComparer : IComparer<IndexOptions>
    {
        private IndexOptionsComparer() { } // No instance

        public static IComparer<IndexOptions> Default { get; } = new IndexOptionsComparer();

        public int Compare([DisallowNull] IndexOptions x, [DisallowNull] IndexOptions y)
        {
            return ((int)x).CompareTo((int)y);
        }
    }
}
```
Since this class does not exist in Lucene, please place it in the `src/Lucene.Net/Support/Index/` directory. Also, please add XML documentation comments (just copy the documentation format from another .NET comparer and change the wording).

Then the calls can be changed accordingly.

```c#

// Instead of:
public override bool HasFreqs => fieldInfo.IndexOptions.CompareTo(IndexOptions.DOCS_AND_FREQS) >= 0;

// Use:
// LUCENENET specific - to avoid boxing, changed from CompareTo() to IndexOptionsComparer.Compare()
public override bool HasFreqs => IndexOptionsComparer.Default.Compare(fieldInfo.IndexOptions, IndexOptions.DOCS_AND_FREQS) >= 0;

```

theolivenbaum:
That makes the find & replace a bit more challenging :)
But will do the changes and get back to you tomorrow with an updated version!

theolivenbaum:
By the way, just went for the odd name because adding an extension method with the name of the existing enum.CompareTo didn't change anything, the boxing version of the method was still being called...

NightOwl888:
Thanks again for identifying this, but I beat you to it in #384 :).

NightOwl888:
BTW - this is the 3rd boxing issue that has been found recently. Be on the lookout for more.

PR #381, opened 2020-10-24
TaskMergeScheduler error on use _mergethreads Tracking Master

denisisack:
Transform use of IList to BlockingCollection as it's more viable in concurrent scenario as simply doesn't require to have create a new copy so as error related to use of List as it's internal structure change. BlockingCollection allow many features suitable for such concurrent scenario.
lucenetest


NightOwl888:
Thanks for the PR.

Unfortunately, this causes the performance of `TaskMergeScheduler` to become unacceptably slow.

## Without Change

image


## With Change

image

Blocking would seem to defeat the purpose of a concurrent merge scheduler. However, if you would like to take another stab at it, I will leave this PR open. See #354 for a description of what we are looking for.

denisisack:
I have tried understanding somehow might be because of my computing resource but after improving UpdateMergeThreads results at least deterministic as per my limited Computing resource was some convinced it had better result but being limited does provide me less clarity check you may suggest other way or just close PR

denisisack:
lucenetest2

denisisack:
Some test are skipped because of low computing resources but if it doesn't help then you can just close as it of more practical advantage to get big picture, will try checking error remaining to see options presents.

NightOwl888:
Thanks for the attempt. However, this didn't meet the expectation of a high-performance merge scheduler that we were looking for.

That said, you did identify a previously unknown issue with `DocTermOrds.OrdWrappedTermsEnum` (#405) that helped to move us a step in the right direction, and for that we thank you for your contribution.

Issue #362, opened 2020-10-06
API: Lucene.Net.Documents.Document: Add overloads that accept string format and IFormatProvider


NightOwl888:
The [`Document`](https://github.com/apache/lucenenet/blob/f0930a5f8df377cccc428dcfeba02ea4098a6dee/src/Lucene.Net/Document/Document.cs) type has members that get values from fields, but has no way to specify the culture to format culture-sensitive field types into. We simply need to add overloads that accept format and `IFormatProvider` and cascade this info to the `GetStringValue()` method overloads to remedy.

## Current API

```c#
namespace Lucene.Net.Index
{
    public interface IIndexableField
    {
    }
}
namespace Lucene.Net.Documents
{
    public sealed class Document : IEnumerable<IIndexableField>
    {
        public string[] GetValues(string name);
        public string Get(string name);
        public override string ToString();
    }
    public partial class Field : IIndexableField
    {
        public override string ToString();
    }
}
```

## Proposed API

```c#
namespace Lucene.Net.Index
{
    public interface IIndexableField
    {
        string ToString(IFormatProvider provider);
    }
}
namespace Lucene.Net.Documents
{
    public sealed class Document : IEnumerable<IIndexableField>
    {
        public string[] GetValues(string name);
        public string[] GetValues(string name, string format);
        public string[] GetValues(string name, IFormatProvider provider);
        public string[] GetValues(string name, string format, IFormatProvider provider);
        public string Get(string name);
        public string Get(string name, string format);
        public string Get(string name, IFormatProvider provider);
        public string Get(string name, string format, IFormatProvider provider);
        public override string ToString();
        public virtual string ToString(IFormatProvider provider);
    }
    public partial class Field : IIndexableField
    {
        public override string ToString();
        public virtual string ToString(IFormatProvider provider);
    }
}
```

Note that the `GetStringValue()` method is being called twice in both `Get()` and `GetValue()`. Performance can be improved by altering the code to only call it once.

Tests should also be added to verify the functionality.

Semirk0:
Can I take a look at this?

NightOwl888:
Sure. I have assigned the issue to you. Just submit a PR when the changes are complete, and we would be happy to accept it.

PR #385, opened 2020-11-04
Adding overloads to Document for GetValues and Get

Semirk0:
Adding overload for GetValues() and Get() to support in Document. GetStringValue() method in now only being called once.

### DIFF
--- a/src/Lucene.Net.Misc/Document/LazyDocument.cs
+++ b/src/Lucene.Net.Misc/Document/LazyDocument.cs
@@ -135,7 +135,7 @@ namespace Lucene.Net.Documents
         /// <summary>
         /// @lucene.internal 
         /// </summary>
-        public class LazyField : IIndexableField
+        public class LazyField : IIndexableField, IFormattable
         {
             private readonly LazyDocument outerInstance;
 
@@ -355,9 +355,30 @@ namespace Lucene.Net.Documents
                 return GetRealValue().GetTokenStream(analyzer);
             }
 
+            // LUCENENET specific - method added for better .NET compatibility
+            public override string ToString()
+                => ToString(null, J2N.Text.StringFormatter.CurrentCulture);
+
+            // LUCENENET specific - method added for better .NET compatibility
+            public virtual string ToString(string format)
+                => ToString(format, J2N.Text.StringFormatter.CurrentCulture);
+
+            // LUCENENET specific - method added for better .NET compatibility
             public virtual string ToString(IFormatProvider provider)
+                => ToString(null, provider);
+
+            // LUCENENET specific - method added for better .NET compatibility
+            public virtual string ToString(string format,IFormatProvider provider)
             {
-                return GetRealValue().ToString(provider);
+                IIndexableField realValue = GetRealValue();
+                if(realValue is IFormattable formattable)
+                {
+                    return formattable.ToString(format, provider);
+                }
+                else
+                {
+                    return realValue.ToString();
+                }
             }
         }
     }
--- a/src/Lucene.Net.Misc/Document/LazyDocument.cs
+++ b/src/Lucene.Net.Misc/Document/LazyDocument.cs
@@ -357,15 +357,21 @@ namespace Lucene.Net.Documents
 
             // LUCENENET specific - method added for better .NET compatibility
             public override string ToString()
-                => ToString(null, J2N.Text.StringFormatter.CurrentCulture);
+            {
+                return ToString(null, J2N.Text.StringFormatter.CurrentCulture);
+            }
 
             // LUCENENET specific - method added for better .NET compatibility
             public virtual string ToString(string format)
-                => ToString(format, J2N.Text.StringFormatter.CurrentCulture);
+            {
+                return ToString(format, J2N.Text.StringFormatter.CurrentCulture);
+            }
 
             // LUCENENET specific - method added for better .NET compatibility
             public virtual string ToString(IFormatProvider provider)
-                => ToString(null, provider);
+            {
+                return ToString(null, provider);
+            }
 
             // LUCENENET specific - method added for better .NET compatibility
             public virtual string ToString(string format,IFormatProvider provider)
--- a/src/Lucene.Net.Misc/Document/LazyDocument.cs
+++ b/src/Lucene.Net.Misc/Document/LazyDocument.cs
@@ -354,6 +354,11 @@ namespace Lucene.Net.Documents
             {
                 return GetRealValue().GetTokenStream(analyzer);
             }
+
+            public virtual string ToString(IFormatProvider provider)
+            {
+                return GetRealValue().ToString(provider);
+            }
         }
     }
 }
--- a/src/Lucene.Net.Tests/Index/TestIndexWriterExceptions.cs
+++ b/src/Lucene.Net.Tests/Index/TestIndexWriterExceptions.cs
@@ -2043,6 +2043,11 @@ namespace Lucene.Net.Index
             {
                 return null;
             }
+
+            public virtual string ToString(IFormatProvider provider)
+            {
+                return GetStringValue().ToString(provider);
+            }
         }
 
         // See LUCENE-4870 TooManyOpenFiles errors are thrown as
--- a/src/Lucene.Net.Tests/Index/TestIndexWriterExceptions.cs
+++ b/src/Lucene.Net.Tests/Index/TestIndexWriterExceptions.cs
@@ -2043,11 +2043,6 @@ namespace Lucene.Net.Index
             {
                 return null;
             }
-
-            public virtual string ToString(IFormatProvider provider)
-            {
-                return GetStringValue().ToString(provider);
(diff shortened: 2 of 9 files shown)


src/Lucene.Net.Misc/Document/LazyDocument.cs
```diff
@@ -356,6 +356,11 @@ public virtual TokenStream GetTokenStream(Analyzer analyzer)
             {
                 return GetRealValue().GetTokenStream(analyzer);
             }
+
+            public virtual string ToString(IFormatProvider provider)
```
NightOwl888, line 360: Please see my comment about supporting the other `ToString()` overloads here.

src/Lucene.Net.Tests/Index/TestIndexWriterExceptions.cs
```diff
@@ -2043,6 +2043,11 @@ public TokenStream GetTokenStream(Analyzer analyzer)
             {
                 return null;
             }
+
+            public virtual string ToString(IFormatProvider provider)
```
NightOwl888, line 2047: Please see my comment about the leaky abstraction - ultimately this method can be removed after removing `IIndexWriter.ToString(IFormatProvider)`.

src/Lucene.Net.Tests/Index/TestIndexableField.cs
```diff
@@ -225,6 +225,11 @@ public TokenStream GetTokenStream(Analyzer analyzer)
             {
                 return GetReaderValue() != null ? analyzer.GetTokenStream(Name, GetReaderValue()) : analyzer.GetTokenStream(Name, new StringReader(GetStringValue()));
             }
+
+            public virtual string ToString(IFormatProvider provider)
```
NightOwl888, line 229: Please see my comment about the leaky abstraction - ultimately this method can be removed after removing `IIndexWriter.ToString(IFormatProvider)`.

src/Lucene.Net/Index/IndexableField.cs
```diff
@@ -201,5 +201,11 @@ public interface IIndexableField
         ///         a non-null value if the field is to be indexed </returns>
         /// <exception cref="IOException"> Can be thrown while creating the <see cref="TokenStream"/> </exception>
         TokenStream GetTokenStream(Analyzer analyzer);
+
+        /// <summary>
```
NightOwl888, line 205: Please see my comment about the leaky abstraction - ultimately this method can be removed by supporting `IFormattable` instead.

src/Lucene.Net/Document/Field.cs
```diff
@@ -835,6 +835,26 @@ public override string ToString()
...
+        /// Prints a <see cref="Field"/> for human consumption.
+        /// </summary>
+        /// <param name="provider">An object that supplies culture-specific formatting information. This parameter has no effect if this field is non-numeric.</param>
+        public virtual string ToString(IFormatProvider provider)
+        {
+            StringBuilder result = new StringBuilder();
+
+            if(FieldsData != null)
+            {
+                result.AppendFormat(provider, "{0}<{1}:{3}>", m_type.ToString(), m_name.ToString(), FieldsData);
```
NightOwl888, line 848: Note that the index is incorrect for the 3rd format argument, it should be `2`, not `3`.

NightOwl888 (changes requested):
Thanks for the PR. Great work! This is looking good so far, but I have requested a few changes before this can be merged.

While it wasn't part of the original scope, after seeing the impact of this it occurred to me that the user experience can be improved by doing the following:

1. Remove `ToString(IFormatProvider provider)` from `IIndexableField`, as this is a leaky abstraction. The fact this forces `ToString(IFormatProvider)` into the tests is proof of this. Instead, we should allow `IIndexableField` implementations to opt in if they want to support formatting by implementing `IFormattable`. Don't inherit `IFormattable` from `IIndexableField`, allow them to be used independently.
2. We technically only need 1 implementation of `ToString` on each `IIndexableField` type, the rest of the overloads can just cascade their call. See [this example in J2N](https://github.com/NightOwl888/J2N/blob/67369681e1198f27fd1144e24dedf2f8dd6588cb/src/J2N/Collections/Generic/Dictionary.cs#L1239-L1299).
3. Change logic of formattable `Field.ToString()` methods to test for `if (field is IFormattable formattable)` and then call the `formattable.ToString(string, IFormatProvider)` overload. The reasonable default values of the parameters are `null` and `J2N.Text.StringFormatter.CurrentCulture` for those overloads that don't include both. In the case `null` is passed, just use the original string format (in the case of field it was `{0}<{1}:{2}>`). If the formattable check doesn't pass, simply call the default `field.ToString()` overload.
4. Implement `IFormattable` on the built-in `Field` and `LazyField` types. Also add the other 3 overloads `ToString()`, `ToString(string format)` and `ToString(IFormatProvider provider)` so each built-in `IIndexableField` type has all 4 overloads.
5. Change the `Document` class to be similar, implementing `IFormattable` on the built-in types but allowing implementers to opt out of formatting. However, since a format string doesn't make much sense in this case, explicitly implement `IFormattable` (which will hide it from the API of `Document`). Therefore, the `Document` class will still only have 2 public overloads `ToString()` and `ToString(IFormatProvider)` that will cascade their call to an explictly implmented `ToString(string, IFormatProvider)` method.

If you analyze the source of [`StringBuilder.Append`](https://referencesource.microsoft.com/#mscorlib/system/text/stringbuilder.cs,1322), it shows exactly what is expected of how to support `IFormattable`.

I think we should consider supporting `ICustomFormatter` too, but we need to ponder exactly how it should be done. No need to make it part of this PR.

Semirk0:
Can you explain what you mean by 

> 3. Change logic of formattable `Field.ToString()` methods to test for `if (field is IFormattable formattable)`

Do you mean to change logic in class `Field` or `Document`?

Also, I don't really understand 5. either. Especially 

> Therefore, the Document class will still only have 2 public overloads ToString() and ToString(IFormatProvider) that will cascade their call to an explictly implmented ToString(string, IFormatProvider) method.

Can you elaborate?

NightOwl888:
I guess it is simpler just to show the code than to describe it.

We test for `IFormattable` in the if statement to make it an optional feature of implementations of either `IIndexableField` or `FieldsValue`. If the type implements `IFormattable`, we call `IFormattable.ToString(string, IFormatProvider)`. If not, we just skip formatting and call `ToString()` on the object. This makes it possible to use an `IIndexableField` implementation that has no `ToString()` implementations on it or to implement both `IIndexableField` and `IFormattable` to provide localizable formatting support.

### In all implementations of IIndexableField (4 overloads of `ToString()`)

```c#
public override string ToString()
{
    return ToString(null, J2N.Text.StringFormatter.CurrentCulture);
}

public virtual string ToString(string format)
{
    return ToString(format, J2N.Text.StringFormatter.CurrentCulture);
}

public virtual string ToString(IFormatProvider provider)
{
    return ToString(null, provider);
}

public virtual string ToString(string format, IFormatProvider provider)
{
    StringBuilder result = new StringBuilder();
    result.Append(m_type.ToString());
    result.Append('<');
    result.Append(m_name);
    result.Append(':');

    if (FieldsData != null)
    {
        if (FieldsData is IFormattable formattable)
            result.Append(formattable.ToString(format, provider);
        else
            result.Append(FieldsData.ToString());
    }

    result.Append('>');
    return result.ToString();
}
```

### In Document (2 public and one hidden overload of `ToString()`)

```c#

public override string ToString()
{
    return ToString(null, J2N.Text.StringFormatter.CurrentCulture);
}

public virtual string ToString(IFormatProvider provider)
{
    return ToString(null, provider);
}

string IFormattable.ToString(string format, IFormatProvider provider) // <- explicit implementation
{
    var buffer = new StringBuilder();
    buffer.Append("Document<");
    for (int i = 0; i < fields.Count; i++)
    {
        IIndexableField field = fields[i];
        if (field is IFormattable formattable)
            buffer.Append(formattable.ToString(format, provider));
        else
            buffer.Append(field.ToString());
        if (i != fields.Count - 1)
        {
            buffer.Append(" ");
        }
    }
    buffer.Append(">");
    return buffer.ToString();
}

```

Be sure that all built-in `IIndexableField` types, the `Field.Number` type, and `Document` all implement `IFormattable` themselves.

Semirk0:
Is everything ok with my last commit or should I change something?

NightOwl888 (changes requested):
My apologies for not getting to this sooner. I started working on some tests for this (as per #362), but discovered a formatting bug in J2N and then got sidetracked. In Java, floating point numbers are always formatted with at least 1 decimal place, but in .NET the decimal is removed if it is a whole number. In J2N that 1 decimal place is sometimes being truncated.

I'll work on getting the bug fixed and completing the tests, but in the meantime there are a few changes that are requested from the last commit.

NightOwl888 (approved):
I reverted the merge, since it caused a conflict. For future reference please rebase against master rather than merge it.

I also made some minor changes to close #362.

PR #386, opened 2020-11-05
Fixes #296 - Remove Equals() from Lucene.Net.Util.Automaton.State

### LINKED ISSUE #296: IndexOutOfRangeException when searching
Hello, I'm getting an IndexOutOfRangeException when searching in some cases. It's happening maybe 10% of the time so I'm unsure what is causing this. See below for the search code and stack trace. I feel like it might be something with the query but its more of a guess. Any guidance on this would be very much appreciated.

`var sort = new Sort(new SortField(null, SortFieldType.DOC));
return _searcher.Search(Query, _reader.NumDocs, sort);`

`FATAL	 Update Data Set System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Lucene.Net.Store.ByteArrayDataInput.ReadVInt32()
   at Lucene.Net.Codecs.BlockTreeTermsReader.FieldReader.IntersectEnum.Frame.NextLeaf()
   at Lucene.Net.Codecs.BlockTreeTermsReader.FieldReader.IntersectEnum.Next()
   at Lucene.Net.Search.TermCollectingRewrite`1.CollectTerms(IndexReader reader, MultiTermQuery query, TermCollector collector)
   at Lucene.Net.Search.ConstantScoreAutoRewrite.Rewrite(IndexReader reader, MultiTermQuery query)
   at Lucene.Net.Join.ToParentBlockJoinQuery.Rewrite(IndexReader reader)
   at Lucene.Net.Search.IndexSearcher.Rewrite(Query original)
   at Lucene.Net.Search.IndexSearcher.CreateNormalizedWeight(Query query)
   at Lucene.Net.Search.IndexSearcher.Search(Query query, Int32 n, Sort sort)`

NightOwl888:
It took awhile to track it down, but with a working failure project (thanks @wilson556) and confirmation that the problem started between 4.8.0-beta00007 and 4.8.0-beta00008 the problem was traced to https://github.com/apache/lucenenet/commit/e1ead061df6ab5371979040ae8071b1bf8b18070.

It turns out that this was being caused by a subtle change from using reference equality to implementing an `Equals()` method, which was comparing the `id` value. The presumption that the `id` would be unique. It turns out that the unique identifier is the object reference and `id` might have collisions.

Fixes #296.

### DIFF
--- a/src/Lucene.Net/Util/Automaton/State.cs
+++ b/src/Lucene.Net/Util/Automaton/State.cs
@@ -41,7 +41,7 @@ namespace Lucene.Net.Util.Automaton
     /// <para/>
     /// @lucene.experimental
     /// </summary>
-    public class State : IComparable<State>, IEquatable<State> // LUCENENET specific: Implemented IEquatable, since this class is used in hashtables
+    public class State : IComparable<State>
     {
         internal bool accept;
         [WritableArray]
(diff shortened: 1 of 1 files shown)


rclabo:
"The presumption that the id would be unique. It turns out that the unique identifier is the object reference and id might have collisions."  Wow, tracking that down is impressive.  Amazing work.

NightOwl888:
@rclabo - Thanks Rob. It certainly wasn't the faulty end index conversion math I was expecting it to be. Without a reliable red/green indicator and the knowledge that prior versions didn't present the problem, this would have been much more difficult to track down. Fortunately, Git is a wonderful debugging tool if you are armed with that arsenal.

PR #391, opened 2020-11-18
Website updates (from oct 2020)

Shazwazza:
Streamlines some of the build operations:

* Uses powershell flags instead of 0/1
* Updates docs on how to run the scripts including their params
* re-enables our docfx plugins with custom MD parsers so the lucene version is automatically applied throughout the docs
* Ensures the correct base path is used when building the docs for release so that linking works correctly

I have run this locally and the docs build so perhaps it fixes this? https://github.com/apache/lucenenet/issues/380

### DIFF
--- a/src/docs/JavaDocToMarkdownConverter/JavaDocToMarkdownConverter.csproj
+++ b/src/docs/JavaDocToMarkdownConverter/JavaDocToMarkdownConverter.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <!--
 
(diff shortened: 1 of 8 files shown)


NightOwl888:
Thanks. This takes care of #380. 

I added a fix for trimming the trailing slash if the user enters one and also a hack for setting the environment variable, since the plugin doesn't read it from the command line version.

There are still a lot of links that aren't resolving ([example 1](https://lucenenet.apache.org/docs/4.8.0-beta00012/api/analysis-common/overview.html), [example 2](https://lucenenet.apache.org/docs/4.8.0-beta00012/api/facet/package.html), [example 3](https://lucenenet.apache.org/docs/4.8.0-beta00012/api/join/Lucene.Net.Join.html)), but this is not a new issue.

PR #393, opened 2020-11-23
Make Highlighter docs readable

LordMike:
Doesn't fix the docs, but at least puts it in a code block where needed

### DIFF
--- a/src/Lucene.Net.Highlighter/Highlight/package.md
+++ b/src/Lucene.Net.Highlighter/Highlight/package.md
@@ -34,24 +34,24 @@ Fragmenter, fragment Scorer, and Formatter classes.
       QueryParser parser = new QueryParser("notv", analyzer);
       Query query = parser.parse("million");
     
-  TopDocs hits = searcher.search(query, 10);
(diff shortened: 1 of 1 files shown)


NightOwl888:
Thanks for the PR.

This is something that @Shazwazza has been working on, but the way the [API docs are currently deployed](https://lucenenet.apache.org/contributing/documentation.html#api-docs), these changes will be overwritten during the next documentation release.

There is some discussion about this topic in

- https://github.com/apache/lucenenet/pull/231
- https://github.com/apache/lucenenet/pull/229#issuecomment-520831453 (scroll down to the "code samples" heading)

In general, we need to get the `JavaToMarkdownConverter` tool automated to the point we can "freeze" the documentation at Lucene 4.8.1 without the risk of important info being reverted if it is re-run and to flag any things that need to be dealt with manually, and then we will be able to go through and convert all of the code samples and platform-specific information to target C#/.NET.

LordMike:
Ah. I see.. I'll close this then, and let you guys work on a general fix :).

PR #410, opened 2021-02-04
remove facetconfig locks

eladmarg:


eladmarg:
failed because of environment problems

Error: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.

src/Lucene.Net.Facet/FacetsConfig.cs
```diff
@@ -157,17 +153,14 @@ public virtual DimConfig GetDimConfig(string dimName)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public virtual void SetHierarchical(string dimName, bool v)
         {
-            lock (syncLock)
+            // LUCENENET: Eliminated extra lookup by using TryGetValue instead of ContainsKey
+            if (!fieldTypes.TryGetValue(dimName, out DimConfig fieldType))
```
NightOwl888, line 157: After removing the lock, these operations are no longer atomic.

However you are correct that the `ConcurrentDictionary` can be used to replace the lock by calling the appropriate overload of [`AddOrUpdate`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-5.0). Please change to use this `AddOrUpdate` rather than multiple method calls on the `ConcurrentDictionary`.

src/Lucene.Net.Facet/FacetsConfig.cs
```diff
@@ -178,17 +171,14 @@ public virtual void SetHierarchical(string dimName, bool v)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public virtual void SetMultiValued(string dimName, bool v)
         {
-            lock (syncLock)
+            // LUCENENET: Eliminated extra lookup by using TryGetValue instead of ContainsKey
+            if (!fieldTypes.TryGetValue(dimName, out DimConfig fieldType))
```
NightOwl888, line 175: After removing the lock, these operations are no longer atomic.

However you are correct that the `ConcurrentDictionary` can be used to replace the lock by calling the appropriate overload of [`AddOrUpdate`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-5.0). Please change to use this `AddOrUpdate` rather than multiple method calls on the `ConcurrentDictionary`.

src/Lucene.Net.Facet/FacetsConfig.cs
```diff
@@ -200,17 +190,14 @@ public virtual void SetMultiValued(string dimName, bool v)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public virtual void SetRequireDimCount(string dimName, bool v)
         {
-            lock (syncLock)
+            // LUCENENET: Eliminated extra lookup by using TryGetValue instead of ContainsKey
+            if (!fieldTypes.TryGetValue(dimName, out DimConfig fieldType))
```
NightOwl888, line 194: After removing the lock, these operations are no longer atomic.

However you are correct that the `ConcurrentDictionary` can be used to replace the lock by calling the appropriate overload of [`AddOrUpdate`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-5.0). Please change to use this `AddOrUpdate` rather than multiple method calls on the `ConcurrentDictionary`.

src/Lucene.Net.Facet/FacetsConfig.cs
```diff
@@ -221,17 +208,14 @@ public virtual void SetRequireDimCount(string dimName, bool v)
         /// </summary>
         public virtual void SetIndexFieldName(string dimName, string indexFieldName)
         {
-            lock (syncLock)
+            // LUCENENET: Eliminated extra lookup by using TryGetValue instead of ContainsKey
+            if (!fieldTypes.TryGetValue(dimName, out DimConfig fieldType))
```
NightOwl888, line 212: After removing the lock, these operations are no longer atomic.

However you are correct that the `ConcurrentDictionary` can be used to replace the lock by calling the appropriate overload of [`AddOrUpdate`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-5.0). Please change to use this `AddOrUpdate` rather than multiple method calls on the `ConcurrentDictionary`.

NightOwl888 (changes requested):
Thanks for the PR.

However as is, many of the methods are no longer atomic and may cause concurrency issues. Please change the declaration of `fieldTypes` from:

```c#
private readonly IDictionary<string, DimConfig> fieldTypes = new ConcurrentDictionary<string, DimConfig>();
```

to:

```c#
private readonly ConcurrentDictionary<string, DimConfig> fieldTypes = new ConcurrentDictionary<string, DimConfig>();
```

to expose the methods that make `ConcurrentDictionary` atomic and see the other comments about the changes to each of the methods.

src/Lucene.Net.Facet/FacetsConfig.cs
eladmarg: totally agree

NightOwl888:
I got it wrong. Upon closer inspection of the [`AddOrUpdate`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-5.0) documentation:

> If you call AddOrUpdate simultaneously on different threads, addValueFactory **may be called multiple times**, but its key/value pair might not be added to the dictionary for every call.

> For modifications and write operations to the dictionary, ConcurrentDictionary<TKey,TValue> uses fine-grained locking to ensure thread safety. (Read operations on the dictionary are performed in a lock-free manner.) However, the addValueFactory and updateValueFactory delegates are **called outside the locks** to avoid the problems that can arise from executing unknown code under a lock. Therefore, AddOrUpdate is not atomic with regards to all other operations on the ConcurrentDictionary<TKey,TValue> class.

So in other words, this PR is **not** the equivalent functionality of the locks that exist now, as this method does not make the entire block atomic like the lock does.

I am curious, what issues are you actually seeing with these locks?

eladmarg:
> I am curious, what issues are you actually seeing with these locks?

nothing special, just performance penalty on GetTopChildrens method

eladmarg:
any updates with this?

NightOwl888:
@eladmarg 

Given the fact that these locks exist in Lucene and this is not a reasonable locking alternative, we should keep them in place here. I would suggest holding off on this until we investigate https://github.com/apache/lucenenet/issues/417 further, as it is still up in the air whether we should just use `Lazy<T>` or build our own wrapper around `LazyInitializer`. Once we have that solution  (or solutions) worked out, we can apply it to `FacetConfig` as well.

Your last commit, https://github.com/apache/lucenenet/pull/410/commits/0f1894540dd877b47db4a0b86e3ec93cdf2906f4 does contain some useful removal of casts except for 1 call to `Math.Abs`. On [this line](https://github.com/apache/lucenenet/pull/410/commits/0f1894540dd877b47db4a0b86e3ec93cdf2906f4#diff-10b89dfb9b661aa00c8521f46673d331f50fe6ec82a342af7bd58ca7f8b21defR53), this is a bit twiddle that is equivalent to Java's `>>>` operator (unsigned right shift). The cast was done to make it unsigned, however, there is a faster approach that doesn't require a cast (at least not with `int`) is to use [`TripleShift`](https://github.com/NightOwl888/J2N/blob/b24a28da203b55aba860f16d955ce3b1788fcd42/src/J2N/Numerics/BitOperation.cs#L531-L645).

I removed more than 1000 casts last night and have passing tests, but I would like to run some benchmarks to see how removing them compares with the casts in place before submitting the PR. Benchmarking the `TripleShift(this int number, int bits)` method against the two casts shows almost a 20% improvement in performance, and during the tests there seems to be less locking contention.

eladmarg:
cool, the expensive parts are the casts, if you have this on separate PR, this one can be closed.
waiting to see it in action

eladmarg:
closed in #423

PR #415, opened 2021-02-05
Fixed bug that was causing test to fail.

rclabo:


NightOwl888:
Good find! And a bit subtle of a problem. Thanks for looking into this.

Issue #259, opened 2019-10-11
Review tests for Lucene.Net assembly


clambertus:
Occasionally, tests are still being found with important lines that have been commented because the functionality didn't exist at the time they were ported. Additionally, some test conditions have been changed from what they were in Java (although in some cases this is a necessary change due to a difference in platforms and in other cases it is a bug). We need some assurance that none of our green tests are false positives.

A line-by-line review would be best, but at the very least we should be checking that the implementations are complete and the test conditions are the same as in [Java Lucene 4.8.0](https://github.com/apache/lucene-solr/tree/releases/lucene-solr/4.8.0). Checking at the method level to ensure they all exist and have the right attributes has been completed already on Lucene.Net (core).

The tests in question need to be analyzed to ensure:

1. That the test conditions weren't changed from what they were in Lucene (without some very good reason) 
2. That none of the asserts or parts of the test setup were simply commented out to make the test pass rather than fixing the underlying problem
3. (for tests other than Lucene.Net (core)): The test methods/classes are not skipped in Lucene.NET unless they were skipped in Lucene or have some [known problem](https://github.com/apache/lucenenet/issues/269)

There may be other unforeseen issues with the tests (such as an incorrect translation of the line), as well, which is why a line-by-line comparison would be best.

## Some Other Things to Look For

1. Missing `[Test]` attributes. Note that in Lucene tests are run by naming convention for any parameterless method named starting with `test`, but in .NET the attribute is required.
2. Formatting of arrays over multiple lines
3. Stripped comments which would provide needed context
4. Partially implemented classes/tests
5. Method calls such as `Convert.ToString(int)` that require an explicit `CultureInfo.Invariant` parameter to match Java
6. APIs that accept/return object that are using value types (boxing/unboxing)
7. APIs that require casting in order to use them, which is something we should strive to solve
8. APIs that accept/return non-generic collections, which we should strive to make generic

While it isn't the most important part of the task, the tests are also the easiest place to spot usability issues with the API, so if any are discovered (that aren't already marked `LUCENENET TODO`) we should open new issues for them as well.

The test projects that need review are:

* [ ] Lucene.Net.Tests._A-D (See #1018)
* [x] Lucene.Net.Tests._E-I (See #1014)
* [x] Lucene.Net.Tests._I-J (See #926)
* [x] Lucene.Net.Tests._J-S (See #914)
* [x] Lucene.Net.Tests._T-Z (See #897)
[...]

PR #420, opened 2021-02-12
Reviewed classification tests and made minor changes. Part of issue #259

rclabo:
Reviewed classification tests. Made minor changes and uncommented 1 test and verified it passes.  See issue #259

### DIFF
--- a/src/Lucene.Net.Tests.Classification/Utils/DataSplitterTest.cs
+++ b/src/Lucene.Net.Tests.Classification/Utils/DataSplitterTest.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Text;
 using Lucene.Net.Analysis;
 using Lucene.Net.Classification.Utils;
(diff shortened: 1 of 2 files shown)


eladmarg (approved):
Thanks

src/Lucene.Net.Tests.Classification/Utils/DataSplitterTest.cs
```diff
@@ -32,20 +32,20 @@ namespace Lucene.Net.Classification
      */
     public class DataSplitterTest : Util.LuceneTestCase
     {
-        private AtomicReader _originalIndex;
-        private RandomIndexWriter _indexWriter;
+        private AtomicReader originalIndex;
+        private RandomIndexWriter indexWriter;
         private Directory _dir;
```
NightOwl888, line 37: While either fields with or without an underscore are fine (without underscore preferred), we should use the same convention within the context of a single file. Please change this to `dir` as well.

Do note that some classes in `Lucene.Net.Classification` also use a leading underscore. No objection if you wish to change those as well.

NightOwl888 (changes requested):

src/Lucene.Net.Tests.Classification/Utils/DataSplitterTest.cs
rclabo: Of course.  That was a brain fart.  I must have gotten interrupted. Sorry about that.

rclabo: Sounds good. I will clean up Lucene.Net.Classification leading underscores in a separate pull request.

NightOwl888 (approved):

PR #427, opened 2021-02-20
Fixed TestRollingCharBuffer::Test() to prevent OOM when running with Verbose enabled

NightOwl888:


rclabo:
That's a nice simple solution. Very pragmatic.

Issue #300, opened 2020-06-30
Docs cleanup/improvements


Shazwazza:
Based on the resent release of the beta00009 docs, there's still plenty more that needs to be done as the structure of the site had to change due to docfx and xrefs since you cannot have overlapping xrefs within the same docfx site build, they are unique. The changes are now consistent with how the Lucene api docs are structured which is per 'package'.

* [ ] Remove any hard coded values and have these values passed ot the build as part of automation, this means re-enabling some custom docfx plugins and investigating why they conflict with some default xref functionality. 
* [ ] Fix parts of the site that don't load from correct relative (_rel) paths that were needed to be hard coded for now, this includes the index.json search document.
*  [ ] Ensure the main docs site's links link through to the correct 'package' site, for example, Codecs is linking through to the 'Core' package Codecs but it should link through to the Codecs package just like the Lucene docs site does. There are probably others and this is an issue with xrefs.
*  [ ] Investigate and correct xref links that aren't being processed correctly, these are mostly on individual package docs
*  [ ] Investigate warnings during docfx build to try to remove them all
*  [ ] Update site.ps1 build script to be consistent with the docs.ps1 script parameters
*  [ ] Update the documentation for building the website/docs since some of this has changed with the new docfx build
*  [ ] Update the documentation for adding new and 'overwrite' docs to the API documentation
*  [ ] Automate more of the steps involved in creating and deploying the website/docs

paulirwin:
Need to review this to see how many of the original items in this issue are still valid.

Issue #284, opened 2020-05-12
Docs - convert java snippets to c#


Shazwazza:
> Requires: https://github.com/apache/lucenenet/issues/283

> Once we have an idea about how this works https://github.com/apache/lucenenet/issues/283 we can start the snippet conversion. This could be broken down into smaller tasks and listed as up for grabs.

> _~15 hrs_

---------------------------------------------------------------------------------------

This issue no longer depends on #283, since that issue was addressed by #396. So this issue is no longer blocked.

Here is a list of all of the documentation files that will require a review to update code examples. Note that links are a different issue that is being worked on in #300. So in this issue, we should focus on:

1. Updating Java code examples to C#. Make sure the examples will compile the way they are written after the conversion.
2. Updating references to Java technologies (types, JVM, class path, etc) to their .NET equivalents. Some of this will require some research. Note that some things like links over to Java-related wikipedia articles are probably best to leave alone and other things like types that exist in Java but don't in .NET and are still required to make the flow of the docs work might need to be clarified for the average .NET developer.
3. Make sure the formatting of the Markdown document is similar to how it was in Lucene - that is, make sure the numbered and unordered lists come up correctly, tables are legible, etc.
3. Look out for issues with API usability and open issues about them if any are found.

> **NOTE:** The main documents to focus on here are named `overview.md` and `package.md`. The rest could probably just use a cursory review.

- [x] [src\dotnet\Lucene.Net.Replicator.AspNetCore\examples.md](https://github.com/apache/lucenenet/tree/master/src/dotnet/Lucene.Net.Replicator.AspNetCore/examples.md)
- [x] [src\dotnet\tools\lucene-cli\docs\analysis\index.md](https://github.com/apache/lucenenet/tree/master/src/dotnet/tools/lucene-cli/docs/analysis/index.md)
- [x] [src\dotnet\tools\lucene-cli\docs\analysis\kuromoji-build-dictionary.md](https://github.com/apache/lucenenet/tree/master/src/dotnet/tools/lucene-cli/docs/analysis/kuromoji-build-dictionary.md)
- [x] [src\dotnet\tools\lucene-cli\docs\analysis\stempel-compile-stems.md](https://github.com/apache/lucenenet/tree/master/src/dotnet/tools/lucene-cli/docs/analysis/stempel-compile-stems.md)
[...]

PR #451, opened 2021-03-25
Fixes code snippets (#284) and cross-links (#300) in Lucene.Net and Lucene.Net.Analysis.Common (partial)

NightOwl888:
This fixes code snippets, formatting, converts Java references to .NET (#284) in the following docs:

-  src\Lucene.Net\overview.md
-  src\Lucene.Net\Document\package.md
-  src\Lucene.Net\Analysis\package.md
-  src\Lucene.Net.Analysis.Common\overview.md

Also fixed the relative links in the MIGRATE.md file, updated the `docs.ps1` script to update the `toc.yml` with the latest `PackageVersion`, and fixed the copyright symbol on the API docs.

> NOTE: Since some `xref`s were not functioning under certain scenarios, relative links to the HTML documents were put in their place for the time being (see #300 about a fix for this). There were also a few direct links that were added to Microsoft's site for references to .NET types.

### DIFF
--- a/src/Lucene.Net/Analysis/Analyzer.cs
+++ b/src/Lucene.Net/Analysis/Analyzer.cs
@@ -1,4 +1,4 @@
-using Lucene.Net.Util;
+﻿using Lucene.Net.Util;
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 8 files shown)


src/Lucene.Net/Analysis/Analyzer.cs
```diff
@@ -46,22 +46,22 @@ namespace Lucene.Net.Analysis
     /// <para/>
     /// For some concrete implementations bundled with Lucene, look in the analysis modules:
     /// <list type="bullet">
-    ///   <item><description>Common:
+    ///   <item><description>[Common](../analysis-common/overview.html):
```
NightOwl888, line 49: @Shazwazza  - Is this what we should be doing for links in source code to "namespace" documents, or is there a fix in the works so that `<xref:Lucene.Net.Analysis.Common>` would work here?

PR #454, opened 2021-03-26
Documentation work

rclabo:


NightOwl888 (changes requested):
I know some people don't pay much attention to Microsoft's branding, but let's try to be accurate with the terminology to avoid confusion.

> Change `.Net 5.0` to `.NET 5.0`
> Change `.Net Framework` to `.NET Framework`
> Change `NetStandard` to `.NET Standard`
> Change from `Latest Version - Lucene 4.8.0 Beta` to `Latest Version - Lucene.NET 4.8.0 Beta`

> **NOTE:** I have been trying to make the terminology consistent for our name as well. Given it is .NET (all caps) rather than .Net, we'd prefer the name Lucene.NET to Lucene.Net. However, 2 exceptions to this rule are when referring to the namespace `Lucene.Net` and the assembly name `Lucene.Net`.

Also, since in .NET 5 we actually have 7,899 tests, let's change that to 7,800+ passing tests. We may end up shifting some to J2N later or adding some for the Lucene.Net.Support namespace but this should still cover us so this remains accurate without having to update it again prior to the release.

rclabo:
I'm normally pretty sensitive to how brands are communicated but this this case I was asleep at the wheel.  Thanks for catching these and providing great feedback.  I will endeavor to do better. :-)

NightOwl888 (approved):

PR #461, opened 2021-04-02
Fixed 1 GroupingSearch bug and added additional GroupingSearch tests to demonstrate usage

rclabo:


NightOwl888 (changes requested):
Great work on the tests, these are much easier to understand. Do note the tests are failing on Linux - I believe I have spotted the cause, please see my comments.

I also came up with an idea that might work to eliminate casting, but of course we need to separate Field and Function based `GroupingSearch` into their own classes (or own methods) in order to use separate generic constraints for each one.

NightOwl888 (approved):

PR #475, opened 2021-04-22
Improved Grouping Search API and added 8 more grouping tests

rclabo:


src/Lucene.Net.Grouping/GroupingSearch.cs
```diff
@@ -161,86 +196,261 @@ public virtual ITopGroups<object> Search(IndexSearcher searcher, Filter filter,
...
+        /// <param name="searcher">The <see cref="IndexSearcher"/> instance to execute the grouped search on.</param>
+        /// <param name="filter">The filter to execute with the grouping</param>
+        /// <param name="query">The query to execute with the grouping</param>
+        /// <param name="groupOffset">The group offset</param>
+        /// <param name="groupLimit">The number of groups to return from the specified group offset</param>
+        /// <returns>the grouped result as a <see cref="ITopGroups{Object}"/> instance</returns>
+        /// <exception cref="IOException">If any I/O related errors occur</exception>
+        // LUCENENET additional method signature. Makes searching by function easier due to ability to specify type of MutableValue returned.
+        public virtual ITopGroups<TMutableValue> SearchByFunction<TMutableValue>(IndexSearcher searcher, Filter filter, Query query, int groupOffset, int groupLimit)
+            where TMutableValue : MutableValue
```
NightOwl888, line 271: Seems like the generic type parameter should still be named `TGroupValue` to make it clear that is what it is grouping on. The user can still see that the type must be `MutableValue` or a subclass via Intellisense.

src/Lucene.Net.Grouping/GroupingSearch.cs
```diff
@@ -161,86 +196,261 @@ public virtual ITopGroups<object> Search(IndexSearcher searcher, Filter filter,
...
+                searcher.Search(query, filter, cachedCollector);
+            }
+            else
+            {
+                searcher.Search(query, filter, firstRound);
+            }
+
+            if (allGroups)
+            {
+                matchingGroups = (ICollection)allGroupsCollector.Groups;
```
NightOwl888, line 365: Is there a way we can make this cast use a generic collection?

I am hoping to eliminate the use of non-generic collections in Lucene.NET. There are 2 reasons for this:

1. Custom user collections probably won't bother implementing non-generic interfaces, so casts like this aren't guaranteed to succeed.
2. J2N doesn't currently support comparing/equating non-generic collections.

NightOwl888 (approved):
I am approving this, but have a couple of comments for you to review.

I added some commits to make this PR current and to fix some minor issues.

It looks like you are well on the road to eliminating the need for the covariant interfaces, although we probably should keep `ICollector`. `ITopGroups<T>` could potentially be eliminated, though.

src/Lucene.Net.Grouping/GroupingSearch.cs
rclabo: I think TMutableValue is more clear since the T passed must inherit from MutableValue.  But if you feel strongly about it I can change it.

src/Lucene.Net.Grouping/GroupingSearch.cs
rclabo: I think so.  I could also probably convert the  matchingGroups member var from a `ICollection` to a `ICollection<TGroupValue>` if that's preferred for the reasons you cited. It would mean that [ICollection GetAllMatchingGroups()](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Grouping/GroupingSearch.cs#L707) would need to be removed but it appears to only be used by one test which can easily be upgraded to use the generic version of the same method.  So I can make those changes if that sounds good.

src/Lucene.Net.Grouping/GroupingSearch.cs
NightOwl888: What type it is is clear, however, its purpose (that this is the type it is grouping on) is not. While we could make this clear by adding a [`<typeparam>`](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/typeparam) value in the documentation, I still think that naming the generic parameter after its purpose rather than its type is important.

src/Lucene.Net.Grouping/GroupingSearch.cs
NightOwl888: Yes, that sounds good.

src/Lucene.Net.Grouping/GroupingSearch.cs
rclabo: I yield.  Yep, will clean it up.

PR #488, opened 2021-05-12
Fix SonarQube's "Any() should be used to test for emptiness" / Code Smell

marodev:
Hello,

This pull request is _very simple_ and aims to reduce the technical debt. It specifically targets the "Any() should be used to test for emptiness" [rule](https://rules.sonarsource.com/csharp/RSPEC-1155) in SonarQube.

Let me know if you are interested in similar contributions.

Best,
Martin

### DIFF
--- a/src/Lucene.Net.Grouping/AbstractSecondPassGroupingCollector.cs
+++ b/src/Lucene.Net.Grouping/AbstractSecondPassGroupingCollector.cs
@@ -54,7 +54,7 @@ namespace Lucene.Net.Search.Grouping
         {
 
             //System.out.println("SP init");
-            if (groups.Count() == 0)
+            if (!groups.Any())
             {
                 throw new ArgumentException("no groups to collect (groups.Count is 0)");
             }
--- a/src/Lucene.Net.Grouping/AbstractSecondPassGroupingCollector.cs
+++ b/src/Lucene.Net.Grouping/AbstractSecondPassGroupingCollector.cs
@@ -54,7 +54,7 @@ namespace Lucene.Net.Search.Grouping
         {
 
             //System.out.println("SP init");
-            if (!groups.Any())
+            if (!groups.Any()) // LUCENENET TODO: Change back to .Count if/when IEnumerable<T> is changed to ICollection<T> or IReadOnlyCollection<T>
             {
                 throw new ArgumentException("no groups to collect (groups.Count is 0)");
             }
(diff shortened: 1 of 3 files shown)


marodev:
Hi @NightOwl888, may I ask you to review this pull request?

NightOwl888 (approved):
Thanks for the PR.

Since this is a line-by-line port of [Java Lucene 4.8.0](https://github.com/apache/lucene/tree/releases/lucene-solr/4.8.0), our technical debt is a bit different than a regular C# project. Basically, the further away from the original source we drift, the more technical debt we have accrued.

However, that doesn't mean we shouldn't diverge from Java in cases where performance and/or usability is a factor. So, we would be willing to accept more PRs like this one provided they are applied sensibly with regard to how far from the original Java code we diverge vs how much benefit the change provides. Of course, if the change involves a lot of lines of code, it would be better to have a conversation about it on the [dev mailing list](https://github.com/apache/lucenenet#start-a-discussion) before sinking a lot of time into it.

Given the fact that `Lucene.Net.Tests.Cli` was code that was originally written by Microsoft, we don't have an issue with the change. 

In the case of `AbstractSecondPassGroupingCollector`, in Java the `.size()` method was used, which corresponds to the `Count` property in .NET. The fact that the LINQ `Count()` method was used here is a symptom of a bigger problem - namely, since the `GroupingSearch` class was generalized so much in Lucene we had to resort to covariance to get it to compile in .NET. `ICollection<T>` is not a covariant interface, so we had to change to `IEnumerable<T>`, which doesn't expose `Count` as a property.

Ideally, we hope to fix this in a way that doesn't allocate an enumerator. There are 2 possibilities:

1. Change `IEnumerable<T>` to [`IReadOnlyCollection<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlycollection-1?view=net-5.0) to expose the `Count` property.
2. Factor out the generalized [`GroupingSearch.Search()` overloads](https://github.com/apache/lucenenet/blob/07f9e29a8cb1d9f178e6bded1e5afdfcc850e65b/src/Lucene.Net.Grouping/GroupingSearch.cs#L101-L196) and use more specific `SearchByXXX()` methods exclusively, in which case we can revert back to using the non-covariant [`ICollection<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.icollection-1?view=net-5.0) interface and strong types.

@rclabo has made significant progress toward the second option in #461 and #475 by adding some of the specific methods we would need to remove the need for the `Search()` method overloads.

Since moving to `.Any()` provides a real performance benefit for `AbstractSecondPassGroupingCollector` in its current form, I will accept the PR as is, but I have added comments to indicate that the `Count` property should be used instead once it is available.

PR #490, opened 2021-05-27
Improved debugger experience for BytesRef

rclabo:
Improved debugger experience for BytesRef.  In addition to the decimal bytes values it now shows the BytesRef as a Utf8String.  If the BytesRef is not a Utf8 string that representation will just be gibberish.


rclabo:
This should be all ready to go.  When I submitted this initially in a new branch I forgot to including the necessary using statement which is why it failed _all_ the tests.  But I added that in the 2nd commit so it's all happy now.  (I realized my mistake about .5 seconds after I submitted the PR. Dooh!)  Anyway, all is well that ends well...right?

NightOwl888 (approved):

PR #498, opened 2021-07-24
Lucene.Net.Support: Factored out Number class in favor of using J2N's parsers and formatters

NightOwl888:


NightOwl888:
Superseded by #504

PR #513, opened 2021-08-18
Two additional ControlledRealTimeReopenThread tests

rclabo:
Added two additional demonstration tests for ControlledRealTimeReopenThread that make it easier to see how to use the class.  One of the tests, TestMultithreadedWaitForGeneration, demonstrates that the current implemention of the ControlledRealTimeReopenThread class is flawed when used in multi-thread situations. This test will fail currently but I will then add another commit that improves the ControlledRealTimeReopenThread class so that it's a more faithful port of the java version and then the test will pass.

### DIFF
--- a/src/Lucene.Net/Search/ControlledRealTimeReopenThread.cs
+++ b/src/Lucene.Net/Search/ControlledRealTimeReopenThread.cs
@@ -42,6 +42,7 @@ namespace Lucene.Net.Search
     public class ControlledRealTimeReopenThread<T> : ThreadJob, IDisposable
          where T : class
     {
+        // LUCENENET: java final converted readonly
         private readonly ReferenceManager<T> manager;
         private readonly long targetMaxStaleNS;
         private readonly long targetMinStaleNS;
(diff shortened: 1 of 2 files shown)


src/Lucene.Net/Search/ControlledRealTimeReopenThread.cs
```diff
@@ -105,45 +108,38 @@ private void RefreshDone()
...
 
         /// <summary>
-        /// Releases all resources used by the <see cref="ControlledRealTimeReopenThread{T}"/>.
+        /// Kills the thread and releases all resources used by the
+        /// <see cref="ControlledRealTimeReopenThread{T}"/>. Also joins to the
+        /// thread so that when this method returns the thread is no longer alive.
         /// </summary>
         public void Dispose()
         {
-            Dispose(true);
```
NightOwl888, line 120: This class is not sealed. We should keep the dispose pattern in place so inheritors have a way to dispose any unmanaged resources that they own and so the finalizer doesn't automatically run.

It would also be wise to keep track of whether `Dispose(bool)` has been called so it is safe to call it more than once. See https://stackoverflow.com/a/5306896. Also, please include tests for multiple-dispose.

NightOwl888 (changes requested):

src/Lucene.Net/Search/ControlledRealTimeReopenThread.cs
rclabo: Ok, I'll add back a Dispose(bool) method.

PR #520, opened 2021-09-15
Fixed build after Microsoft deprecated .NET Core 2.1 and removed the SDK download

NightOwl888:
Changed test target for `netstandard2.0` from `netcoreapp2.1` to `net461`. This involved many changes to the project structure to pass the `<SetTargetFramework>` element to each project under test so it can be used to override the default dependency behavior of `net461` during testing.

The test framework for `Lucene.Net.Tests.CodeAnalysis` was also changed from `netcoreapp2.1` to `net5.0` for the same reason.

The `Generate-TestWorkflows.ps1` script was also modified to generate GitHub workflows that account for the change and the workflows were regenerated.

### DIFF
--- a/.github/workflows/Generate-TestWorkflows.ps1
+++ b/.github/workflows/Generate-TestWorkflows.ps1
@@ -38,7 +38,7 @@
 
  .PARAMETER TestFrameworks
     A string array of Dotnet target framework monikers to run the tests on. The default is
-    @('net5.0','netcoreapp2.1','net48').
+    @('net5.0','net461','net48').
 
  .PARAMETER OperatingSystems
     A string array of Github Actions operating system monikers to run the tests on.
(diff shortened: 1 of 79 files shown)


rclabo:
Awesome work Shad!  One questions, if a project is targeting `net461` why also target `net4.8`?

NightOwl888:
@rclabo 

`net48` is to test the "natural" project loading (that is, to test the `net45` target).

`net461` would naturally launch the `net45` target. However, we are intervening by using the `SetTargetFramework` element to force it to test the `netstandard2.0` target. Without intervening, there would be no way to test `netstandard2.0` because one of the other targets would load (unless we use an older mobile framework for testing). Admittedly, now that .NET Core 2.x is not supported, the amount of additional coverage `netstandard2.0` gains is tiny, but it is not very difficult to continue supporting it.

PR #533, opened 2021-10-26
Website build automation

Shazwazza:
NOTE: This should be squash merged.

* Updates website build action to automatically extract the latest tagged lucene version or use a specific formatted tag. The update also allows for this to run on the main repo or on any fork (for testing). The site trigger will be when any file in `- 'websites/site/**'` is changed or when there's a tag formatted like: `Website_4_8_0_beta00013`
* Creates a new docs build actions, fixes the docfx plugin build and begins attempting to fully automate the docs building process. This isn't fully working yet! But it's almost there. This merge won't break the current docs building process. The site trigger will be when any file in `'websites/apidocs/**'`, `'src/docs/LuceneDocsPlugins/**'` is changed or when there's a tag formatted like: `Docs_4_8_0_beta00013`

I will test if this task is complete and working in our repo too https://issues.apache.org/jira/browse/INFRA-21453.

### DIFF
--- a/.github/workflows/Lucene-Net-Website.yml
+++ b/.github/workflows/Lucene-Net-Website.yml
@@ -34,7 +34,7 @@ on:
     #branches:
     #- master
     paths:
-    - 'websites/site/**/*'
+    - 'websites/site/**'
 
 env:
   # If a tag is specified, the tag will be in the format: Website_4_8_0_beta00013 which


.github/workflows/Lucene-Net-Website.yml
```diff
@@ -15,95 +15,107 @@
...
-    tags:
-      - Website_*
+  workflow_dispatch:
   push:
+    tags:
+    - Website_*
     #branches:
     #- master
     paths:
-    - 'websites/site/**/*'
```
NightOwl888, line 37: @Shazwazza - Why was this changed from from detecting changes on subdirectories? Ron did a commit yesterday on the website, but it didn't kick off the deployment. This seems like the likely culprit. Is there a specific reason behind setting the glob pattern up like this, or if this is a bug that needs to be addressed?

Ditto on the `Lucene-Net-Documentation.yml` file.

Shazwazza: Because according to the GitHub documentation, the double ** will match sub directories. For example https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-using-positive-and-negative-patterns-1

NightOwl888: That makes sense. But, apparently the documentation is wrong. Maybe we could run a test, but I know for sure that the `/**/*` glob syntax works because it is used in the other templates: https://github.com/apache/lucenenet/blob/master/.github/workflows/Lucene-Net-Tests-Analysis-Common.yml#L44.

Shazwazza: Yep the intention was to include sub directories so prob just best to change it back to what actually works :)

NightOwl888: Agreed. I am finding more and more that for build automation the more explicit the syntax, the better. At the very least, it makes the intention clear even if it doesn't function.

NightOwl888: I have updated the paths in #597, and submitted #598 as a test. But the automation still didn't kick off. 

The first thing I thought of was maybe I committed the `.yml` file with invalid encoding, but it is UTF-8 w/BOM, same as the other templates.

Being that #598 didn't kick off the automation the first time, the commit can be re-used in additional PRs until you get it to function, as it contains a change we want on the website.

PR #578, opened 2021-12-14
Lucene.Net.Facet: Added culture-sensitve ToString() overload on FacetResult and LabelAndValue

NightOwl888:
Added overload of `ToString()` that accepts `IFormatProvider` to `FacetResult` and `LabelAndValue` classes. Also implemented `IFormattable` on both classes.

The default format has changed from invariant culture to current culture to match .NET.

### DIFF
--- a/src/Lucene.Net.Facet/FacetResult.cs
+++ b/src/Lucene.Net.Facet/FacetResult.cs
@@ -4,6 +4,7 @@ using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Text;
+#nullable enable
 
 namespace Lucene.Net.Facet
 {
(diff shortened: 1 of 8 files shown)


src/Lucene.Net.Facet/FacetResult.cs
```diff
@@ -4,6 +4,7 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Text;
+#nullable enable
```
eladmarg, line 7: why this is needed?

NightOwl888: This is just to enabled the [nullable reference type](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references) feature. It allows us to make it explicit that `IFormatProvider` accepts a `null` value (which causes it to use the current culture).

I am slowly going through and adding this (rather inconsistently) instead of making a separate task. Since the project is large, enabling this feature at the project level isn't practical. While this doesn't seem to make a lot of sense on small files, enabling this on larger files does help to identify potential `null` reference bugs so they can be avoided. The whole `J2N` library was converted to use this and it turned out to be helpful.

PR #596, opened 2022-01-08
Closes #364 - Added Cross-Platform Build Script

### LINKED ISSUE #364: Task: Make cross-OS build script
We currently have a `.bat` script file to launch the build from the command line that wraps a Powershell file. Clearly, since `.bat` is Windows-only, we will need to make a `.sh` script file in order to be able to build on other operating systems.

The [actual `build.bat` file](https://github.com/apache/lucenenet/blob/f0930a5f8df377cccc428dcfeba02ea4098a6dee/build.bat) does nothing more than read command line parameters (in any order they are entered), setting defaults where appropriate, and then launch Powershell with the PSake module.

While we could provide instructions to launch the Powershell script using when not running on Windows, the command structure would be simpler for end users if they had a similar `build.sh` wrapper script with the same inputs as the `build.bat` file (and if it is possible to use the `.sh` file on Windows, simply eliminate the `.bat` file).

Note that the `build.ps` script also [generates a replacement `build.bat` file](https://github.com/apache/lucenenet/blob/f0930a5f8df377cccc428dcfeba02ea4098a6dee/build/build.ps1#L583-L663) that eliminates the `version` and `packageVersion` parameters when building a release. This ensures a [release distribution](https://lucenenet.apache.org/download/download.html)'s `build.bat` file will always build the current release version unless the user manually updates the `version.props` file that is also in the release distribution. `build.ps` will need to be updated to generate a `build.sh` file during a release for the same reasons.

NightOwl888:
Closes #364.

- Restructured bootstrap to use 3 scripts
  1. `build.bat` (Windows Batch File)
  2. `build` (BASH script)
  3. `build.ps1` (Powershell script to parse the arguments, import the PSake module, and call `runbuild.ps1`)
- Upgraded PSake to 4.9.0
- Renamed `build` folder `.build` to avoid naming collisions on Linux
- Renamed `build/build.ps1` > `.build/runbuild.ps1`
- Added `PrintTargetFrameworks` target to root `Directory.Build.targets` file so the build can check which targets are configured that are valid on the current OS.
- Modified `runbuild.ps1` to call `PrintTargetFrameworks` during testing to eliminate hard-coded special cases.
- Updated `.gitattributes` to check out batch and BASH files with the correct line endings.
- Updated `runbuild.ps1` to track files that were generated so they can are removed if the build is cancelled (files that were changed were already being reverted, but files that were added were being left behind)
- Modified `.github/workflows/Generate-TestWorkflows.ps1` to remove hard-coded special cases when specifying target frameworks.
- Added `install-dotnet-sdk.yml` template to disable telemetry and install the .NET SDK during both Build and Test stages.
- ~~Removed installation of legacy SDKs from both Azure DevOps and GitHub Actions.~~ (Reverted - these are required)
- Updated `README.md` and other documentation to reflect changes.
- Confirmed command line build and test both function on Ubuntu 20.04.
- Removed `dotnet-install.ps1` script.
- Fixed Release stage of `azure-pipelines.yml` so we don't distribute the `websites` folder or `.github` folder in the release.


rclabo:
Wow, look like good work.  There is still so much about build automation I don't understand.  I think I will watch some youtube videos about GitHub Actions today and try to raise my knowledge on this a bit.

PR #599, opened 2022-01-10
docs: Repo README page edits, and two minor web page content edits.

rclabo:
Added gray boxes around images on Contributor "Setup Java Debugging" web page.
Fixed "lines of code" indicated on Contributor Source Code web page and linked to related image.
Added welcome and more overview content, plus minor edits to repo README page

### DIFF
--- a/websites/site/contributing/how-to-setup-java-lucene-debugging.md
+++ b/websites/site/contributing/how-to-setup-java-lucene-debugging.md
@@ -1,5 +1,5 @@
 ﻿---
-uid: setupJavaDebugging
+uid: contributing/setup-java-debugging
 ---
 
 # How to Setup Java Lucene 4.8 Debugging
--- a/websites/site/contributing/how-to-setup-java-lucene-debugging.md
+++ b/websites/site/contributing/how-to-setup-java-lucene-debugging.md
@@ -1,4 +1,8 @@
-﻿# How to Setup Java Lucene 4.8 Debugging
+﻿---
+uid: setupJavaDebugging
+---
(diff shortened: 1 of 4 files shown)


websites/site/contributing/how-to-setup-java-lucene-debugging.md
```diff
@@ -1,4 +1,8 @@
-﻿# How to Setup Java Lucene 4.8 Debugging
+﻿---
+uid: setupJavaDebugging
```
NightOwl888, line 2: I could be wrong, but I believe the convention to use here are all lowercase characters with dashes between them.

```
uid: setup-java-debugging
```

NightOwl888: Actually, this is true, but the others also include the contributing folder:

```
uid: contributing/make-release
```

So, to follow the convention:

```
uid: contributing/setup-java-debugging
```

rclabo: Good catch. It should be consistent for sure and I hadn't realized other pages already used the uid feature.

I wish the standard chosen wasn't so verbose.  It's a bit overkill to have such fully qualified uids on a site so small. But such an approach does accommodate a lot of growth for the site longer term.  Sadly it results in overly verbose css targeting and the main reason I needed the uid was for css targeting. 

```
article[data-uid="contributing/setup-java-debugging"] img {
    border: solid 1px #dedcd3
}
```
But it works.  I'll make the change

Issue #604, opened 2022-01-20
The given key 'Lucene.Net.Support.WeakDictionary`2+WeakKey`1[System.Object,System.Collections.Generic.IDictionary`2[Lucene.Net.Search.FieldCacheImpl+CacheKey,System.Object],System.Object]' was not present in the dictionary.


hamzashoukat94:
```

 Sort sort = new Sort(new SortField("Category2", SortFieldType.STRING_VAL));
 TopFieldDocs fields = searcher.Search(query, filter, TOTAL_ITEMS, sort, true, false);


```
I'm executing the above mentioned code. This is giving me exception stating that 
`The given key 'Lucene.Net.Support.WeakDictionary`2+WeakKey`1[System.Object,System.Collections.Generic.IDictionary`2[Lucene.Net.Search.FieldCacheImpl+CacheKey,System.Object],System.Object]' was not present in the dictionary.`

I debugged the LuceneNet 4.8 code. And I found the particular faulty code is in WeakDictionary.cs line no 156:165.
```
  public TValue this[TKey key]
        {
            get => _hm[new WeakKey<TKey>(key)]; 
            set
            {
                CleanIfNeeded();
                _hm[new WeakKey<TKey>(key)] = value;
            }
        }

```
Kindly update the issue asap.

NightOwl888:
Thanks for the report.

Could you please provide a complete example that we can run to reproduce the issue?

`WeakDictionary` doesn't tell us anything other than that you are using .NET Framework. Unfortunately, Microsoft didn't expose the enumerator of `ConditionalWeakTable` until .NET Standard 2.1, so we are left without a reasonable alternative. `WeakDictionary` is just a workaround for .NET Framework.

Do note that it is not supposed to keep references internally for any objects that are put in it. You have to hold onto the references from outside of the application (assuming that you are adding them explicitly), otherwise there is no guarantee that they will be retrievable.

hamzashoukat94:
Thanks for the quick response and I'm expecting quick response till further. 
Steps to reproduce it:
1. Initialize Lucene Directory and IndexWriter.
2. Build Index e.g 1000 Documents.
3. Run the below mention code. 
4. In result of step 3, you will get the reported issue. 
```
               if (!analyzer.GetType().Equals(new WhitespaceAnalyzer(version).GetType()))
                    filterTerm = filterTerm.ToLower();

                Filter filter = new QueryWrapperFilter(new TermQuery(new Term(fieldName, filterTerm)));

[...]

PR #613, opened 2022-01-31
Factored out WeakDictionary in favor of weak events using Prism.Core

### LINKED ISSUE #256: Port ConditionalWeakTable from .NET Core 3.x to .NET Standard 2.0
<p>As per <a href="https://issues.apache.org/jira/browse/LUCENENET-610" title="Reduce locking in FieldCacheImpl::Cache::Get" class="issue-link" data-issue-key="LUCENENET-610">LUCENENET-610</a>, the WeakDictionary that has been created to support FieldCache and a few other Lucene.NET features does not perform well enough in highly concurrent environments.</p>

<p>The <a href="https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2" class="external-link" rel="nofollow">ConditionalWeakTable<TKey, TValue></a> is a suitable replacement on .NET Standard 2.1, but on .NET Standard 2.0, it doesn't expose the enumerator or the <tt>AddOrUpdate</tt> method. All of Lucene.NET's usages require one or the other.</p>

<p>So, it would definitely be worth the effort to port the full implementation to .NET Standard 2.0.</p>

<p>This port should be added to <a href="https://github.com/NightOwl888/J2N" class="external-link" rel="nofollow">J2N</a> in the <tt>J2N.Runtime.CompilerServices</tt> namespace (to match .NET's namespace convention).</p>
<i>JIRA link - <a href="https://issues.apache.org/jira/browse/LUCENENET-636">[LUCENENET-636]</a> created by nightowl888</i>

NightOwl888:
Closes #256, #604, and #605.

This removes the `Lucene.Net.Support.WeakDictionary<TKey, TValue>` class (and its tests) that were carried over from Lucene.Net 3.0.3, and completely replaces it on .NET Standard 2.0 and .NET Framework with `ConditionalWeakTable<TKey, TValue>`. Since the enumerator was not available on .NET Standard 2.0 and prior, this workaround uses weak events from [Prism.Core](https://github.com/PrismLibrary/Prism) to either call methods or return strong references to types that are inside of the `ConditionalWeakTable<TKey, TValue>`.

This adds an `AddOrUpdate` extension method (non-threadsafe) for cases where a "put" is required.

Also adds `Lucene.Net.Util.ExcludeFromRamUsageEstimationAttribute` to mark fields so they don't get included in the calculation (which prevents infinite recursion on the event fields).


rclabo:
WOW! You are amazing!

laimis:
I second the WOW! comment, can't believe you find a solution for this...

eladmarg:
Even though it's not relevant for me (I'm using net core for years), 
I'm really curious, did this solve the memory leak issue?

NightOwl888:
@eladmarg 

Not sure. Being that `ConditionalWeakTable` requires the key and values to be dereferenced before it will clean them up, this poses a problem to be able to check whether they actually still exist (because they cannot be looked up without those references). So, maybe the only way to check is to profile/benchmark the app to see if it makes a difference, but other suggestions are welcome.

However, testing on .NET Framework seems to have settled at around 1-2 min longer than .NET 5, where before it varied and was for the most part around 14 minutes instead of around 11-12 minutes.

NightOwl888:
I have altered this a bit:

1. The events were migrated to a new `Lucene.Net.Util.Events` internal static class.
2. Removed the `GetRamBytesUsedEvent` and `GetDocIdSetsEvent` and reused the `GetCacheKeysEvent` in these cases, looking up the values from the `ConditionalWeakTable` based on the keys. While this makes the event functionality a bit slower, it avoids putting `DocIdSet` and `CachedOrds` objects into the finalizer queue and shifts the responsibility of cleaning up events to `IndexReader.Dispose()`. This method is almost certainly guaranteed to be called by the consumer, which takes it out of the finalizer queue.
3. Changed the event tracking to allow multiple instances of the same event to be attached to an `IndexReader` instance, which solves any potential limitations if the same `IndexReader` instance is involved in 2 or more different contexts that require events.

So, this basically avoids any finalizer calls to enable the weak event functionality.

But even if that were not so, avoiding the [huge overhead of re-allocating a Dictionary repeatedly](https://github.com/apache/lucenenet/blob/41ffbc7b54cd5427a87e63c8bc3c4c3c290e1161/src/Lucene.Net/Support/WeakDictionary.cs#L63-L89) by relying only on `ConditionalWeakTable` and weak events is certainly more efficient than `WeakDictionary`.

PR #620, opened 2022-02-14
PERFORMANCE: Lucene.Net.Index.BaseCompositeReader: Removed unnecessary list allocation

NightOwl888:
Removed allocation of new `List<T>` that was added during the port to handle generic casting. Lucene [didn't have this extra allocation](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/index/BaseCompositeReader.java#L69).

### DIFF
--- a/src/Lucene.Net/Index/BaseCompositeReader.cs
+++ b/src/Lucene.Net/Index/BaseCompositeReader.cs
@@ -1,7 +1,6 @@
 ﻿using J2N.Collections.Generic.Extensions;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


eladmarg:
Great catch! 
Any chance to push this to the upcoming release?

NightOwl888:
> Great catch! Any chance to push this to the upcoming release?

Unfortunately, not. The binaries have already been built. But then, like any release there is always one more feature that would be nice to have. :point_up:

PR #624, opened 2022-03-07
Update StandardQueryParser.cs

rauhs:
Note: The doc string is currently incorrect. There is no property `Analyzer` on the `QueryConfigHandler`.

### DIFF
--- a/src/Lucene.Net.QueryParser/Flexible/Standard/StandardQueryParser.cs
+++ b/src/Lucene.Net.QueryParser/Flexible/Standard/StandardQueryParser.cs
@@ -47,7 +47,8 @@ namespace Lucene.Net.QueryParsers.Flexible.Standard
     /// <code>
     /// // the query config handler returned by StandardQueryParser is a
     /// // StandardQueryConfigHandler
-    /// queryParserHelper.QueryConfigHandler.Analyzer = new WhitespaceAnalyzer();
+    /// queryParserHelper.Analyzer = new WhitespaceAnalyzer();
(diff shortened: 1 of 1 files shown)


NightOwl888 (changes requested):
While setting `AllowLeadingWildcard` [wasn't in the docs of Lucene 4.8.1](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/standard/StandardQueryParser.java#L61), [it is set in the most recent documentation](https://github.com/apache/lucene/blob/f725b27e12a992f12aab12a8e8bfb9b31125bdae/lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/standard/StandardQueryParser.java#L57), so this makes sense.

However, in Java the docs appear to be wrong, since there are no setters on `StandardQueryConfigHandler`. Not sure what to make of this, but since the setters exist on `StandardQueryParser` it shouldn't be an issue. It looks like the `QueryConfigHandler` property is just there to allow `StandardQueryParser` subclasses to add their own config settings that are passed through the pipeline.

That being said, a shortcut for

```c#
queryParserHelper.QueryConfigHandler.Set(ConfigurationKeys.ALLOW_LEADING_WILDCARD, true);
```

is to use

```c#
queryParserHelper.AllowLeadingWildcard = true;
```

which would be the preferred way to show in the documentation.

NightOwl888:
Since people tend to copy code examples it would be best to show the "common" way. That being said, we could optionally add a note that there is an alternative syntax just below the example.

NightOwl888 (approved):
Thanks for the PR!

PR #634, opened 2022-05-16
Add DebuggerDisplay for SegToken

BrandonStudio:
It is so hard to debug without a friendly display. I just add a `DebuggerDisplay` for `SegToken`, you can also add one for `SegTokenPair` and others.
In C# 10, the attribute should be 
```csharp
[System.Diagnostics.DebuggerDisplay($"\{new string({nameof(CharArray)})\}")]
```

You can also override the `ToString` method instead.


NightOwl888 (changes requested):
Thanks for the PR.

Please add the missing closing parenthesis for `DebuggerDisplay`.

You are welcome to add a `DebuggerDisplay` also for `SegTokenPair` if you feel it will be helpful, but do note that class is marked internal.

NightOwl888 (approved):

PR #571, opened 2021-12-11
Add PersianStemmer

raminmjj:

PR #636, opened 2022-05-22
Lucene.Net.Analysis.Fa/PersianAnalyzer: Removed PersianStemmer as per apache/lucene#904

NightOwl888:
Reverted changes from #571 as was done in apache/lucene#904. Changed `TestPersianStemFilter` to use mocks.


mocobeta:
We still have `PersianStemmer` in `PersianAnalyzer` on the `main` branch, it's a great addition - on the other hand, we have to consider backward compatibility on a major release (not to force users to reindex the whole data) so we can't expose the new stemmer for the default analyzer in 9.x. We will definitely have it in PerianAnalyzer in Lucene 10.0...

PR #650, opened 2022-10-13
Remove NET45, NET451, NET452 Support & update website framework ver

rclabo:
+Removed NET45, NET451, and NET452 support. These are no longer supported by Microsoft.  
+Changed the website project from Framework 4.0  to Framework 4.6.2 to allow for easier loading in VS2022.


.github/workflows/Generate-TestWorkflows.ps1
```diff
@@ -65,7 +65,7 @@ param(
 
     [string]$RepoRoot = (Split-Path (Split-Path $PSScriptRoot)),
 
-    [string[]]$TestFrameworks = @('net6.0', 'net5.0','net461','net48'), # targets under test: net6.0, netstandard2.1, netstanard2.0, net45
+    [string[]]$TestFrameworks = @('net6.0', 'net5.0','net461','net48'), # targets under test: net6.0, netstandard2.1, netstanard2.0, net48
```
NightOwl888, line 68: The framework under test in this case would be net462

Lucene.Net.sln
```diff
@@ -263,7 +263,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Lucene.Net.Tests.AllProject
 EndProject
 Project("{E24C65DC-7377-472B-9ABA-BC803B73C61A}") = "websites", "websites\", "{8988CDA4-8420-4BEE-869A-66825055EED2}"
 	ProjectSection(WebsiteProperties) = preProject
-		TargetFrameworkMoniker = ".NETFramework,Version%3Dv4.0"
+		TargetFrameworkMoniker = ".NETFramework,Version%3Dv4.6.2"
```
NightOwl888, line 266: This issue only affects contributors who open the project in Visual Studio 2022. We should probably just bump this to 4.8, since that framework is more likely to be installed.

rclabo: Initially I thought the same thing but then resolved that that chasing the latest version doesn't make sense. Further, the solution produces Framework 4.6.2 assemblies and 4.6.2 is not yet depreciated, so I decided it made more sense to use 4.6.2 for the website project.

nikcio: But shouldn't the website be updated as much as possible without too much effort to avoid doing it again too soon in the future? This would maybe also make it easier for new contributes to make contributions? (Just a thought)😅

nikcio:
@NightOwl888 @rclabo I've noticed that the only project still targeting a lower version than `net462` is `Lucene.Net.Replicator.AspNetCore` targeting  `net461` https://github.com/apache/lucenenet/blob/dfae964ce0a1b066fd808275fd81d90385cde3fe/src/dotnet/Lucene.Net.Replicator.AspNetCore/Lucene.Net.Replicator.AspNetCore.csproj#L28

As every project was updated to `net462` in this PR would it then make sense to have `Lucene.Net.Replicator.AspNetCore` targeting  `net462`  instead?

rclabo:
@nikcio  I initially planned to remove Net461 and ran into a snag.  It turns out that to support one of the NetStandard targets we use Net461 for that build and more importantly need it to be a different framework version then any other build so that we can sense that NetStandand build by triggering on the targeted framework.  Doing something different was a bit of a rathole so I left it.

NightOwl888:
@nikcio - Thanks for pointing that out. I just opened #747, since there is quite a bit to figuring out what to target and whether or not we need another package to include extension methods for wiring this up in AspNetCore across .NET Core and .NET Framework.

This is one of the few areas in the project that could be considered "new" development.

I suspect that `net461` was picked here because one of the dependencies didn't have support for any target other than `netstandard2.0`, so we couldn't support anything lower than that.

PR #655, opened 2022-10-14
fix: Aligned ReplaceFrom with the Java code (FrenchStemmer.cs)

nikcio:
[File: FrenchStemmer.cs]

Java reference: https://github.com/apache/lucene/blob/f01152a5909fa6059f4f1d4aeb4e14968ef1d8c2/lucene/analysis/common/src/java/org/apache/lucene/analysis/fr/FrenchStemmer.java#L486-L502

I'm unsure if this is covered by unit tests.


nikcio:
Looking into the approach commented in #654 instead.

Issue #663, opened 2022-10-15
Review for CA1822: Mark members as static


NightOwl888:
https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&id=nikcio_lucenenet&open=AYPAuPfOhbfJOGLOobHM
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1822

Members that do not access instance data or call instance methods can be marked static. This may diverge from the Java implementation, so we should mark these members with the comment `// LUCENENET: CA1822: Mark members as static`.

A check should be made against the Lucene source code first before applying this change. If there is no `final` modifier on a method in Java...

```java
public void DoSomething() {
   // impl
}
```

...it means it should be declared `virtual` in .NET (unless the class is sealed).

```c#
public virtual void DoSomething()
{
   // impl
}
```

In this case, we cannot declare the method/property `static` and instead add the missing `virtual` modifier.

> `**NOTE:** The entire codebase has been manually reviewed already to catch the virtual/non-virtual declarations, however, since this review was manual, there may have been some that were missed.

nikcio:
Similar issues can be found here: https://sonarcloud.io/project/issues?resolved=false&rules=external_roslyn%3ACA1822&severities=INFO&id=nikcio_lucenenet

nikcio:
I'll look into this one

NightOwl888:
Great work!

PR #690, opened 2022-10-16
Review for CA1822: Mark members as static

nikcio:
#663 

This should target almost all of the issues currently reported in #663 but I think there will come more of the same issues after this is merged and scanned due to all the new methods now being static. So don't close the issue just yet.

It's mostly just adding `static` and a comment. But some places I've also changed the code to work with the new static methods. I should only have changed the code in some tests.

I've also added `virtual` to methods that I found had the `final` keyword in the Java source code.

### DIFF
--- a/src/Lucene.Net.TestFramework/Support/Util/LuceneRandomSeedInitializer.cs
+++ b/src/Lucene.Net.TestFramework/Support/Util/LuceneRandomSeedInitializer.cs
@@ -44,7 +44,7 @@ namespace Lucene.Net.Util
         /// <param name="seed">The random seed for a new <see cref="Random"/> instance.
         /// Note this is a subclass of <see cref="Random"/>, since the default doesn't produce consistent results across platforms.</param>
         /// <returns><c>true</c> if the seed was found in context; <c>false</c> if the seed was generated.</returns>
-        private bool TryGetRandomSeedsFromContext(Test test, out long seed, out long? testSeed)
+        private static bool TryGetRandomSeedsFromContext(Test test, out long seed, out long? testSeed) // LUCENENET: CA1822: Mark members as static
         {
             //bool generate;
             seed = default;
--- a/src/Lucene.Net.TestFramework/Support/Util/LuceneRandomSeedInitializer.cs
+++ b/src/Lucene.Net.TestFramework/Support/Util/LuceneRandomSeedInitializer.cs
@@ -44,7 +44,7 @@ namespace Lucene.Net.Util
         /// <param name="seed">The random seed for a new <see cref="Random"/> instance.
         /// Note this is a subclass of <see cref="Random"/>, since the default doesn't produce consistent results across platforms.</param>
         /// <returns><c>true</c> if the seed was found in context; <c>false</c> if the seed was generated.</returns>
-        private static bool TryGetRandomSeedsFromContext(Test test, out long seed, out long? testSeed) // LUCENENET: CA1822: Mark members as static
+        private static bool TryGetRandomSeedsFromContext(Test test, out long seed, out long? testSeed)
         {
             //bool generate;
             seed = default;
--- a/src/Lucene.Net.TestFramework/Support/Util/LuceneTestFrameworkInitializer.cs
+++ b/src/Lucene.Net.TestFramework/Support/Util/LuceneTestFrameworkInitializer.cs
@@ -276,7 +276,7 @@ namespace Lucene.Net.Util
         /// // tight loop with many invocations.
         /// </code>
         /// </summary>
-        protected static Random Random => LuceneTestCase.Random; // LUCENENET: CA1822: Mark members as static
+        protected Random Random => LuceneTestCase.Random;
 
         /// <summary>
         /// Overridden in a derived class, provides a way to set <see cref="CodecFactory"/>,
--- a/src/Lucene.Net.TestFramework/Support/Util/LuceneTestFrameworkInitializer.cs
+++ b/src/Lucene.Net.TestFramework/Support/Util/LuceneTestFrameworkInitializer.cs
@@ -276,7 +276,7 @@ namespace Lucene.Net.Util
         /// // tight loop with many invocations.
         /// </code>
         /// </summary>
-        protected Random Random => LuceneTestCase.Random;
+        protected static Random Random => LuceneTestCase.Random; // LUCENENET: CA1822: Mark members as static
 
         /// <summary>
         /// Overridden in a derived class, provides a way to set <see cref="CodecFactory"/>,
--- a/src/Lucene.Net.Tests.Analysis.Phonetic/Language/MatchRatingApproachEncoderTest.cs
+++ b/src/Lucene.Net.Tests.Analysis.Phonetic/Language/MatchRatingApproachEncoderTest.cs
@@ -90,19 +90,19 @@ namespace Lucene.Net.Analysis.Phonetic.Language
         [Test]
         public void TestRemoveSingleDoubleConsonants_BUBLE_RemovedSuccessfully()
         {
-            Assert.AreEqual("BUBLE", this.StringEncoder.RemoveDoubleConsonants("BUBBLE"));
+            Assert.AreEqual("BUBLE", MatchRatingApproachEncoder.RemoveDoubleConsonants("BUBBLE"));
         }
 
         [Test]
         public void TestRemoveDoubleConsonants_MISSISSIPPI_RemovedSuccessfully()
         {
-            Assert.AreEqual("MISISIPI", this.StringEncoder.RemoveDoubleConsonants("MISSISSIPPI"));
+            Assert.AreEqual("MISISIPI", MatchRatingApproachEncoder.RemoveDoubleConsonants("MISSISSIPPI"));
         }
 
         [Test]
         public void TestRemoveDoubleDoubleVowel_BEETLE_NotRemoved()
         {
-            Assert.AreEqual("BEETLE", this.StringEncoder.RemoveDoubleConsonants("BEETLE"));
+            Assert.AreEqual("BEETLE", MatchRatingApproachEncoder.RemoveDoubleConsonants("BEETLE"));
         }
 
         [Test]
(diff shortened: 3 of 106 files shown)


src/Lucene.Net.TestFramework/Support/Util/LuceneRandomSeedInitializer.cs
```diff
@@ -44,7 +44,7 @@ internal class LuceneRandomSeedInitializer
         /// <param name="seed">The random seed for a new <see cref="Random"/> instance.
         /// Note this is a subclass of <see cref="Random"/>, since the default doesn't produce consistent results across platforms.</param>
         /// <returns><c>true</c> if the seed was found in context; <c>false</c> if the seed was generated.</returns>
-        private bool TryGetRandomSeedsFromContext(Test test, out long seed, out long? testSeed)
+        private static bool TryGetRandomSeedsFromContext(Test test, out long seed, out long? testSeed) // LUCENENET: CA1822: Mark members as static
```
NightOwl888, line 47: No need for a comment in a Support folder. At least for stuff that we created as a shim to make Lucene work. But do be aware that there are 2 packages in Benchmark and one in Phonetic that we have made part of the assembly instead of creating separate NuGet packages and repositories to maintain.

src/Lucene.Net.Tests.Analysis.Phonetic/Language/MatchRatingApproachEncoderTest.cs
```diff
@@ -90,19 +90,19 @@ public void TestAccentRemoval_NullValue_ReturnNullSuccessfully()
         [Test]
         public void TestRemoveSingleDoubleConsonants_BUBLE_RemovedSuccessfully()
         {
-            Assert.AreEqual("BUBLE", this.StringEncoder.RemoveDoubleConsonants("BUBBLE"));
+            Assert.AreEqual("BUBLE", MatchRatingApproachEncoder.RemoveDoubleConsonants("BUBBLE"));
```
NightOwl888, line 93: Encoder instances are meant to be interchangeable, so we need to keep these methods available through `this.StringEncoder`, not as a static methods. I am referring to all of the methods under test in this class.

https://github.com/apache/commons-codec/blob/1.9/src/test/java/org/apache/commons/codec/language/MatchRatingApproachEncoderTest.java#L87

src/Lucene.Net/Index/DocumentsWriterFlushQueue.cs
```diff
@@ -299,7 +299,7 @@ protected FlushTicket(FrozenBufferedUpdates frozenUpdates)
                 indexWriter.PublishFlushedSegment(newSegment.segmentInfo, segmentUpdates, globalPacket);
             }
 
-            protected void FinishFlush(IndexWriter indexWriter, FlushedSegment newSegment, FrozenBufferedUpdates bufferedUpdates)
+            protected virtual void FinishFlush(IndexWriter indexWriter, FlushedSegment newSegment, FrozenBufferedUpdates bufferedUpdates)
```
NightOwl888, line 302: This is marked final in Lucene 4.8.0, so should not be made virtual. Also, it is more sensible as an instance method.

```java
protected final void finishFlush(IndexWriter indexWriter, FlushedSegment newSegment, FrozenBufferedUpdates bufferedUpdates)
            throws IOException {
```

src/Lucene.Net/Index/FreqProxTermsWriterPerField.cs
```diff
@@ -387,7 +387,7 @@ internal override int BytesPerPosting()
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        public void Abort()
+        public static void Abort()
```
NightOwl888, line 390: This method was not marked final in Lucene 4.8.0, therefore it should be made `virtual`.

```java
  public void abort() {}
```

src/Lucene.Net.TestFramework/Support/Util/LuceneTestFrameworkInitializer.cs
```diff
@@ -276,7 +276,7 @@ internal void DoTestFrameworkTearDown()
         /// // tight loop with many invocations.
         /// </code>
         /// </summary>
-        protected Random Random => LuceneTestCase.Random;
+        protected static Random Random => LuceneTestCase.Random; // LUCENENET: CA1822: Mark members as static
```
NightOwl888, line 279: This is just meant to be a shortcut so we don't have to access the static `LuceneTestCase.Random` method in a subclass of `LuceneTestFrameworkInitializer`. It should be an instance property, and should not be `virtual`.

src/Lucene.Net/Index/TermVectorsConsumerPerField.cs
```diff
@@ -141,7 +141,7 @@ internal override bool Start(IIndexableField[] fields, int count)
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        public void Abort()
+        public static void Abort() // LUCENENET: CA1822: Mark members as static
```
NightOwl888, line 144: This method was not marked final in Lucene 4.8.0, therefore it should be made `virtual`.

```java
  public void abort() {}
```

NightOwl888 (changes requested):

NightOwl888:
> I've also added virtual to methods that I found had the final keyword in the Java source code.

virtual is the opposite of final.

## Java

- `public`/`protected` methods implicitly default to `virtual` (there is no `virtual` keyword in Java, but they are overridable)
- To make `public`/`protected` methods non-overridable, explicitly mark them `final`
- When a subclass overrides a method, it can be explicitly marked `final` to make it non-overridable

## C#

- `public`/`protected` methods (and properties) implicitly default to `sealed`
- To make `public`/`protected` methods overridable, explicitly mark them `virtual`
- When a subclass overrides a method, it can be explicitly marked `sealed` to make it non-overridable

nikcio:
Oh then I misunderstood your description in #663. I'll look at correcting the mistakes later this week😅

src/Lucene.Net.Tests.Analysis.Phonetic/Language/MatchRatingApproachEncoderTest.cs
nikcio: This has been reverted and methods used here have also been reverted to non-static.

src/Lucene.Net/Index/FreqProxTermsWriterPerField.cs
nikcio: This is a sealed class and cannot be marked virtual 😄

src/Lucene.Net/Index/TermVectorsConsumerPerField.cs
nikcio: This is a sealed class and cannot be marked virtual 😄

nikcio:
@NightOwl888 Review changes have now been made.

src/Lucene.Net/Index/FreqProxTermsWriterPerField.cs
NightOwl888: I took a look and this method is not being called, so it is dead code. I ran the tests to confirm it has no purpose. So, the appropriate thing to do would be to remove it and replace it with a comment.

```
// LUCENENET: Removed Abort() method because it is not in use.
```

src/Lucene.Net/Index/TermVectorsConsumerPerField.cs
NightOwl888: I took a look and this method is not being called, so it is dead code. I ran the tests to confirm it has no purpose. So, the appropriate thing to do would be to remove it and replace it with a comment.

```
// LUCENENET: Removed Abort() method because it is not in use.
```

NightOwl888 (changes requested):
So, looks like most of what remains are `private` which are fine to make static. There are no protected members in the changeset. And I have reviewed the `public` members. After we remove the two `Abort()` methods and replace them with a comment (see the inline conversations), this is good to merge.

nikcio:
The `Abort` methods have now been replaced with a comment.

NightOwl888:
Actually, have you kept track of the methods we decided not to mark static that are triggering this warning? If so, we should suppress those via attribute.

```c#
[SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "By design")]
```

If you haven't kept track:

1. We can merge this to master and then scan, and then submit a new PR with these attributes.
2. Scan this branch to find the CA1822 warnings and fix them before the merge.
3. Use Visual Studio to scan the entire solution for CA1822, and submit another commit to this PR with these attributes.
4. Use the conversations above to try to work it out, but I think we may miss some that way. Most likely we would need another PR to add attributes where they are missing.

Any option is fine.

nikcio:
@NightOwl888 There's two that I couldn't change due to them causing errors these are:

https://sonarcloud.io/project/issues?issues=AYPAuN5xhbfJOGLOoaDV&open=AYPAuN5xhbfJOGLOoaDV&id=nikcio_lucenenet

https://sonarcloud.io/project/issues?issues=AYPAuOHjhbfJOGLOoaMN&open=AYPAuOHjhbfJOGLOoaMN&id=nikcio_lucenenet

I've marked them in SonarCloud.

But regarding the rest of the changes in the review phase here I haven't had the time to mark them in SonarCloud yet. Therefore, I think the best approach would be to merge these changes as is and create a new PR with the attributes on the warnings we can rule out like the two links above and the ones that have been caught in the review process here.

I think that doing a branch sweep for the warning is a little to early just yet because the new static methods might have opened up for more methods being available to be made static. Therefore, we cannot be sure that the next scan is only the cases we have found to be unable to be marked static.

nikcio:
We should be able to find all the cases where we have decided that we cannot mark methods static in this commit: https://github.com/apache/lucenenet/pull/690/commits/bbf4289e7a689fddec14a7575702f2335e6d94a7

NightOwl888:
> We should be able to find all the cases where we have decided that we cannot mark methods static in this commit: [bbf4289](https://github.com/apache/lucenenet/commit/bbf4289e7a689fddec14a7575702f2335e6d94a7)

Yea, that looks like it could work for what we have in this PR. You can go ahead and move the attribute up to the class level for the one test class instead of putting it on every test method (although, not sure why we are getting a test class, since the CA1822 warning suppression is explicitly configured in all of the test projects).

But, you are right about the cascade effect of warnings that happen due to fixing these. We had already fixed these once, and are finding them again. So it may take a few iterations to catch them all.

NightOwl888 (approved):

PR #693, opened 2022-10-18
chore: Added suppression attributes to CA1822

nikcio:
Following #690 I've now added suppression attributes to the methods we decided not to make static.

Adds to #663 

I also had another look at [This issue](https://sonarcloud.io/project/issues?issues=AYPAuN5xhbfJOGLOoaDV&open=AYPAuN5xhbfJOGLOoaDV&id=nikcio_lucenenet) and found that the reason this caused problems is because I forgot to change the `this.` to `BasicModelBE.` in the `Score` method. But I don't know if this is moving too far away from the original code?

The change would be:

```csharp
public override sealed float Score(BasicStats stats, float tfn)
{
    double F = stats.TotalTermFreq + 1 + tfn;
    // approximation only holds true when F << N, so we use N += F
    double N = F + stats.NumberOfDocuments;
    return (float)(-SimilarityBase.Log2((N - 1) * Math.E) + BasicModelBE.F(N + F - 1, N + F - tfn - 2) - BasicModelBE.F(F, F - tfn));
}
/// <summary>
/// The <em>f</em> helper function defined for <em>B<sub>E</sub></em>. </summary>
private static double F(double n, double m) // LUCENENET: CA1822: Mark members as static
{
    return (m + 0.5) * SimilarityBase.Log2(n / m) + (n - m) * SimilarityBase.Log2(n);
}
```

### DIFF
--- a/src/Lucene.Net.Analysis.Phonetic/Language/MatchRatingApproachEncoder.cs
+++ b/src/Lucene.Net.Analysis.Phonetic/Language/MatchRatingApproachEncoder.cs
@@ -1,5 +1,6 @@
 ﻿// commons-codec version compatibility level: 1.9
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Text;
 
@@ -84,6 +85,7 @@ namespace Lucene.Net.Analysis.Phonetic.Language
         /// </summary>
         /// <param name="name">The name to be cleaned.</param>
         /// <returns>The cleaned name.</returns>
+        [SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "By design")]
         internal string CleanName(string name)
         {
             string upperName = LOCALE_ENGLISH.TextInfo.ToUpper(name);
@@ -135,6 +137,7 @@ namespace Lucene.Net.Analysis.Phonetic.Language
         /// </summary>
         /// <param name="name">The string to get the substrings from.</param>
         /// <returns>Annexed first &amp; last 3 letters of input word.</returns>
+        [SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "By design")]
         internal string GetFirst3Last3(string name)
         {
             int nameLength = name.Length;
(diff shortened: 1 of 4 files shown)


NightOwl888:
> I also had another look at [This issue](https://sonarcloud.io/project/issues?issues=AYPAuN5xhbfJOGLOoaDV&open=AYPAuN5xhbfJOGLOoaDV&id=nikcio_lucenenet) and found that the reason this caused problems is because I forgot to change the `this.` to `BasicModelBE.` in the `Score` method. But I don't know if this is moving too far away from the original code?
> 
> The change would be:
> 
> ```cs
> public override sealed float Score(BasicStats stats, float tfn)
> {
>     double F = stats.TotalTermFreq + 1 + tfn;
>     // approximation only holds true when F << N, so we use N += F
>     double N = F + stats.NumberOfDocuments;
>     return (float)(-SimilarityBase.Log2((N - 1) * Math.E) + BasicModelBE.F(N + F - 1, N + F - tfn - 2) - BasicModelBE.F(F, F - tfn));
> }
> /// <summary>
> /// The <em>f</em> helper function defined for <em>B<sub>E</sub></em>. </summary>
> private static double F(double n, double m) // LUCENENET: CA1822: Mark members as static
> {
>     return (m + 0.5) * SimilarityBase.Log2(n / m) + (n - m) * SimilarityBase.Log2(n);
> }
> ```

```java
 
  /** The <em>f</em> helper function defined for <em>B<sub>E</sub></em>. */
  private final double f(double n, double m) {
    return (m + 0.5) * log2(n / m) + (n - m) * log2(n);
  }
```

In the original code the function `f` was lowercase. I first thought that forced them to use uppercase `F` as a varible name. But, judging by the documentation and comments, the capital F variable name seems to be a math thing, and therefore intentional.

Let's revert this and add the `SuppressMessage` attribute.

We should also add the `[MethodImpl(MethodImplOptions.AggressiveInlining)]` attribute to prompt the compiler to inline this math in the `Score` function, which will give a similar benefit to making the method static.

I have also opened #694 which is related, but should be done in a separate PR since it goes beyond this one class.

NightOwl888 (changes requested):
@nikcio - Forgot to go through the review process when I left my [last comment](https://github.com/apache/lucenenet/pull/693#issuecomment-1282104390).

nikcio:
The PR has now been updated after the comment above

NightOwl888 (approved):

PR #696, opened 2022-10-18
SWEEP: Add a nested comment explaining why this method is empty.

### LINKED ISSUE #681: Add a nested comment explaining why this method is empty, throw a 'NotSupportedException' or complete the implementation
Mentioned in #648 

Issues found: https://sonarcloud.io/project/issues?resolved=false&rules=csharpsquid%3AS1186&id=nikcio_lucenenet

- [x] Looks like the TagSoup members weren't all reviewed to ensure that members were marked `virtual` in .NET if they aren't explicitly marked `final` in Java: https://android.googlesource.com/platform/external/tagsoup/+/donut-release/src/org/ccil/cowan/tagsoup/PYXWriter.java. Accessibility should be checked and corrected on every file in that project.

- [x] Similar issues exist in `Lucene.Net.Facet`. Interfaces don't need to be overridden, but if the member overrides an interface in Java without explicitly being marked `final`, it should be `virtual` in .NET. Any classes in `Lucene.Net.Facet` where an internal/public/protected method exists should be checked for accessibility against Lucene 4.8.1. If there are many that were missed, the whole `Lucene.Net.Facet` project should be checked.

- [ ] If a member simply doesn't do anything, just add a comment `// LUCENENET: Intentionally blank`.

NightOwl888:
Fixes #681.

### DIFF
--- a/src/Lucene.Net/Index/DocumentsWriterFlushControl.cs
+++ b/src/Lucene.Net/Index/DocumentsWriterFlushControl.cs
@@ -563,17 +563,17 @@ namespace Lucene.Net.Index
 
         private IEnumerator<ThreadState> GetPerThreadsIterator(int upto)
         {
-            return new IteratorAnonymousClass(this, upto);
+            return new EnumeratorAnonymousClass(this, upto);
         }
 
-        private class IteratorAnonymousClass : IEnumerator<ThreadState>
+        private class EnumeratorAnonymousClass : IEnumerator<ThreadState>
         {
             private readonly DocumentsWriterFlushControl outerInstance;
             private ThreadState current;
             private readonly int upto;
             private int i;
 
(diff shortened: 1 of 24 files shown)


src/Lucene.Net/Index/DocumentsWriterFlushControl.cs
```diff
@@ -563,17 +563,17 @@ public IEnumerator<ThreadState> AllActiveThreadStates()
 
         private IEnumerator<ThreadState> GetPerThreadsIterator(int upto)
         {
-            return new IteratorAnonymousClass(this, upto);
+            return new EnumeratorAnonymousClass(this, upto);
         }
 
-        private class IteratorAnonymousClass : IEnumerator<ThreadState>
+        private class EnumeratorAnonymousClass : IEnumerator<ThreadState>
```
NightOwl888, line 569: Oops, this change should have been part of #698.

PR #727, opened 2022-10-28
Disabled Json Source Generator.

rclabo:
Works around a recent VS2022 bug that causes a StackOverflowException when building the solution in VS2022.


rclabo:
I ran tests for this PR in AzureDev Ops and nothing failed. I did this since Apache appears to be out of build agents today.

PR #728, opened 2022-10-28
Updated Azure DevOps Mac OS used for Net3.1 Testing

rclabo:
The macOS-10.15 environment is deprecated and will be removed from Azure DevOps on December 1st, 2022. For more details, see https://github.com/actions/virtual-environments/issues/5583 
So this PR updates our use of MacOS-10.15 to macOS-latest.


rclabo:
I ran tests for this PR in Azure DevOps as well and and nothing failed.

Issue #662, opened 2022-10-15
Change 'static readonly' fields to 'const', where appropriate


NightOwl888:
For private fields that are either primitive type or string, we can simply change these to const with no ill effects (which can be its own PR).

For internal/public/protected fields, generally the answer is yes, we should change these. However, changing to const means compiling the value into a consuming assemblies' metadata, so we need to be judicious on how to deal with these.

1. If the value will certainly never change, we can change to `const`.
2. If the value is likely to change, we should leave as `static readonly`.
3. All other cases need to be considered carefully.

For example, we need to consider the implications of what will happen if someone uses an older version of one of the modules, such as `Lucene.Net.Analysis.Common` with a newer version of `Lucene.Net` that has updates to these constant values. If there is any doubt, these should be `static readonly`.

Do note that making this change may result in compile errors which means we need to either revert the change or apply a fix, if appropriate.

> **NOTE:** In Java, there is only one option, `static final`, so this is a .NET-specific translation problem and there is no need to add a `// LUCENENET` comment for it.

paulirwin:
Bumping this to beta 18 in light of #1014 feedback.

PR #729, opened 2022-10-29
#679 Added sonar suggestions 

### LINKED ISSUE #679: Change the visibility of 'EOS_FLAG_BIT' or make it 'const' or 'readonly'
Mentioned in #648 

Specific issue: https://sonarcloud.io/project/issues?issues=AYPAuQX3hbfJOGLOob4n&open=AYPAuQX3hbfJOGLOob4n&id=nikcio_lucenenet

Make it const

Similar issues: https://sonarcloud.io/project/issues?resolved=false&rules=csharpsquid%3AS2223&statuses=OPEN&id=nikcio_lucenenet

sachdevlaksh:
Fixes #679 related to code quality issue detected by Sonar(access modifier to reduce visibility)

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/HyphenationTree.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/HyphenationTree.cs
@@ -99,7 +99,7 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
             byte v = m_vspace[k++];
             while (v != 0)
             {
-                char c = (char)(v.TripleShift(4) - 1 + '0');;
+                char c = (char)(v.TripleShift(4) - 1 + '0');
                 buf.Append(c);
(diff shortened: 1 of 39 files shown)


NightOwl888 (changes requested):
Thanks for the PR.

#729 was only intended to fix the `EOS_FLAG_BIT` field as was indicated in the title. However, a list of related issues made it into the description, which made it confusing. Since this extra work is now already done, I reviewed all of the changes so we can remove more items from the scan. There are a lot of things that need to be fixed, though.

Note that some of this would be addressed by fixing #662, and as a result we have marked some fields `static readonly` when they should be marked `const`. These were marked `final` in Lucene, so the `readonly` modifier was the minimum required to make it behave the same in .NET. But since Java has no `const`, we have to make a judgement call whether it should be `static readonly` or `const` in .NET. I will leave it up to you whether you want to fix them now and mark them `const` or leave them as-is where they are more lines to review in #662.

Please see the inline comments for details.

PR #735, opened 2022-10-31
Removed .NET Core 3.1 tests and lucene-cli support for it. 

rclabo:
.NET Core 3.1 will reach end of life in 45 days on Dec 13, 2022 so it was removed.


TestTargetFramework.props
```diff
@@ -28,14 +28,13 @@
...
     <!--<TargetFramework>net48</TargetFramework>-->
-    <!--<TargetFramework>netcoreapp3.1</TargetFramework>-->
     <TargetFramework>net5.0</TargetFramework>
     <!--<TargetFramework>net6.0</TargetFramework>-->
 
     <!-- Allow the build script to pass in the test frameworks to build for.
       This overrides the above TargetFramework setting. 
       LUCENENET TODO: Due to a parsing bug, we cannot pass a string with a ; to dotnet msbuild, so passing true as a workaround -->
-    <TargetFrameworks Condition=" '$(TestFrameworks)' == 'true' ">net6.0;net5.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(TestFrameworks)' == 'true' ">net6.0;net5.0;</TargetFrameworks>
```
NightOwl888, line 37: There is an extra `;` at the end here that can be removed. When this value is concatenated with the values below, we end up with an empty array element `;;`.

src/dotnet/tools/lucene-cli/docs/index.md
```diff
@@ -1,10 +1,10 @@
-# Lucene.NET command line interface (CLI) tools
+﻿# Lucene.NET command line interface (CLI) tools
 
 The Lucene.NET command line interface (CLI) is a new cross-platform toolchain with utilities for maintaining Lucene.NET and demos for learning basic Lucene.NET functionality.
 
 ## Prerequisites
 
-- [.NET Core 3.1 Runtime or Higher](https://dotnet.microsoft.com/en-us/download/dotnet)
+- [.NET Core 5.0 Runtime or Higher](https://dotnet.microsoft.com/en-us/download/dotnet)
```
NightOwl888, line 7: Technically, it is `.NET 5.0 Runtime or Higher`.

NightOwl888 (changes requested):

NightOwl888 (approved):

rclabo:
Wow! @NightOwl888  you have more commits on this PR then I do.  Looks like I only got the party started.  I appreciate you buddy!

NightOwl888:
Yea, it took some work to solve some issues:

1. Pre-pack the lucene-cli tool prior to the test, which is much faster than having the test runner do it.
2. Set the Azure DevOps environment variable so the tests know where to look for the `.nupkg` file.
3. Bump the `Microsoft.Extensions.Configuration` version from 3.1.0 to 3.1.6. Now the tests work both for lucene-cli and for the `net461` target.

Issue #644, opened 2022-09-22
Robustness issue in QuadPrefixTree


chenhh021:
Hi,

when trying to index a point with coordinate (144502.06, 639062.07) in Lucene.Net.Spatial, a System.ArgumentOutOfRangeException is thrown out.  The stack trace is listed below. This bug can be reproduced by [this test](https://github.com/apache/lucenenet/commit/956c78c9eed7548a1069f5633510c7aa351a9ad4).

`System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
Actual value was 0.
   at J2N.Collections.Generic.List‘1.get_Item(Int32 index)
   at Lucene.Net.Spatial.Prefix.Tree.QuadPrefixTree.GetCell(IPoint p, Int32 level) in D:\chh\SE\projects\lucenenet\src\Lucene.Net.Spatial\Prefix\Tree\QuadPrefixTree.cs:line 167
   at Lucene.Net.Spatial.Prefix.Tree.SpatialPrefixTree.GetCells(IPoint p, Int32 detailLevel, Boolean inclParents) in D:\chh\SE\projects\lucenenet\src\Lucene.Net.Spatial\Prefix\Tree\SpatialPrefixTree.cs:line 286
   at Lucene.Net.Spatial.Prefix.Tree.SpatialPrefixTree.GetCells(IShape shape, Int32 detailLevel, Boolean inclParents, Boolean simplify) in D:\chh\SE\projects\lucenenet\src\Lucene.Net.Spatial\Prefix\Tree\SpatialPrefixTree.cs:line 204
   at Lucene.Net.Spatial.Prefix.PrefixTreeStrategy.CreateIndexableFields(IShape shape, Double distErr) in D:\chh\SE\projects\lucenenet\src\Lucene.Net.Spatial\Prefix\PrefixTreeStrategy.cs:line 139
   at Lucene.Net.Spatial.Prefix.PrefixTreeStrategy.CreateIndexableFields(IShape shape) in D:\chh\SE\projects\lucenenet\src\Lucene.Net.Spatial\Prefix\PrefixTreeStrategy.cs:line 133
   at Lucene.Net.Spatial.StrategyTestCase.newDoc(String id, IShape shape) in D:\chh\SE\projects\lucenenet\src\Lucene.Net.Tests.Spatial\StrategyTestCase.cs:line 220
[...]

PR #619, opened 2022-02-11
BREAKING: Lucene.Net.Spatial Updates

NightOwl888:
- Upgraded to new [Spatial4n v0.4.1.1 NuGet package](https://www.nuget.org/packages/Spatial4n/0.4.1.1) (which unifies the types from `Spatial4n.Core` and `Spatial4n.Core.NTS` in a single assembly, as spatial4j does)
- **SWEEP:** `Lucene.Net.Spatial`: Updated namespaces from `Spatial4n.Core` to `Spatial4n`
- **SWEEP:** `Lucene.Net.Spatial`: Updated namespaces from `Spatial4n.Core.Shapes.Impl` to `Spatial4n.Shapes`
- **SWEEP:** `Lucene.Net.Spatial`: Upgraded to new API names in Spatial4n
- **SWEEP:** `Lucene.Net.Spatial`: De-Nested public/protected types
- **SWEEP:** `Lucene.Net.Spatial`: Enabled nullable reference type support
- **SWEEP:** `Lucene.Net.Spatial`: Added guard clauses, where appropriate

PR #738, opened 2022-11-02
Robustness issue in QuadPrefixTree

chenhh021:
Fix [#644](https://github.com/apache/lucenenet/issues/644) by mimic lucene code. For reference, see this [commit](https://github.com/apache/lucene/commit/26628b2717a73235db56fde94f7f5b64cbc5b8b2) on lucene side. It was for lucene 9.4, but much of the change is valid for the present version of lucene.net.  Contents are:

1. Reimplement QuadPrefixTree's getCell method.
2. Add a version check so the older versions still use the older implementation.

I have confirmed that all spatial tests pass after these changes.

Could anyone help review this PR?


NightOwl888 (commented):
Thanks for the PR.

Having the spatial module return 8.3.0 behavior when the user requests 4.8.0 compatibility is an issue that would have to be overcome for this to be accepted.

In addition, where we have taken liberties to patch things, we ported over the entire patch from Lucene so we can at least step through the code and debug on the specific version of Lucene the patch was released in. This PR doesn't contain all of the changes in https://github.com/apache/lucene-solr/pull/824/files#diff-9e5d8fe4fee51beb8928021158c41f2e65c9450fa290d873fa682aee42fb0982R65 and [LUCENE-8755](https://issues.apache.org/jira/browse/LUCENE-8755) and contains changes to the execution path that 4.8.0 follows, making it difficult to debug should we need to - the way we find bugs is by checking where the execution paths diverge between Lucene and Lucene.NET.

Given that the LUCENE-8755 patch has classes that don't yet exist in 4.8.0 and the fact that these are public abstract classes that you can build your own customizations on top of in Lucene 4.8.0, I don't believe it is worth the effort to maintain this in our repository until we have reached line-by-line parity with Lucene 8.3.0+.

#619 was intended to be the last update to `Lucene.Net.Spatial` APIs and behavior before the 4.8.0 release. I can say with confidence it is "done" and I am pretty sure the rest of the Lucene.NET community whom are awaiting the 4.8.0 release would like to keep it that way.

chenhh021:
Thanks for the review and telling the reasons that this PR cannot get accepted.  I'll try to fix the related problem in my own fork.

I'm wondering how to treat the related issue, is it okay that just let it open there?

rclabo:
I agree with @NightOwl888 .  That said, @chenhh021 I'd also like to thank you for the PR.  While it doesn't make sense at this point to merge this into LuceneNET 4.8, having this PR linked from the related issue provides code that may help someone in the future if they run up against this issue in LuceneNET 4.8.  So your contribution may well end up helping others (even me!).  So it's very much appreciated.  Thank you.

NightOwl888:
> Thanks for the review and telling the reasons that this PR cannot get accepted. I'll try to fix the related problem in my own fork.
> 
> I'm just wondering how to treat the related issue, is it okay that just let it open there?

Being that this issue is technically a duplicate of [LUCENE-8755(https://issues.apache.org/jira/browse/LUCENE-8755) and will go away when we upgrade the port, I have just closed it. However, if you could work out how to extend `QuadPrefixTree` (or replace it, whichever is sensible) and share your final solution there, I am sure others looking for an answer to this would find it helpful.

Issue #271, opened 2019-07-13
TokenStream.IncrementToken() is called after Dispose() is called


clambertus:
<p>When overriding <tt>Dispose(bool)</tt> in either <tt>TokenStream</tt> subclasses, Lucene.Net will call <tt>Dispose()</tt> before it is done using the <tt>TokenStream</tt>, and call <tt>IncrementToken()</tt> again.</p>

<p> </p>

<p>The behavior can be observed in the <tt>Lucene.Net.Collation.TestICUCollationKeyFilter.TestCollationKeySort()</tt> when the <tt>ICUCollationKeyFilter</tt> implements <tt>Dispose()</tt>.</p>
<i>JIRA link - <a href="https://issues.apache.org/jira/browse/LUCENENET-611">[LUCENENET-611]</a> created by nightowl888</i>

NightOwl888:
I looked into this a little, and it appears that the bug that was fixed in the test framework didn't solve this. When adding a `Dispose(bool)` override to the `ICUCollationKeyFilter`, `Dispose(true)` is still being called followed by additional calls to `IncrementToken()`.

This is occuring in `DocFieldProcessor` which [does a tight loop](https://github.com/apache/lucenenet/blob/081edeed35b190c1d535dcfdfeb91143f0ef818f/src/Lucene.Net/Index/DocFieldProcessor.cs#L280-L284) around some fields. It is possible that the consumer will dispose the `TokenFilter` in one loop and then another loop will retrieve the same instance of that `TokenFilter` by calling `Field.GetTokenStream(Analyzer)` and call `IncrementToken()` on it after it is disposed.

At this point I haven't confirmed this is how it works in Java so it is still unclear if this is expected behavior or a bug. If the former, we should consider the possibility of making `TokenStream` "close" and do that separately than "dispose" (see #265).

NightOwl888:
I started by trying to solve the two [TestRandomChains]() failures (see #269), but it ended up unraveling back to this issue again.

I isolated a specific case that fails in .NET:

```c#
[Test]
public void TestWordDelimiterFilterChain()
{
    Analyzer a = Analyzer.NewAnonymous(createComponents: (fieldName, reader) =>
    {
        Tokenizer tokenizer = new NGramTokenizer(LuceneVersion.LUCENE_48, reader, 70, 79);
        TokenStream filter = tokenizer;
[...]

Issue #265, opened 2019-10-10
API: Review to ensure IDisposable is being used correctly and disposable pattern implemented correctly


clambertus:
There have been several issues found recently with the disposable pattern not being implemented correctly. I have also been made aware that there is at least one class (Lucene.Net.Store.Lock, if I recall correctly) that is designed to be re-opened after it is closed.

We need a review to ensure all classes that implement disposable are doing it correctly and have correctly implemented the dispose pattern both for sealed and unsealed types. We also need to have a close look at whether any classes should be reverted back to using Close() instead of Dispose() on account that the class instance was designed to be used again after the Dispose() call.

We should also explicitly check to ensure that `Dispose()` is set up to be called more than once safely. That is, after the first call, all additional calls should be a no-op.

<i>JIRA link - <a href="https://issues.apache.org/jira/browse/LUCENENET-626">[LUCENENET-626]</a> created by nightowl888</i>

nikcio:
Issues found: https://sonarcloud.io/project/issues?resolved=false&rules=csharpsquid%3AS3881&id=nikcio_lucenenet

Mentioned in #648 :
Note this is part of https://github.com/apache/lucenenet/issues/265. Most of the issues I have spotted were failure to call `base.Dispose(disposing)` after exiting the `if (disposing)` block. Others are because they are on a private class that should be sealed, but it is not (Enumerators).

PR #746, opened 2022-11-06
fix: Aligned disposable patterns

nikcio:
Related to #265 

I've run through most of the issues reported in Sonarcloud related to the use of the IDisposable pattern (https://sonarcloud.io/project/issues?resolved=false&rules=csharpsquid%3AS3881&id=apache_lucenenet)

This PR is meant to further expose which implementations of the pattern require more investigation while also removing the easiest areas where the pattern just needs a simple alignment. This means that if any of the following changes exposes an issue requiring much more effort this is best to be moved to comment on the existing issue and have a solution opened in another PR. I realize that a correct implementation of this pattern can be advanced and is related to at least a handful of issues opened in this repository which is why I want to take the approach stated 😄 .

To help the review I've outlined the changes here in some categories:

## Issues from private classes were marked `sealed`
Private classes don't have to uphold the same pattern as other classes if they are marked `sealed`.

## Basic implementation of pattern in some classes
The following files had a class where the following template was implemented.
```csharp
public void Dispose()
{
    Dispose(true);
    GC.SuppressFinalize(this);
}

protected virtual void Dispose(bool disposing)
{
    if (disposing)
    {
        // Cleanup
    }
}
```

1. BufferedCharFilter
2. MemoryDocValuesConsumer
3. PerFieldDocValuesFormat
4. PerFieldPostingsFormat
5. IndexWriter
6. TaskMergeScheduler

## `base.Dispose(disposing)` referencing code
The `base.Dispose(disposing)` line in `BufferedCharFilter` references the dispose method in `Lucene.Net.Analysis.CharFilter`

## `base.Dispose(disposing)` referencing disposing methods in `System.IO`
The changes in the following files have `base.Dispose(disposing)` methods which reference code in `System.IO`

1. SourceCodeSectionReader
2. SafeTextWriterWrapper

## Fixed `GC.SuppressFinalize` call
In `PrefixCodedTerms` the call to `GC.SuppressFinalize` had the object parameter set to `true` instead of `this`

## `FSDirectory`
There's a commented `base.Disposing(disposed)` change in `FSDirectory` which when activated causes at least 140 tests to fail (I didn't run it all the way through when I noticed the error). Please validate that the `Dispose(bool)` method isn't supposed to call its base.

## Other `base.Dispose(dispoing)`
Other places that aren't part of the categories above reference an empty `Dispose(bool)` method in its base. Therefore, this shouldn't cause any side effects and only make future development more reliable as the classes will follow the same dispose pattern.

## Sidenote
Assuming the comment around this `Dispose` method is correct this issue can be removed as intended either in SonarCloud or in source: https://sonarcloud.io/project/issues?issues=AYRH0T17_qq9ReJdi40Q&open=AYRH0T17_qq9ReJdi40Q&id=apache_lucenenet

### DIFF
--- a/src/Lucene.Net.Facet/Taxonomy/WriterCache/CollisionMap.cs
+++ b/src/Lucene.Net.Facet/Taxonomy/WriterCache/CollisionMap.cs
@@ -213,7 +213,7 @@ namespace Lucene.Net.Facet.Taxonomy.WriterCache
             return memoryUsage;
         }
 
-        private class EntryEnumerator : IEnumerator<Entry>
+        private sealed class EntryEnumerator : IEnumerator<Entry>
         {
             internal Entry next; // next entry to return
             internal int index; // current slot
--- a/src/Lucene.Net.Facet/Taxonomy/WriterCache/CollisionMap.cs
+++ b/src/Lucene.Net.Facet/Taxonomy/WriterCache/CollisionMap.cs
@@ -213,7 +213,7 @@ namespace Lucene.Net.Facet.Taxonomy.WriterCache
             return memoryUsage;
         }
 
-        private sealed class EntryEnumerator : IEnumerator<Entry>
+        private sealed class EntryEnumerator : IEnumerator<Entry> // LUCENENET: Marked sealed
         {
             internal Entry next; // next entry to return
             internal int index; // current slot
--- a/src/Lucene.Net.Analysis.Common/Analysis/Miscellaneous/PrefixAndSuffixAwareTokenFilter.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Miscellaneous/PrefixAndSuffixAwareTokenFilter.cs
@@ -1,4 +1,4 @@
-// Lucene version compatibility level 4.8.1
+﻿// Lucene version compatibility level 4.8.1
 namespace Lucene.Net.Analysis.Miscellaneous
 {
     /*
@@ -96,6 +96,7 @@ namespace Lucene.Net.Analysis.Miscellaneous
             {
                 suffix.Dispose();
             }
+            base.Dispose(disposing);
         }
 
         public override void End()
--- a/src/Lucene.Net.Analysis.Common/Analysis/Miscellaneous/PrefixAndSuffixAwareTokenFilter.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Miscellaneous/PrefixAndSuffixAwareTokenFilter.cs
@@ -96,7 +96,7 @@ namespace Lucene.Net.Analysis.Miscellaneous
             {
                 suffix.Dispose();
             }
-            base.Dispose(disposing);
+            base.Dispose(disposing); // LUCENENET specific - disposable pattern requires calling the base class implementation
         }
 
         public override void End()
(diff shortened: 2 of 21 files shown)


src/Lucene.Net.Facet/Taxonomy/WriterCache/CollisionMap.cs
```diff
@@ -213,7 +213,7 @@ internal virtual int GetMemoryUsage()
             return memoryUsage;
         }
 
-        private class EntryEnumerator : IEnumerator<Entry>
+        private sealed class EntryEnumerator : IEnumerator<Entry>
```
NightOwl888, line 216: Please add a comment to the end of this line ` // LUCENENET: Marked sealed` to indicate we intentionally diverged from Lucene here.

NightOwl888 (changes requested):
Thanks for the PR.

The token streams are still a WIP (see https://github.com/apache/lucenenet/issues/271#issuecomment-1305099323), but I think having them set up consistently for .NET is a good starting point even if we eventually change them back to using `Close()` instead of `Dispose()`. So, we can keep those changes in this PR.

There are just a couple of minor changes to do before this can be accepted. See the comments inline.

NightOwl888:
> ## Sidenote
> Assuming the comment around this `Dispose` method is correct this issue can be removed as intended either in SonarCloud or in source: https://sonarcloud.io/project/issues?issues=AYRH0T17_qq9ReJdi40Q&open=AYRH0T17_qq9ReJdi40Q&id=apache_lucenenet

It looks like this was mainly done to prevent the end user from being able to override the default `Dispose(bool)` implementation and forget to call `base.Dispose(disposing);`. We can probably just make the protected `Dispose(bool)` into a virtual method and move the implementation there to address this warning as long as we include doc comments to always call the base implementation when overriding `Dispose(bool)`.

NightOwl888:
> ## `FSDirectory`
> There's a commented `base.Disposing(disposed)` change in `FSDirectory` which when activated causes at least 140 tests to fail (I didn't run it all the way through when I noticed the error). Please validate that the `Dispose(bool)` method isn't supposed to call its base.

I took a look and the only thing that `BufferedIndexOutput` does in its `Dispose(bool)` method is call `Flush()`. There is a comment at the top of `FSIndexOutput` that give some insight:

```c#
// LUCENENET specific: Since FileStream does its own buffering, this class was refactored
// to do all checksum operations as well as writing to the FileStream. By doing this we elminate
// the extra set of buffers that were only creating unnecessary memory allocations and copy operations.
```

So, the inheritance chain was left unchanged (for consistency with Lucene), but we aren't actually using the functionality of the base class. Instead, we call `Flush()` on `FileStream` and dispose it.

The `BufferedIndexOutput` class doesn't actually have any `IDisposable` objects in it.

This looks fine.

NightOwl888:
> ## `base.Dispose(disposing)` referencing code
> The `base.Dispose(disposing)` line in `BufferedCharFilter` references the dispose method in `Lucene.Net.Analysis.CharFilter`

The `BufferedCharFilter` is a patch to fix `TextReader`, since .NET doesn't have a `BufferedTextReader`. So, we decorate a `TextReader` with this class in order to make it support the extra state and methods to make it "buffered".

NightOwl888:
@nikcio - Thanks again for doing this. Your help is really appreciated.

I went ahead and updated the comments to finish this up (for the most part). We will know what else needs to be done in the scan, and we are tracking #265 and we also have a plan to fix #271.

I didn't do anything with `TaxononmyReader` yet. We need to get to the bottom of why we are seeing locking contention in the tests that didn't exist in Lucene before tackling that one.

NightOwl888 (approved):

PR #751, opened 2022-11-07
Andjela

AndjelaSmiljanic:


NightOwl888 (commented):
Thanks for the PR.

Please note we already [have SonarCloud set up](https://github.com/apache/lucenenet/blob/dfae964ce0a1b066fd808275fd81d90385cde3fe/.github/workflows/sonar.yml), but not everything that comes up in the scan is approved to work on.

We appreciate you may just be trying to help, but please work on issues that are [up for grabs](https://github.com/apache/lucenenet/issues?q=is%3Aissue+is%3Aopen+label%3Aup-for-grabs) or ask on the [dev mailing list](https://lucenenet.apache.org/contributing/mailing-lists.html#developers-dev) prior to doing the work.

Unfortunately, we have to reject this because it is mostly changes that diverge from the [Lucene 4.8.0 source](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0) and/or are introducing possible bugs.

Issue #717, opened 2022-10-25
Add .Net 7 Tests


nikcio:
> (Before edit)
> As .Net 7 is just around the corner it would be great to look into what is needed to add explicit support to the new version.

Assignment is stated in the response from @NightOwl888 

- [x] Remove `netcoreapp3.1` (https://github.com/apache/lucenenet/pull/735)
- [ ] Add .Net 7 test support
- [ ] Update the [build instructions](https://github.com/apache/lucenenet/blob/dc2955699e2b8a2bf31813a819d17749eb7d4a4b/README.md#visual-studio) to indicate we no longer support building with VS2019

NightOwl888:
Unless we are going to make the effort to migrate to new APIs, IMO it doesn't make much sense to add a new target. In fact, we are evaluating whether to drop one of the targets (either `net462` or `netstandard2.0`), since these have significant overlap and `netstandard2.1` pretty much covers all other platforms except .NET Framework and .NET 6+.

However, adding .NET 7 tests to Azure DevOps is a must. Adding .NET 7 tests to GitHub Actions should only be done if we remove the tests for .NET 6, since we only need one set of tests to validate the `net6.0` target and we want the PR tests to run as fast as possible.

Adding a target for `net6.0` made sense because:

1. J2N uses [hardware intrinsics](https://devblogs.microsoft.com/dotnet/hardware-intrinsics-in-net-core/) for some of the low level bit operations and `net6.0` supports both Windows and Linux
2. Normal dictionaries support deleting while iterating forward, so we don't have to use `ConcurrentDictionary<TKey, TValue>`

Being that we only have a couple of months left of `netcoreapp3.1` support, we should drop those tests but keep the `net5.0` tests for testing the `netstandard2.1` target. `net5.0` makes more sense to use because it supports detecting fatal crashes, where `netcoreapp3.1` does not. It would be best to first do a PR for dropping `netcoreapp3.1` tests, then a PR for adding the `net7.0` tests.

### Azure DevOps / build.bat

x64 + x86 on Windows only

- `net7.0` -> tests `net6.0` tfm
- `net6.0` -> tests `net6.0` tfm
- `net5.0` -> tests `netstandard2.1` tfm
- `net48` -> tests `net462` tfm
- `net461` -> tests `netstandard2.0` tfm (will probably drop both of these)

### GitHub Actions

x64 only on all targets

- `net7.0` -> tests `net6.0` tfm
[...]

PR #753, opened 2022-11-08
Remove .NET5 from builds

iamcarbon:
.NET5.0 is no longer supported and will not receive any additional security updates.

This PR removes .NET5.0 configurations from the build & test targets.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -39,7 +39,7 @@
   </PropertyGroup>
   
   <!-- Features in .NET Standard, .NET Core, .NET 5.x, and .NET 6.x only (no .NET Framework support) -->
-  <PropertyGroup Condition=" $(TargetFramework.StartsWith('netstandard')) Or $(TargetFramework.StartsWith('netcoreapp')) Or '$(TargetFramework)' == 'net5.0' Or '$(TargetFramework)' == 'net6.0' ">
+  <PropertyGroup Condition=" $(TargetFramework.StartsWith('netstandard')) Or $(TargetFramework.StartsWith('netcoreapp')) Or '$(TargetFramework)' == 'net6.0' ">
     
     <DefineConstants>$(DefineConstants);NETSTANDARD</DefineConstants>
     <DefineConstants>$(DefineConstants);FEATURE_ARRAYEMPTY</DefineConstants>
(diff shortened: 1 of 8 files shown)


Directory.Build.targets
```diff
@@ -60,15 +60,15 @@
...
+  <!-- Features in .NET Standard 2.x and .NET 6.x -->
+  <PropertyGroup Condition=" $(TargetFramework.StartsWith('netstandard2.')) Or '$(TargetFramework)' == 'net6.0' ">
 
     <DefineConstants>$(DefineConstants);FEATURE_ICONFIGURATIONROOT_PROVIDERS</DefineConstants>
 
   </PropertyGroup>
 
-  <!-- Features in .NET Framework 4.5+, .NET Standard 2.x, .NET Core 2.x, .NET Core 3.x, .NET 5.x, and .NET 6.x  -->
-  <PropertyGroup Condition=" $(TargetFramework.StartsWith('net4')) Or $(TargetFramework.StartsWith('netstandard2.')) Or $(TargetFramework.StartsWith('netcoreapp2.')) Or $(TargetFramework.StartsWith('netcoreapp3.')) Or '$(TargetFramework)' == 'net5.0' Or '$(TargetFramework)' == 'net6.0' ">
+  <!-- Features in .NET Framework 4.5+, .NET Standard 2.x, and .NET 6.x  -->
```
NightOwl888, line 70: Please leave the legacy frameworks in this file so it can still be comprehended why these features exist.

NightOwl888 (changes requested):
Thanks for the PR.

As was pointed out in https://github.com/apache/lucenenet/issues/717#issuecomment-1290655689, dropping our tests for `net5.0` leaves a gap in our test coverage - namely, there are no tests to cover the `netstandard2.1` target. We won't be dropping these tests until we have an [alternative runtime](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1) to test the `netstandard2.1` target framework on. `net7.0` and `net6.0` will both consume the `net6.0` target unless we force them otherwise with `SetTargetFramework`. But when we do that, it is not actually a real-world scenario that we are testing.

Please revert all changes from

- `Directory.Build.targets`
- `TestTargetFramework.props`
- `azure-pipelines.yml`

It is fine to remove the `net5.0` target for `lucene-cli`, its tests, and the `Lucene.Net.Tests.CodeAnalysis`. However, we need to re-generate the GitHub Actions workflows using [Generate-TestWorkflows.ps1](https://github.com/carbon/lucenenet/blob/12488d4f1e2d735f56a58a02708b9bf5bd3db622/.github/workflows/Generate-TestWorkflows.ps1) (once `TestTargetFramework.props` has been reverted to restore `net5.0` tests).

PR #754, opened 2022-11-08
Improve Code Quality

iamcarbon:
- Enables C#10
- Use null propagation
- Removes various unused using statements

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -26,7 +26,7 @@
   </PropertyGroup>
   
   <PropertyGroup>
-    <LangVersion>9.0</LangVersion>
+    <LangVersion>10</LangVersion>
     <GitHubOrganization>apache</GitHubOrganization>
     <GitHubProject>lucenenet</GitHubProject>
   </PropertyGroup>
--- a/src/Lucene.Net/Analysis/TokenAttributes/FlagsAttribute.cs
+++ b/src/Lucene.Net/Analysis/TokenAttributes/FlagsAttribute.cs
@@ -1,6 +1,4 @@
-﻿using System;
-
-namespace Lucene.Net.Analysis.TokenAttributes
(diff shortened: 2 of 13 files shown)


iamcarbon:
Ready for review.

Directory.Build.props
```diff
@@ -26,7 +26,7 @@
   </PropertyGroup>
   
   <PropertyGroup>
-    <LangVersion>9.0</LangVersion>
+    <LangVersion>10</LangVersion>
```
NightOwl888, line 29: This breaks support for VS2019, which we are still using. Please revert.

At least until someone upgrades [Git Tools 2019](https://marketplace.visualstudio.com/items?itemName=yysun.git-tools-2019) to support VS2022. See https://github.com/yysun/git-tools/issues/40.

NightOwl888 (changes requested):
Thanks for the PR.

There is just one requested change to make, please see the inline comments.

Also, please be aware that this is a port from Java, and generally these "code quality" changes are rejected because although they may make the code more easy to read for a C# developer, they make us do more mental gymnastics when comparing against the [original source](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java#L431-L432). We are planning to upgrade to a newer version of Lucene and the more lines that differ, the harder that will be to do. We will accept this change because it generally doesn't disrupt the flow and is easy to comprehend.

Just so we don't have to reject any future PRs because they diverge too far from the original source, please either work on something that is [up for grabs](https://github.com/apache/lucenenet/issues?q=is%3Aissue+is%3Aopen+label%3Aup-for-grabs), or ask whether it is a change that will be accepted on the [dev mailing list](https://lucenenet.apache.org/contributing/mailing-lists.html#developers-dev).

Issue #763, opened 2022-11-17
Support async task


tonyguo2022:
We'd like to build Lucene Index over cloud storage. Most cloud storage operations are asynchronized. I am wondering if Lucene.NET library can add a group of API to Directory and IndexInput/IndexOutput classes to support async operations. e.g.:

existing API in class DataInput:
public virtual void ReadBytes(byte[] b, int offset, int len, bool useBuffer)

New API:
public virtual async Task ReadBytesAsync(byte[] b, int offset, int len, bool useBuffer, CancellationToken cancellationToken)

There could be a default implementation to call the sync method, but people can also override it to call async cloud API to read from remote. 

Thanks.

NightOwl888:
This is an exact duplicate of #485.

While we do have a little better handle on locking contention since then, there are still known concurrency issues to track down in `Lucene.Net.Facet` and in the dictionary-based `BreakIterator` in [ICU4N](https://github.com/NightOwl888/ICU4N) (See #269).

However, our goal of upgrading to a newer version of Lucene when the port is completed has not changed, and not only is this not part of that planned work but it is also something that could potentially slow down the upgrade process. Given the choice between async/await support and the latest Lucene features, I suspect most people would rather have the latter.

tonyguo2022:
I think we do need this feature. Not sure if this is rare. Maybe we can clone this repo and make change there. Thanks!

NightOwl888:
Well, being that we just started our 9th year porting Lucene 4.8.0, this isn't a priority.

I still stand by what was said last year. The Lucene team is the place to submit this request. Let them work out at a high level how to make async/await work with the design. Then when we port the next version it will automatically be part of our TODO list, rather than something that is going to sidetrack us.

It might help to be very specific about which APIs need to be async. If you can narrow the focus it will be more likely something the Lucene team would consider doing.

tonyguo2022:
@rclabo  Async wait could free the calling thread to do other tasks instead being waiting/blocked.
Please check reference: https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/task-asynchronous-programming-model

tonyguo2022:
> It might help to be very specific about which APIs need to be async. If you can narrow the focus it will be more likely something the Lucene team would consider doing.

@NightOwl888 , so far we need to update Lucene Index stored in the cloud. I think we need a few APIs(see below list)  to be async. We can clone this repo to another repo and invite you to review. Thanks.

[...]

PR #769, opened 2022-11-22
Align InputStreamDataInput and OutputStreamDataOutput closer to the Java Source

jeme:
From talks in #763

Fixed InputStreamDataInput and OutputStreamDataOutput so that they don't wrap their input streams to BinaryReader/Writers to align the code closer to the Java source which just uses the input.

I also aligned the dispose method of the two so they behave the same.

This is a PR done a bit in blind as I can't build the full lucene source at the moment.

### DIFF
--- a/src/Lucene.Net/Store/InputStreamDataInput.cs
+++ b/src/Lucene.Net/Store/InputStreamDataInput.cs
@@ -29,7 +29,7 @@ namespace Lucene.Net.Store
 
         public InputStreamDataInput(Stream @is)
         {
-            this._is = @is ?? throw new ArgumentNullException(nameof(@is));
+            this._is = @is ?? throw new ArgumentNullException(nameof(@is)); // LUCENENET specific - added null guard clause
         }
 
         public override byte ReadByte()
--- a/src/Lucene.Net/Store/InputStreamDataInput.cs
+++ b/src/Lucene.Net/Store/InputStreamDataInput.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.IO;
 
 namespace Lucene.Net.Store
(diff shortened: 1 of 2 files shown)


src/Lucene.Net/Store/InputStreamDataInput.cs
```diff
@@ -25,16 +25,16 @@ namespace Lucene.Net.Store
     /// </summary>
     public class InputStreamDataInput : DataInput, IDisposable
     {
-        private BinaryReader _reader;
+        private readonly Stream _is;
 
         public InputStreamDataInput(Stream @is)
         {
-            this._reader = new BinaryReader(@is);
+            this._is = @is ?? throw new ArgumentNullException(nameof(@is));
```
NightOwl888, line 32: Please add a comment after this line `// LUCENENET specific - added null guard clause`

src/Lucene.Net/Store/OutputStreamDataOutput.cs
```diff
@@ -25,21 +25,21 @@ namespace Lucene.Net.Store
     /// </summary>
     public class OutputStreamDataOutput : DataOutput, IDisposable
     {
-        private readonly BinaryWriter _writer;
+        private readonly Stream _os;
 
         public OutputStreamDataOutput(Stream os)
         {
-            this._writer = new BinaryWriter(os);
+            this._os = os;
```
NightOwl888, line 32: Please add a null guard clause here, as well. Add a comment after this line `// LUCENENET specific - added null guard clause`

NightOwl888 (changes requested):

NightOwl888:
Oh, as for being blind, we now have the tests running automatically when you submit a PR, so you can see the build and tests happen without running them locally.

NightOwl888 (approved):

PR #773, opened 2022-11-25
Added 5 part Lucene + Blazor series

thbst16:


NightOwl888:
Thanks for the great contribution! Love it.

Suggestion for Part 6: Build a custom analyzer and use Lucene.Net.TestFramework to test the implementation. The test framework is underutilized IMO. There aren't many tools out there that come with their own tests that can be used to ensure you build extensions correctly. Of course, the extension doesn't have to be an analyzer, but I think that is one of the most common ones. It would also potentially help us find any usability issues with the test framework, since NUnit is nothing like JUnit so we had to pave our own way to porting it.

NightOwl888:
The changes are now live: https://lucenenet.apache.org/contributing/community-links.html. I love automation.

PR #783, opened 2022-12-15
DisjunctionMaxQuery shouldn't depend on disjunct order for equals checks

### LINKED ISSUE #779: The order of disjuncts in DisjunctionMaxQuery affects equals()
See [Lucene-9940](https://issues.apache.org/jira/browse/LUCENE-9940). I find it valid for current Lucene.Net.

DisjunctionMaxQuery stores its disjuncts in a java array, and its equals() implementation uses Arrays.equal() when checking equality. This means that two queries with the same disjuncts but added in a different order will compare as different, even though their results will be identical. We should replace the array with a Set.
 
Fix it may break current behavior, but in case someone need it, I will create a PR.

chenhh021:
Fix #779 . The original fix in Lucene relied on its own implementation of multiset, a similar one is ported in this PR. 

The changes break the consistence with the behavior of Lucene 4.8, but at least provide a solution for those who meet the same problem.

### DIFF
--- a/src/Lucene.Net.Tests.QueryParser/Xml/TestParser.cs
+++ b/src/Lucene.Net.Tests.QueryParser/Xml/TestParser.cs
@@ -107,13 +107,21 @@ namespace Lucene.Net.QueryParsers.Xml
         public void TestDisjunctionMaxQueryXML()
         {
             Query q = Parse("DisjunctionMaxQuery.xml");
-            assertTrue(q is DisjunctionMaxQuery);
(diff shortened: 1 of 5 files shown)


paulirwin:
For posterity, this is a port of https://github.com/apache/lucene/pull/110, although there are some significant differences from how it was done there. This also changes the public API surface of DisjunctionMaxQuery's Disjuncts property, which is a breaking change, so I think my vote would be to hold this for a future release. I'm going to put this in the Future milestone unless @NightOwl888 wants me to pull it in for the next beta.

PR #786, opened 2023-01-09
Test/net crash test

### LINKED ISSUE #768: Port missing test: TestIndexWriterOnJRECrash
This test was excluded because we don't have a JRE. But just because we don't have a JRE doesn't mean we don't have to contend with runtime crashes.

The test would need to be refactored to emulate a crash on .NET. The idea is that when a crash occurs, it doesn't corrupt the index.

- [TestIndexWriterOnJRECrash.cs](https://github.com/apache/lucenenet/blob/4e49612d6867194440694b77db95fd0ed756c9a9/src/Lucene.Net.Tests/Index/TestIndexWriterOnJRECrash.cs)
- [TestIndexWriterOnJRECrash.java](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/test/org/apache/lucene/index/TestIndexWriterOnJRECrash.java)

For anyone curious what it is like to port Java code to .NET, this is a relatively small task that can satisfy your curiosity and will help to ensure Lucene.NET is as stable as it can be.

Jeevananthan-23:
Closes #768

### DIFF
--- a/src/Lucene.Net.Tests/Index/TestIndexWriterOnJRECrash.cs
+++ b/src/Lucene.Net.Tests/Index/TestIndexWriterOnJRECrash.cs
@@ -41,12 +41,14 @@ namespace Lucene.Net.Index
     using Lucene.Net.Support.IO;
     using System.Diagnostics;
     using System.Reflection;
+    using Lucene.Net.Util;
 
     /// <summary>
     /// Runs TestNRTThreads in a separate process, crashes the JRE in the middle
     /// of execution, then runs checkindex to make sure its not corrupt.
     /// </summary>
     [TestFixture]
+    [Category("CrashTest")]
     public class TestIndexWriterOnJRECrash : TestNRTThreads
     {
         private DirectoryInfo TempDir;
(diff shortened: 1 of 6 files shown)


src/Lucene.Net.Tests/Index/TestIndexWriterOnJRECrash.cs
```diff
@@ -1,282 +1,294 @@
...
+        public void ForkTest()
+        {
+            //get the full location of the assembly with DaoTests in it
+            string fullPath = System.Reflection.Assembly.GetAssembly(typeof(TestIndexWriterOnJRECrash)).Location;
+
+            //get the folder that's in
+            string theDirectory = Path.GetDirectoryName(fullPath);
+            // Set up the process to run the console app
+            ProcessStartInfo startInfo = new ProcessStartInfo
+            {
```
laimis, line 142: The paths in the test are specific to your machine. Can you make them more generic so that the tests could run anywhere? I get an error if I try to run this locally.

Also, would recommend declaring a variable and setting the workingDirectory there and then reusing it in Arguments as well as WorkingDirectory properties.

laimis (changes requested):
Are you able to run this locally end to end? For me, once I modified the working directory path to match what's on my machine, I was able to start the test, but when the test gets invoked via Process.Start, dotnet test is encountering locking issues:

C:\Program Files\dotnet\sdk\7.0.104\Microsoft.Common.CurrentVersion.targets(4862,5): error MSB3027: Could not copy "....\Lucene.Net.TestFramework.dll" to "bin\Debug\net7.0\Lucene.Net.TestFramework.dll". Exceeded retry count of 10. Failed. The file is locked by: "testhost (5360)" [..\src\Lucene.Net.Tests._I-J\Lucene.Net.Tests._I-J.csproj]

PR #787, opened 2023-01-12
Fix typo

vrdc90:


rclabo (commented):
Thanks for this.  Please fix one other typo in the same sentence:  "which means we what to do exact" should be "which means we want to do exact"

vrdc90:
No problemo :)
I opened new PR [here](https://github.com/apache/lucenenet/pull/790)

PR #794, opened 2023-03-29
Make TextStartPos and TextEndPos of TextFragment publicly readable

triandco:
I am using ```Lucene.Net.Search.Highlight``` to display search result. Right now, ```Highlighter.GetBestTextFragments``` is returning an array of ```TextFragment```.  

I would like to highlight these fragments in the context of the full document as well as doing some further processing depending on where the fragment occurs. To do so, I need to identify the start and end position of these ```TextFragment``` in the document. 

It seems that these positions are already available in ```TextFragment```:
https://github.com/apache/lucenenet/blob/0bbc3798c1a8abcf06fac3f53df6e07a7312c3d5/src/Lucene.Net.Highlighter/Highlight/TextFragment.cs#L46-L48

If they represent the actual position of the fragment in the original text, they can be made public to enable the mentioned use cases. 

The pull request is a small change to make them publicly readable.

### DIFF
--- a/src/Lucene.Net.Highlighter/Highlight/TextFragment.cs
+++ b/src/Lucene.Net.Highlighter/Highlight/TextFragment.cs
@@ -43,9 +43,9 @@ namespace Lucene.Net.Search.Highlight
             protected internal set => score = value;
         }
 
-        // LUCENENET specific - made these fields into properties, since they are for internal consumption
-        internal int TextEndPos { get; set; }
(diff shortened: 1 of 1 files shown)


triandco:
I’m closing this PR because have realised that this TextStartPos and Text EndPos does not represent the original position of the token after further testing. It is only used to keep track of the sequence fragment length, which is not particularly useful for downstream processing.

Issue #670, opened 2022-10-15
Remove this call from a constructor to the overridable 'xxx' method/property


nikcio:
Mentioned in #648 

Issues found: https://sonarcloud.io/project/issues?resolved=false&rules=csharpsquid%3AS1699&id=apache_lucenenet

This is a difference in initialization order between Java and .NET. In Java it is safe to make a call to a virtual member from a constructor, but in .NET it is not. Same issue in https://github.com/apache/lucenenet/issues/408. I still haven't found a good solution to this, but it would be nice to have a list of all of the places where it is a problem so we can consider what solution(s) will work.

laimis:
All the issues that were reported by this rule have been addressed. Thank you for setting this up and reporting it!

The fixing PRs have been linked to the issue and can be reviewed to see what was done. In some cases we were able to ignore the warning, in other cases we created private methods that can be called from the constructor, as well as the virtual methods where it was safe to do so. In publicly exposed classes that needed the functionality we introduced factory classes that are passed to the constructor and can be extended to introduce custom behavior.

Issue #408, opened 2021-02-04
Fix initialization issue with DirectoryTaxonomyWriter


NightOwl888:
In [this test scenario](https://github.com/apache/lucenenet/blob/ae13dc4dfe0b8a2d305b9f8f860ac0849a16ac80/src/Lucene.Net.Tests.Facet/Taxonomy/Directory/TestDirectoryTaxonomyReader.cs#L298-L302), an alternate `IndexWriter` is being setup to be returned from the `OpenIndexWriter` method. However, since the constructor of the class calls `OpenIndexWriter` there is an issue where it cannot be extended directly as it was intended in Java. A constructor of a base class always runs before the subclass, so there is a chicken and egg problem with overriding `OpenIndexWriter`.

The current "solution" uses a static variable to set the `iw` variable before instantiating the class.

```c#
        private class DirectoryTaxonomyWriterAnonymousInnerClassHelper2 : DirectoryTaxonomyWriter
        {
            internal static IndexWriter iw = null;

            public DirectoryTaxonomyWriterAnonymousInnerClassHelper2(Directory dir) 
                : base(dir)
            {
            }

            protected override IndexWriter OpenIndexWriter(Directory directory, IndexWriterConfig config) 
            {   
                return iw;
            }
        }
```

While this makes the test pass, it clearly is going to be an issue for concurrent environments to do it this way.

We need to find a solution that allows using instance variables throughout, even if it means diverging a little from the original Lucene design.

## Potential Fix

One possible solution would be to add constructor overloads to `DirectoryTaxonomyReader` to allow the `OpenIndexWriter` method to be passed in as a `Func<IndexWriter, Directory, IndexWriterConfig>` and to disallow it to be overridden by base classes.

```c#
private readonly Func<IndexWriter, Directory, IndexWriterConfig) openIndexWriter;

public DirectoryTaxonomyReader(Directory directory, Func<IndexWriter, Directory, IndexWriterConfig> openIndexWriter)
{
    this.dir = directory;
    this.openIndexWriter = openIndexWriter;
 
    // Implementation....

    var x = OpenIndexWriter(y, z); // Call to OpenIndexWriter within constructor

    // Implementation....
}

// Remove "virtual" from the method so it cannot be overridden
protected IndexWriter OpenIndexWriter(Directory directory, IndexWriterConfig config)
{
    if (openIndexWriter != null)
    {
        return openIndexWriter(directory, config);
    }
    else
    {
        // Original implementation...
    }
}
```

[...]

PR #796, opened 2023-04-05
Fix virtual method calls from constructor

laimis:
This PR fixes issues with virtual calls being made from constructors. The issue originally reported by SonarCloud scans: https://sonarcloud.io/project/issues?resolved=false&rules=csharpsquid%3AS1699&id=apache_lucenenet and referenced in this issue: #670 

Starting with TernaryTree to get feedback if the approach is correct and then we can discuss further how to proceed (if to fix all occurrences of this in one shot or split it up in multiple PRs).

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/Shingle/ShingleFilter.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Shingle/ShingleFilter.cs
@@ -638,8 +638,11 @@ namespace Lucene.Net.Analysis.Shingle
             /// <see cref="outputUnigrams"/> = true.
             /// </para>
             /// </summary>
-            public virtual void Reset()
+            public void Reset()
             {
+                // LUCENENET specific - S1699 - marked non-virtual because calling
+                // virtual members from the constructor is not a safe operation in .NET.
+                
                 previousValue = value = minValue;
             }
 
--- a/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/TernaryTree.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/TernaryTree.cs
@@ -1,11 +1,10 @@
-﻿// Lucene version compatibility level 4.8.1
+﻿﻿// Lucene version compatibility level 4.8.1
 using Lucene.Net.Support;
 using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Text;
-using System.Threading;
 
 namespace Lucene.Net.Analysis.Compound.Hyphenation
 {
@@ -123,20 +122,10 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
 
         internal TernaryTree()
         {
-            // Init();  // LUCENENET specific - changes replacing calling virtual member from constructor
+            Init();
         }
 
-        // LUCENENET specific - changes replacing calling virtual member from constructor
-        private object nullIfNotInitialized;
-        protected void EnsureInitialized()
-        {
-            LazyInitializer.EnsureInitialized(ref nullIfNotInitialized, () => {
-                InitInternal();
-                return new object();
-            });
-        }
-
-        private void InitInternal()
+        protected virtual void Init()
         {
             m_root = (char)0;
             m_freenode = (char)1;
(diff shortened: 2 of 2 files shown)


NightOwl888 (changes requested):
For this particular class the constructor is internal. I just checked and it also has no subclasses.

So, the fix for `TernaryTree` would simply be to take the `virtual` modifier of the `Init()` method and then add a comment next to that method `// LUCENENET specific - S1699 - marked non-virtual because calling virtual members from the constructor is not a safe operation in .NET`.

Virtual is the default in Java, so there is no way to tell if they intended this to ever be subclassed, but if so, it will be by a class that we own and we can deal with the subclass when it arrives in Lucene.

Sorry if that wasn't clear. I think the `EnsureInitialized()` fix will work in several of the 53 cases of this, but we will have to analyze and apply the appropriate fix on each one. We only really have a problem if there is a public or protected constructor because there is no way to predict whether the class that subclasses it needs access to the fields of that subclass when inside the virtual member. But if the constructor is private or internal, we own everything and whatever subclasses that exist are all that will exist.

1. For internal constructors that have no subclasses, simply remove the virtual declaration and comment it.
2. For internal constructors that have subclasses, ensure there is at least 1 test covering the call to the constructor of each subclass to ensure it has all of the state it needs. Also, if any subclass has a protected or public constructor, we need to treat the issue the same as we would any other class with a public constructor and virtual member (including any superclasses that don't have public constructors).
3. For public/protected constructors that call virtual members, analyze the problem to see if the `EnsureInitialized()` fix will work. If not, consider whether an abstract factory can be made to replace the virtual member and inject that factory into the constructor. Or if you have any other ideas...

### Fixing Other Classes

While the above fix will suffice here, when applying the `LazyInitializer` fix on other classes and if we have more than one caller to the virtual member (as is the case here), the other callers should still call the virtual member by protecting it with a lock. This is so we don't have 2 threads calling the virtual member at the same time.

```c#
        private object nullIfNotInitialized;
        private object syncLock = new object();
        private bool isInitaialized; // We wouldn't need this, but the overload that allows passing a lock object requires this parameter

        public Init()
        {
            m_root = (char)0;
            m_freenode = (char)1;
            m_length = 0;
            m_lo = new char[BLOCK_SIZE];
            m_hi = new char[BLOCK_SIZE];
            m_eq = new char[BLOCK_SIZE];
            m_sc = new char[BLOCK_SIZE];
            m_kv = new CharVector();
        }

       protected void EnsureInitialized()
            LazyInitializer.EnsureInitialized(ref nullIfNotInitialized, ref isInitialized, ref syncLock, () => {
                Init();
                return new object();
            });
        }
```

And the other call to `Init()` should synchronize on the lock.

```c#
        public virtual void Balance()
        {
            EnsureInitialized(); // LUCENENET specific - changes replacing calling virtual member from constructor
            // System.out.print("Before root splitchar = ");
            // System.out.println(sc[root]);

            int i = 0, n = m_length;
            string[] k = new string[n];
            char[] v = new char[n];
            using (Enumerator iter = new Enumerator(this))
            {
                while (iter.MoveNext())
                {
                    v[i] = iter.Value;
                    k[i++] = iter.Current;
                }
            }
            lock (syncLock) // LUCENENET specific - synchronizing access to Init() because we need LazyInitializer
            {
                Init();
            }
            InsertBalanced(k, v, 0, n);

            // With uniform letter distribution sc[root] should be around 'm'
            // System.out.print("After root splitchar = ");
            // System.out.println(sc[root]);
        }
```

### Another Applicable Issue

BTW - one of these issues was previously reported in #408. It stood out because a test required setting the `IndexWriterConfig` but there was no way to do it. But IMO it would be best to start with the simpler ones to fix before tackling that one because it is much more complex.

NightOwl888 (approved):

laimis:
Thanks for that. I made the adjustments. I see that TernaryTree is being subclassed by HyphenationTree but in this case, it did not override Init and initialized its version of protected members in its own constructor.

src/Lucene.Net.Analysis.Common/Analysis/Shingle/ShingleFilter.cs
```diff
@@ -638,8 +638,11 @@ public virtual void Advance()
             /// <see cref="outputUnigrams"/> = true.
             /// </para>
             /// </summary>
```
laimis, line 640: ShingleFilter has a private CircularSequence class inside of it, so removing virtual designation is enough.

src/Lucene.Net.Analysis.Common/Analysis/Shingle/ShingleFilter.cs
```diff
@@ -638,8 +638,11 @@ public virtual void Advance()
...
-            public virtual void Reset()
+            public void Reset()
             {
+                // LUCENENET specific - S1699 - marked non-virtual because calling
+                // virtual members from the constructor is not a safe operation in .NET.
+
```
laimis, line 645: ShingleFilter has a private CircularSequence class inside of it, so removing virtual designation is enough.

laimis:
Closing this and will reopen a PR that starts with simple fixes.

PR #797, opened 2023-04-06
Remove virtual on a method that's being called from constructor

laimis:
This PR fixes some of the issues with virtual calls being made from constructors. The issues were reported by SonarCloud scans: https://sonarcloud.io/project/issues?resolved=false&rules=csharpsquid%3AS1699&id=apache_lucenenet and referenced in this issue: https://github.com/apache/lucenenet/issues/670

Particularly here we are addressing that have this issue and are either private or internal to Lucene and can be changed without impacting public API surface.


laimis:
To summarize:

- ShingleFilter.cs change is for CircularSequence class that's private inside of ShingleFilter class.
- Lucene45DocValuesProducer.cs change is for the internal class inside of it, Int64BinaryDocValues, that's internal and none of our stuff was overriding the virtual method.
- MinShouldMatchSumScorer.cs is an internal class and none of our stuff was overriding the method.

NightOwl888 (approved):
Thanks for the PR. This one looks good to merge.

PR #798, opened 2023-04-06
BREAKING: Remove virtual on a method that's being called from constructor in TernaryTree

laimis:

I have changed my mind on TernaryTree and believe that making Init not virtual is a better fix than introducing locking/lazy initialization.

TernaryTree is an internal class that's being subclassed by public HyphenationTree class. HyphenationTree does not call Init anywhere. This:

```c#
internal class TernaryTree
{
    public TernaryTree() { Init(); }

    protected virtual void Init() { ... }
}

public class HyphenationTree
{
    public HyphenationTree () 
}
```

Since HyphenationTree is public, by removing the virtual from Init() we are breaking the API. Someone could have subclassed HyphenationTree and overridden Init() method. But it has to be a very low odds occurrence, and if someone has done that, they will get an error and have to deal with initialization as opposed code compiling successfully but then failing in surprising ways once run.

Lucene.NET 3.0.3 has those classes commented out, no one would have done subclassing in that version. So people coming from 4.6.1 beta would be impacted, and we already have breaking changes that are part of beta.

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/HyphenationTree.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/HyphenationTree.cs
@@ -32,6 +32,15 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
     /// fast lookup. It provides the provides the method to hyphenate a word.
     /// <para/>
     /// This class has been taken from the Apache FOP project (http://xmlgraphics.apache.org/fop/). They have been slightly modified. 
+    ///
+    /// Lucene.NET specific note:
+    /// If you are going to extend this class by inheriting from it, you should be aware that the
+    /// base class TernaryTree initializes its state in the constructor by calling its protected Init() method.
+    /// If your subclass needs to initialize its own state, you add your own "Initialize()" method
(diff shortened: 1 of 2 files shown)


NightOwl888 (approved):
This took some consideration, but I have to agree that this is the best choice.

1. The subclass can still call `Init()`, no issue there.
2. If the subclass needs to initialize its own state, it can add its own `Initialize()` method. To work, the `Initialize()` method would need to be called both inside the constructor of the subclass and they would need to override the `Balance()` method and call `Initialize()` before the call to `base.Balance()`. It can use the data that is initialized in the base class after the call to `base.Balance()`.

So the only thing left would be to add some documentation to the class header about this decision. Except, it is so rarely likely to happen and if a user needed to initialize their own state they could probably work out how to do it on their own. I will leave it up to you whether to add docs.

I have labeled this PR as BREAKING so we can add it that way to the release notes.

rclabo:
I think that all things considered this is a fine approach and probably the one I'd have chosen as well.

laimis:
Thanks for taking a look! I added the comments about how to properly override. Will make sure to mention breaking when merging.

laimis:
Oh, no need to mention, it already does it. Nice!

PR #800, opened 2023-04-06
BREAKING: Make AbstractAnalysisFactory.Get non-virtual to avoid issues in constructors

laimis:

In this case, AbstractAnalysisFactory calls "Get" from the constructor. Get is virtual, but it does not feel like the intent here is to make that method extendable. It's a WIDELY extended class inside Lucene via TokenFilterFactory and all the classes that inherit from it, and none of them override Get implementations.


NightOwl888 (approved):
Looks good.

I think this was an intentional design decision, however, I suspect the abstract factories only exist to support SOLR and some Lucene tests. They don't really have much utility in .NET because of the odd constructors and weird way they need to be wired up with string dictionaries.

PR #801, opened 2023-04-06
BREAKING: Remove virtual on methods that are being called from constructors

laimis:

This time I focused on cases where a class is public, we are breaking the API, BUT, the method that was marked virtual was not calling any other virtual methods, was not being overridden in Lucene repo, and/or looked like a method that was not meant to be overridden to change the behavior of the class. Similarly to #798 , preferring explicit/clear/no surprises break in API over introducing locking/lazy initialization/additional allocations when the public consumers have a way to safely extend the class behavior.

This should eliminate the majority of "easy" changes, leaving us with a few "hairer" cases that we will tackle separately.

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
@@ -56,7 +56,9 @@ namespace Lucene.Net.Analysis.Util
             set => m_len = value;
         }
 
-        public virtual void Set(char[] arr, int end)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void Set(char[] arr, int end)
         {
             this.m_buf = arr;
             this.m_len = end;
--- a/src/Lucene.Net.Analysis.Common/Tartarus/Snowball/SnowballProgram.cs
+++ b/src/Lucene.Net.Analysis.Common/Tartarus/Snowball/SnowballProgram.cs
@@ -63,7 +63,9 @@ namespace Lucene.Net.Tartarus.Snowball
         /// <summary>
         /// Set the current string.
         /// </summary>
-        public virtual void SetCurrent(string value)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void SetCurrent(string value)
         {
             m_current = value.ToCharArray();
             m_cursor = 0;
--- a/src/Lucene.Net.Analysis.Common/Tartarus/Snowball/SnowballProgram.cs
+++ b/src/Lucene.Net.Analysis.Common/Tartarus/Snowball/SnowballProgram.cs
@@ -85,7 +85,11 @@ namespace Lucene.Net.Tartarus.Snowball
         /// </summary>
         /// <param name="text">character array containing input</param>
         /// <param name="length">valid length of text.</param>
-        public virtual void SetCurrent(char[] text, int length)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        // this one specifically is not called from constructor, the other overload is, but to be
+        // consistent, marking this as non-virtual as well.
+        public void SetCurrent(char[] text, int length)
         {
             m_current = text;
             m_cursor = 0;
--- a/src/Lucene.Net.Facet/Taxonomy/Directory/DirectoryTaxonomyReader.cs
+++ b/src/Lucene.Net.Facet/Taxonomy/Directory/DirectoryTaxonomyReader.cs
@@ -241,7 +241,9 @@ namespace Lucene.Net.Facet.Taxonomy.Directory
         /// <summary>
         /// Open the <see cref="DirectoryReader"/> from this <see cref="Directory"/>. 
         /// </summary>
-        protected virtual DirectoryReader OpenIndexReader(Directory directory)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        protected DirectoryReader OpenIndexReader(Directory directory)
         {
             return DirectoryReader.Open(directory);
         }
@@ -249,7 +251,9 @@ namespace Lucene.Net.Facet.Taxonomy.Directory
         /// <summary>
         /// Open the <see cref="DirectoryReader"/> from this <see cref="IndexWriter"/>. 
         /// </summary>
-        protected virtual DirectoryReader OpenIndexReader(IndexWriter writer)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        protected DirectoryReader OpenIndexReader(IndexWriter writer)
         {
             return DirectoryReader.Open(writer, false);
         }
--- a/src/Lucene.Net.Highlighter/VectorHighlight/FieldPhraseList.cs
+++ b/src/Lucene.Net.Highlighter/VectorHighlight/FieldPhraseList.cs
@@ -206,7 +206,9 @@ namespace Lucene.Net.Search.VectorHighlight
             }
         }
 
-        public virtual void AddIfNoOverlap(WeightedPhraseInfo wpi)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void AddIfNoOverlap(WeightedPhraseInfo wpi)
         {
             foreach (WeightedPhraseInfo existWpi in PhraseList)
             {
--- a/src/Lucene.Net.Queries/Function/ValueSourceScorer.cs
+++ b/src/Lucene.Net.Queries/Function/ValueSourceScorer.cs
@@ -48,7 +48,9 @@ namespace Lucene.Net.Queries.Function
 
         public virtual IndexReader Reader => m_reader;
 
-        public virtual void SetCheckDeletes(bool checkDeletes)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void SetCheckDeletes(bool checkDeletes)
         {
             this.m_checkDeletes = checkDeletes && m_reader.HasDeletions;
         }
--- a/src/Lucene.Net.QueryParser/Flexible/Standard/Nodes/NumericRangeQueryNode.cs
+++ b/src/Lucene.Net.QueryParser/Flexible/Standard/Nodes/NumericRangeQueryNode.cs
@@ -87,7 +87,9 @@ namespace Lucene.Net.QueryParsers.Flexible.Standard.Nodes
         /// <param name="lowerInclusive"><c>true</c> if the lower bound is inclusive, otherwise, <c>false</c></param>
         /// <param name="upperInclusive"><c>true</c> if the upper bound is inclusive, otherwise, <c>false</c></param>
         /// <param name="numericConfig">the <see cref="Config.NumericConfig"/> that represents associated with the upper and lower bounds</param>
-        public virtual void SetBounds(NumericQueryNode lower, NumericQueryNode upper,
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void SetBounds(NumericQueryNode lower, NumericQueryNode upper,
             bool lowerInclusive, bool upperInclusive, NumericConfig numericConfig)
         {
             if (numericConfig is null)
--- a/src/Lucene.Net.Spatial/Prefix/Tree/Cell.cs
+++ b/src/Lucene.Net.Spatial/Prefix/Tree/Cell.cs
@@ -175,7 +175,9 @@ namespace Lucene.Net.Spatial.Prefix.Tree
         public virtual bool IsLeaf => m_leaf;
 
         /// <summary>Note: not supported at level 0.</summary>
-        public virtual void SetLeaf()
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void SetLeaf()
         {
             if (Debugging.AssertsEnabled) Debugging.Assert(Level != 0);
             m_leaf = true;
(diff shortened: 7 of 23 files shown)


laimis:
Taking a look at the failures, somehow locally I did not trigger them...

NightOwl888 (changes requested):
This is going to take a few rounds of reviews to deal with. I left a few comments on the first 3 to give you some background.

Many of these were intended for extension by end users. For example a `Clear()` method could be implemented by a subclass to clear more than what the base class clears. But if you take a look at all collection types in .NET, none of their members are virtual. Of course, it is a PITA because it means you basically have to write collections from scratch instead of simply extending them.

src/Lucene.Net.Suggest/Spell/SpellChecker.cs
```diff
@@ -142,8 +142,9 @@ public SpellChecker(Directory spellIndex, IStringDistance sd, IComparer<SuggestW
         /// <param name="spellIndexDir"> the spell directory to use </param>
         /// <exception cref="ObjectDisposedException"> if the Spellchecker is already closed </exception>
         /// <exception cref="IOException"> if spellchecker can not open the directory </exception>
-        // TODO: we should make this final as it is called in the constructor
-        public virtual void SetSpellIndex(Directory spellIndexDir)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void SetSpellIndex(Directory spellIndexDir)
```
NightOwl888, line 147: I was thinking this one is a good candidate for `LazyInitializer`. Then I saw the comment that was added by the Lucene team:

```c#
// TODO: we should make this final as it is called in the constructor
```

I checked and as of Lucene 9.5.0 they still haven't changed that. But being that they intend to, I think we can keep this change.

src/Lucene.Net.Queries/Function/ValueSourceScorer.cs
```diff
@@ -48,7 +48,9 @@ protected internal ValueSourceScorer(IndexReader reader, FunctionValues values)
 
         public virtual IndexReader Reader => m_reader;
 
-        public virtual void SetCheckDeletes(bool checkDeletes)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void SetCheckDeletes(bool checkDeletes)
```
NightOwl888, line 53: Let's move this logic to a private `SetCheckDeletesInternal()` method that is called by the constructor and the virtual `SetCheckDeletes()` method. This just needs some docs to tell overriders that their overridden logic won't run in the constructor, so they will need to set this flag appropriately in their own constructor (it's fine to suggest adding a private method that is called in both places).

src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
```diff
@@ -56,7 +56,9 @@ public virtual int Length
             set => m_len = value;
         }
 
-        public virtual void Set(char[] arr, int end)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void Set(char[] arr, int end)
```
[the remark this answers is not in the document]
laimis: Fixed here https://github.com/apache/lucenenet/pull/803

src/Lucene.Net.Analysis.Common/Tartarus/Snowball/SnowballProgram.cs
```diff
@@ -63,7 +63,9 @@ protected SnowballProgram()
         /// <summary>
         /// Set the current string.
         /// </summary>
-        public virtual void SetCurrent(string value)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void SetCurrent(string value)
```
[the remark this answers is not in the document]
laimis: Fixed here: https://github.com/apache/lucenenet/pull/802

src/Lucene.Net/Util/SentinelIntSet.cs
```diff
@@ -89,7 +89,9 @@ public SentinelInt32Set(int size, int emptyVal)
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public virtual void Clear()
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void Clear()
```
NightOwl888, line 94: This one can be made into private `ClearInternal()` that is called by the constructor and this virtual `Clear()` method.

src/Lucene.Net/Search/DisjunctionMaxQuery.cs
```diff
@@ -92,7 +92,9 @@ public DisjunctionMaxQuery(ICollection<Query> disjuncts, float tieBreakerMultipl
         /// <summary>
         /// Add a subquery to this disjunction </summary>
         /// <param name="query"> The disjunct added </param>
-        public virtual void Add(Query query)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void Add(Query query)
```
NightOwl888, line 97: This one can be made into private `AddInternal()` called by the constructor and virtual `Add()`.

src/Lucene.Net.Spatial/Prefix/Tree/Cell.cs
```diff
@@ -175,7 +175,9 @@ private void B_fixLeaf()
         public virtual bool IsLeaf => m_leaf;
 
         /// <summary>Note: not supported at level 0.</summary>
-        public virtual void SetLeaf()
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void SetLeaf()
```
NightOwl888, line 180: This one can be made into private `SetLeafInternal()` called by `B_fixLeaf()` and virtual `SetLeaf()`. In the subclass constructor, the user has access to both `m_leaf` and `IsLeaf` so they can read or set the value on construction after the base class. Public `SetLeaf()` calls can then run whatever business logic that happens when it is set.

NightOwl888 (changes requested):

src/Lucene.Net.Facet/Taxonomy/Directory/DirectoryTaxonomyReader.cs
```diff
@@ -241,15 +241,19 @@ protected override TaxonomyReader DoOpenIfChanged()
         /// <summary>
         /// Open the <see cref="DirectoryReader"/> from this <see cref="Directory"/>. 
         /// </summary>
-        protected virtual DirectoryReader OpenIndexReader(Directory directory)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        protected DirectoryReader OpenIndexReader(Directory directory)
```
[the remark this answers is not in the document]
laimis: OK, I will leave this one alone here and will revisit it separately.

src/Lucene.Net.Queries/Function/ValueSourceScorer.cs
laimis: Made the fix here: https://github.com/apache/lucenenet/pull/805

src/Lucene.Net/Util/SentinelIntSet.cs
laimis: Fixed with https://github.com/apache/lucenenet/pull/806

src/Lucene.Net/Search/DisjunctionMaxQuery.cs
laimis: Fixed with https://github.com/apache/lucenenet/pull/807

src/Lucene.Net.Spatial/Prefix/Tree/Cell.cs
laimis: Fixed with https://github.com/apache/lucenenet/pull/808

laimis:
Thank you for the review. I went ahead and addressed the issues with separate PRs. I left out DirectoryTaxonomyReader.cs and will take it in another PR. Closing this one without merging.

src/Lucene.Net.QueryParser/Flexible/Standard/Nodes/NumericRangeQueryNode.cs
```diff
@@ -87,7 +87,9 @@ private static NumericType GetNumericDataType(J2N.Numerics.Number number)
         /// <param name="lowerInclusive"><c>true</c> if the lower bound is inclusive, otherwise, <c>false</c></param>
         /// <param name="upperInclusive"><c>true</c> if the upper bound is inclusive, otherwise, <c>false</c></param>
         /// <param name="numericConfig">the <see cref="Config.NumericConfig"/> that represents associated with the upper and lower bounds</param>
-        public virtual void SetBounds(NumericQueryNode lower, NumericQueryNode upper,
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void SetBounds(NumericQueryNode lower, NumericQueryNode upper,
```
NightOwl888, line 92: This one can be replaced by a private `SetBoundsInternal()` method called by the constructor and public virtual `SetBounds()`.

BTW - Let's add some documentation on `SetBounds()` letting users know that if they override the method they would also need to add the same logic into the constructor to match Lucene.

src/Lucene.Net.Spatial/Prefix/Tree/Cell.cs
NightOwl888: BTW - Let's add some documentation on `SetLeaf()` letting users know that if they override the method they would also need to add the same logic into the constructor to match Lucene.

src/Lucene.Net/Search/DisjunctionMaxQuery.cs
NightOwl888: BTW - Let's add some documentation on `Add()` letting users know that if they override the method they would also need to add the same logic into the constructor to match Lucene.

src/Lucene.Net/Util/SentinelIntSet.cs
NightOwl888: BTW - Let's add some documentation on `Clear()` letting users know that if they override the method they would also need to add the same logic into the constructor to match Lucene.

src/Lucene.Net/Store/ByteArrayDataOutput.cs
```diff
@@ -51,12 +51,16 @@ public ByteArrayDataOutput()
             Reset(BytesRef.EMPTY_BYTES);
         }
 
-        public virtual void Reset(byte[] bytes)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void Reset(byte[] bytes)
```
NightOwl888, line 56: This can be changed to `ResetInternal()` called by both the constructor and the virtual `Reset()` methods.

BTW - Let's add some documentation on `Reset()` letting users know that if they override the method they would also need to add the same logic into the constructor to match Lucene.

Also, let's add some null guard clauses for `bytes` that throw `ArgumentNullException`.

src/Lucene.Net/Search/Sort.cs
```diff
@@ -139,13 +139,17 @@ public Sort(params SortField[] fields)
         }
 
         /// <summary>Sets the sort to the given criteria. </summary>
-        public virtual void SetSort(SortField field)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        public void SetSort(SortField field)
         {
```
NightOwl888, line 145: These can be made into `SetSortInternal()` called by the constructor and virtual `SetSort()`.

BTW - Let's add some documentation on both overloads of `SetSort()` letting users know that if they override the method they would also need to add the same logic into the constructor to match Lucene.

src/Lucene.Net/Search/IndexSearcher.cs
```diff
@@ -161,7 +161,9 @@ public IndexSearcher(IndexReaderContext context)
         /// Each <see cref="LeafSlice"/> is executed in a single thread. By default there
         /// will be one <see cref="LeafSlice"/> per leaf (<see cref="AtomicReaderContext"/>).
         /// </summary>
-        protected virtual LeafSlice[] Slices(IList<AtomicReaderContext> leaves)
+        // LUCENENET specific - S1699 - marked non-virtual because calling
+        // virtual members from the constructor is not a safe operation in .NET
+        protected LeafSlice[] Slices(IList<AtomicReaderContext> leaves)
```
NightOwl888, line 166: This one can be made into `GetSlicesInternal()` called by the constructor and virtual `GetSlices()` method. Note adding a verb here is intentional to indicate this is an action. And since it is protected, this break won't affect many users. Just be sure to add a note that we renamed this from Lucene.

BTW - Let's add some documentation on `GetSlices()` letting users know that if they override the method they would also need to add the same logic into the constructor to match Lucene.

NightOwl888 (commented):
I added a few more review comments.

src/Lucene.Net.Highlighter/VectorHighlight/FieldPhraseList.cs
laimis, line 1: Fixed in https://github.com/apache/lucenenet/pull/811

src/Lucene.Net.QueryParser/Flexible/Standard/Nodes/NumericRangeQueryNode.cs
laimis: Fixed in https://github.com/apache/lucenenet/pull/812

src/Lucene.Net/Search/Sort.cs
laimis: Fixed in https://github.com/apache/lucenenet/pull/813

src/Lucene.Net/Store/ByteArrayDataOutput.cs
laimis: Fixed in https://github.com/apache/lucenenet/pull/814

src/Lucene.Net/Search/IndexSearcher.cs
laimis: Fixed in https://github.com/apache/lucenenet/pull/818

PR #803, opened 2023-04-07
BREAKING: remove virtual method call from constructor in OpenStringBuilder

laimis:

This one focuses on OpenStringBuilder. The pattern here is to create a private method that the constructor can use and virtual methods can use as well.


laimis:
Merging this one since it was similar to SnowballProgram.

PR #804, opened 2023-04-07
BREAKING: remove virtual designation as it's called from constructor

laimis:

This one focuses on SpellChecker. We are simply removing the virtual designation. It seems like Lucene team has been planning this as well based on TODO: comment above the method but has not pulled the trigger. We are going to go ahead and do that.


src/Lucene.Net.Suggest/Spell/SpellChecker.cs
```diff
@@ -142,8 +142,9 @@ public SpellChecker(Directory spellIndex, IStringDistance sd, IComparer<SuggestW
         /// <param name="spellIndexDir"> the spell directory to use </param>
         /// <exception cref="ObjectDisposedException"> if the Spellchecker is already closed </exception>
         /// <exception cref="IOException"> if spellchecker can not open the directory </exception>
-        // TODO: we should make this final as it is called in the constructor
```
NightOwl888, line 145: We should probably keep this Lucene comment here. Or at least explain that we made this decision because it was intended this way by the Lucene team.

laimis: Sounds good, adding it now

PR #809, opened 2023-04-08
BREAKING: remove virtual call from the constructor for RollingBuffer and related classes

laimis:

This one focuses on RollingBuffer and its subclasses. Instead of providing a virtual NewInstance to override, we accept a factory method in the constructor and use it whenever new instance is needed.

### DIFF
--- a/src/Lucene.Net.Tests/Util/TestRollingBuffer.cs
+++ b/src/Lucene.Net.Tests/Util/TestRollingBuffer.cs
@@ -25,7 +25,7 @@ namespace Lucene.Net.Util
     [TestFixture]
     public class TestRollingBuffer : LuceneTestCase
     {
-        private class Position : RollingBuffer.IResettable
+        private class Position : IResettable
         {
             public int Pos { get; set; }
 
@@ -92,13 +92,13 @@ namespace Lucene.Net.Util
         {
             // LUCENENET specific - removed NewPosition override and using factory instead
             public RollingBufferAnonymousClass()
-                : base(RollingBufferAnonymousClassFactory.Default)
+                : base(PositionFactory.Default)
             {
             }
 
-            private class RollingBufferAnonymousClassFactory : IRollingBufferItemFactory<Position>
+            private class PositionFactory : IRollingBufferItemFactory<Position>
             {
-                public static readonly RollingBufferAnonymousClassFactory Default = new RollingBufferAnonymousClassFactory();
+                public static readonly PositionFactory Default = new PositionFactory();
 
                 public Position Create(object rollingBuffer)
                 {
--- a/src/Lucene.Net.Tests/Util/TestRollingBuffer.cs
+++ b/src/Lucene.Net.Tests/Util/TestRollingBuffer.cs
@@ -90,16 +90,22 @@ namespace Lucene.Net.Util
 
         private sealed class RollingBufferAnonymousClass : RollingBuffer<Position>
         {
+            // LUCENENET specific - removed NewPosition override and using factory instead
             public RollingBufferAnonymousClass()
-                : base(NewInstanceFunc)
+                : base(RollingBufferAnonymousClassFactory.Default)
             {
             }
 
-            public static Position NewInstanceFunc()
+            private class RollingBufferAnonymousClassFactory : IRollingBufferItemFactory<Position>
             {
-                Position pos = new Position();
-                pos.Pos = -1;
-                return pos;
+                public static readonly RollingBufferAnonymousClassFactory Default = new RollingBufferAnonymousClassFactory();
+
+                public Position Create(object rollingBuffer)
+                {
+                    Position pos = new Position();
+                    pos.Pos = -1;
+                    return pos;
+                }
             }
         }
     }
--- a/src/Lucene.Net.Tests/Util/TestRollingBuffer.cs
+++ b/src/Lucene.Net.Tests/Util/TestRollingBuffer.cs
@@ -101,13 +101,6 @@ namespace Lucene.Net.Util
                 pos.Pos = -1;
                 return pos;
             }
-
-            protected override Position NewInstance()
-            {
-                Position pos = new Position();
-                pos.Pos = -1;
-                return pos;
-            }
         }
     }
 }
--- a/src/Lucene.Net/Util/RollingBuffer.cs
+++ b/src/Lucene.Net/Util/RollingBuffer.cs
@@ -44,7 +44,7 @@ namespace Lucene.Net.Util
     /// </summary>
     public class RollingBufferItemFactory<T> : IRollingBufferItemFactory<T> where T : new()
     {
-        public static readonly RollingBufferItemFactory<T> Default = new RollingBufferItemFactory<T>();
+        public static RollingBufferItemFactory<T> Default { get; } = new RollingBufferItemFactory<T>();
         public virtual T Create(object rollingBuffer)
         {
             return new T();
--- a/src/Lucene.Net/Util/RollingBuffer.cs
+++ b/src/Lucene.Net/Util/RollingBuffer.cs
@@ -72,14 +72,14 @@ namespace Lucene.Net.Util
         // How many valid Position are held in the
         // array:
         private int count;
-        private IRollingBufferItemFactory<T> factory;
+        private IRollingBufferItemFactory<T> itemFactory;
 
-        protected RollingBuffer(IRollingBufferItemFactory<T> factory)
+        protected RollingBuffer(IRollingBufferItemFactory<T> itemFactory)
         {
-            this.factory = factory; // LUCENENET specific - storing factory for usage in class
+            this.itemFactory = itemFactory; // LUCENENET specific - storing factory for usage in class
             for (int idx = 0; idx < buffer.Length; idx++)
             {
-                buffer[idx] = factory.Create(this);
+                buffer[idx] = itemFactory.Create(this);
             }
         }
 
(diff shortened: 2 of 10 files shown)


src/Lucene.Net/Util/RollingBuffer.cs
```diff
@@ -57,25 +58,17 @@ public abstract class RollingBuffer<T>
...
-        protected RollingBuffer()
-        {
-            for (var idx = 0; idx < buffer.Length; idx++)
-            {
-                buffer[idx] = NewInstance(); // TODO GIVE ROLLING BUFFER A DELEGATE FOR NEW INSTANCE
-            }
-        }
+        private Func<T> factory;
 
         protected RollingBuffer(Func<T> factory)
```
NightOwl888, line 63: Great work.

However after considering the change a little more, I thought of a way that this is more limited than it was in Lucene. Since a constructor injected `Func<T>` is static, we can no longer reference any state in the subclass.

If we use an abstract factory instead, we can pass in the current instance as an argument.

```c#
public interface IRollingBufferItemFactory<T>
{
    T Create<TRollingBuffer>(TRollingBuffer rollingBuffer) where TRollingBuffer : RollingBuffer;
}
```

And then call the factory like:

```c#
            for (int idx = 0; idx < buffer.Length; idx++)
            {
                buffer[idx] = rollingBufferItemFactory.Create(this);
            }
```

Then the factory can do things based on the state of the subclass. I think if it is declared like above, they won't need to do a cast to see the subclass members.

NightOwl888: I mocked up the above approach in a couple of different ways.

OptionA is more complicated to implement for end users because they have to pass an extra generic type to both `RollingBuffer `and `IRollingBufferItemFactory`. However, inside the factory there is no casting required to see the state of the current buffer instance, as there is no cast to see the state in Java.

OptionB is simpler to implement, but requires a cast from object to the buffer type. Although it is guaranteed to be that type, using an `is` cast as in my example makes it ambiguous what to do in the `else` case. 

Some might consider OptionA to be overengineered. But given the rarity of the case that end users will implement their own `RollingBuffer` I personally think it is fine. And since it gives you a strong type for the buffer instance, it eliminates casting or any confusion over what type we are receiving or ambiguity when casting using `is`.

If we go with OptionA, there is an extra `IRollingBuffer` interface needed to constrain us to a rolling buffer type without referring to any of its generic closing types. In this case we probably can add the public members that don't include `TItem` here instead of having an empty interface.

```c#
    internal class Program
    {
        static void Main(string[] args)
        {
            var a = new OptionA.MyBuffer();
            var b = new OptionB.MyBuffer();

            Console.WriteLine("Hello World!");
        }
    }

    namespace OptionA
    {

        public class Position
        {
        }

        public abstract class RollingBuffer<TItem, TRollingBuffer> : IRollingBuffer
            where TRollingBuffer : IRollingBuffer
        {
            protected RollingBuffer(IRollingBufferItemFactory<TItem, TRollingBuffer> itemFactory)
            {
                TItem[] buffer = new TItem[10];

                for (int idx = 0; idx < buffer.Length; idx++)
                {
                    // This intermediate cast to object is required because of the generics, but it will never fail and is hidden from users.
                    buffer[idx] = itemFactory.Create((TRollingBuffer)(object)this);
                }
            }
        }

        public interface IRollingBuffer // Interface required just so we can "own" the rolling buffer without a generic closing type
        {
        }

        public interface IRollingBufferItemFactory<out TItem, in TRollingBuffer>
             where TRollingBuffer : IRollingBuffer
        {
            TItem Create(TRollingBuffer rollingBuffer);
        }

        public class MyBuffer : RollingBuffer<Position, MyBuffer>
        {
            private object myState = new object();

            public MyBuffer()
                : base(MyPositionFactory.Default)
            {
            }

            public class MyPositionFactory : IRollingBufferItemFactory<Position, MyBuffer>
            {
                public static MyPositionFactory Default { get; } = new MyPositionFactory();

                public Position Create(MyBuffer rollingBuffer)
                {
                    var state = rollingBuffer.myState; // Now we can see the state

                    return new Position();
                }
            }
        }

    }

    namespace OptionB
    {
        public class Position
        {
        }

        public abstract class RollingBuffer<T>
        {
            protected RollingBuffer(IRollingBufferItemFactory<T> itemFactory)
            {
                T[] buffer = new T[10];

                for (int idx = 0; idx < buffer.Length; idx++)
                {
                    buffer[idx] = itemFactory.Create(this);
                }
            }
        }

        public interface IRollingBufferItemFactory<out T>
        {
            T Create(object rollingBuffer);
        }

        public class MyBuffer : RollingBuffer<Position>
        {
            private object myState = new object();

            public MyBuffer()
                : base(MyPositionFactory.Default)
            {
            }

            public class MyPositionFactory : IRollingBufferItemFactory<Position>
            {
                public static MyPositionFactory Default { get; } = new MyPositionFactory();

                public Position Create(object rollingBuffer)
                {
                    if (rollingBuffer is MyBuffer buf)
                    {
                        var state = buf.myState; // Now we can see the state
                    }
                    // The else cannot technically happen, but if we return in the above block, we would need a return or throw after this line, which might not make any sense

                    return new Position();
                }
            }
        }
    }
```

laimis:
After some discussions internally, we decided to go with option B. We will accept a IRollingBufferItemFactory interface and use it to create new items. The callers can then use default (which basically just does new()) or provide their own.

src/Lucene.Net/Util/RollingBuffer.cs
```diff
@@ -37,12 +36,35 @@ public interface IResettable
...
+    public interface IRollingBufferItemFactory<out T>
+    {
+        T Create(object rollingBuffer);
+    }
+
+    /// <summary>
+    /// LUCENENET specific class that provides default implementation for
+    /// <see cref="IRollingBufferItemFactory{T}"/>.
+    /// </summary>
+    public class RollingBufferItemFactory<T> : IRollingBufferItemFactory<T> where T : new()
```
NightOwl888, line 52: I had considered what would be the case if we use the `new()` constraint here. It works for all existing cases, but unfortunately it forces a default constructor into implementations of this where it may not make any sense. I realize this saves some factory implementations, but IMO forcing this constructor to exist isn't worth the tradeoff. We can still share a `PositionFactory` implementation in several cases.

src/Lucene.Net/Util/RollingBuffer.cs
```diff
@@ -37,12 +36,35 @@ public interface IResettable
         }
     }
 
+    /// <summary>
+    /// LUCENENET specific interface to allow overriding rolling buffer item creation
+    /// without having to call virtual methods from the constructor
+    /// </summary>
+    public interface IRollingBufferItemFactory<out T>
```
NightOwl888, line 43: Please eliminate the above `RollingBuffer` static class and move `IResettable` up here to the `Lucene.Net.Util` namespace. Usages of `IResettable` aren't constrained to `RollingBuffer` usage.

src/Lucene.Net/Util/RollingBuffer.cs
```diff
@@ -57,25 +79,17 @@ public abstract class RollingBuffer<T>
...
-        protected RollingBuffer()
-        {
-            for (var idx = 0; idx < buffer.Length; idx++)
-            {
-                buffer[idx] = NewInstance(); // TODO GIVE ROLLING BUFFER A DELEGATE FOR NEW INSTANCE
-            }
-        }
-
-        protected RollingBuffer(Func<T> factory)
+        protected RollingBuffer(IRollingBufferItemFactory<T> factory)
```
NightOwl888, line 84: Let's change the variable name to `itemFactory` to make it more clear what it creates. And on other classes where it is injected, also.

src/Lucene.Net.Tests/Util/TestRollingBuffer.cs
```diff
@@ -90,23 +90,22 @@ public virtual void Test()
...
         {
+            // LUCENENET specific - removed NewPosition override and using factory instead
             public RollingBufferAnonymousClass()
-                : base(NewInstanceFunc)
+                : base(RollingBufferAnonymousClassFactory.Default)
             {
             }
 
-            public static Position NewInstanceFunc()
+            private class RollingBufferAnonymousClassFactory : IRollingBufferItemFactory<Position>
```
NightOwl888, line 99: Maybe just call this `PositionFactory`?

NightOwl888 (changes requested):

src/Lucene.Net/Util/RollingBuffer.cs
NightOwl888: Oh, never mind. This is the implementation, not the interface. This is fine.

laimis: Question, since we are not forcing new() on the RollingBuffer level, is that still an issue? We only force it if someone wants to use the default RollingBufferItemFactory. But that factory is not forced upon anyone. What do you think?

NightOwl888: Yes, that is the conclusion I came to. You can leave this as is.

src/Lucene.Net/Util/RollingBuffer.cs
laimis: Done, please take a look now

src/Lucene.Net/Util/RollingBuffer.cs
```diff
@@ -23,17 +22,32 @@ namespace Lucene.Net.Util
...
+        T Create(object rollingBuffer);
+    }
+
+    /// <summary>
+    /// LUCENENET specific class that provides default implementation for
+    /// <see cref="IRollingBufferItemFactory{T}"/>.
+    /// </summary>
+    public class RollingBufferItemFactory<T> : IRollingBufferItemFactory<T> where T : new()
+    {
+        public static readonly RollingBufferItemFactory<T> Default = new RollingBufferItemFactory<T>();
```
NightOwl888, line 47: Since this is public, let's make it a property. That is the way Microsoft declares these.

```c#
public static RollingBufferItemFactory<T> Default { get; } = new RollingBufferItemFactory<T>();
```

laimis: Done!

NightOwl888 (approved):
Well, that took some head scratching, but together we worked it out. Good work!

laimis:
Yeah! That was a good session of exploring solutions.

PR #814, opened 2023-04-09
BREAKING: remove virtual call from the constructor for ByteArrayDataOutput

laimis:

This one focuses on ByteArrayDataOuput, creating a private method that constructor as well as public virtual methods can call.

### DIFF
--- a/src/Lucene.Net/Store/ByteArrayDataOutput.cs
+++ b/src/Lucene.Net/Store/ByteArrayDataOutput.cs
@@ -1,5 +1,6 @@
 ﻿using Lucene.Net.Diagnostics;
 using Lucene.Net.Support;
+using System;
 
 namespace Lucene.Net.Store
 {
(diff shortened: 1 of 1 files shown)


src/Lucene.Net/Store/ByteArrayDataOutput.cs
```diff
@@ -38,26 +39,51 @@ public class ByteArrayDataOutput : DataOutput
...
-        public virtual void Reset(byte[] bytes, int offset, int len)
+        // LUCENENET specific - created a private method that can be called
+        // from the constructor and the Reset methods to avoid virtual method
+        // calls in the constructor.
+        private void ResetInternal(byte[] bytes, int offset, int len)
         {
+            // LUCENENET: Added guard clause
+            if (bytes is null)
+                throw new ArgumentNullException(nameof(bytes));
+
```
NightOwl888, line 86: We should probably also add the [standard bounds checking guard clauses](https://github.com/NightOwl888/J2N/blob/3d3c7211a39395c00aa3dfd1cdf9c01501391d87/src/J2N/Collections/Generic/List.SubList.cs#L185-L190). Messages are [here](https://github.com/NightOwl888/J2N/blob/3d3c7211a39395c00aa3dfd1cdf9c01501391d87/src/J2N/Resources/Strings.resx).

It would be great to rename these arguments `startIndex` and `length`, also, but we should probably do a sweep of the codebase to get them all later rather than doing them one class at a time.

NightOwl888 (changes requested):

src/Lucene.Net/Store/ByteArrayDataOutput.cs
```diff
@@ -38,26 +39,51 @@ public class ByteArrayDataOutput : DataOutput
...
-        }
+        /// <summary>
+        /// 
+        /// NOTE: When overriding this method, be aware that the constructor of this class calls 
+        /// a private method and not this virtual method. So if you need to override
+        /// the behavior during the initialization, call your own private method from the constructor
+        /// with whatever custom behavior you need.
+        /// </summary>
+        public virtual void Reset(byte[] bytes) =>
+            ResetInternal(bytes, 0, bytes.Length);
```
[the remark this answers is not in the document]
laimis: Went with bytes?.Length ?? 0

src/Lucene.Net/Store/ByteArrayDataOutput.cs
laimis: Added additional checks, left the rename out if we are going to do a sweep of this later

src/Lucene.Net/Store/ByteArrayDataOutput.cs
```diff
@@ -38,26 +39,57 @@ public class ByteArrayDataOutput : DataOutput
 
         public ByteArrayDataOutput(byte[] bytes)
         {
-            Reset(bytes);
+            // LUCENENET: Changed to call private method to avoid virtual method call in constructor
+            ResetInternal(bytes, 0, bytes.Length);
```
NightOwl888, line 43: Looks like we missed a potential `NullReferenceException` here. The rest of this looks good.

NightOwl888 (changes requested):

src/Lucene.Net/Store/ByteArrayDataOutput.cs
laimis: arrg, good catch, fixed it

PR #816, opened 2023-04-09
Supress sonar cloud warnings about constructor making virtual call

laimis:

This one includes all the cases where we will ignore the warning and accept the code as is due to the fix being too invasive and not justifiable (e.g. classes are internal to Lucene.NET)


NightOwl888 (approved):
I had to wait until I had a minute to pull this down so I could check with the SonarLint plugin whether these work. And they do.

PR #819, opened 2023-04-11
BREAKING: Fix/virtual calls from constructors BlockTreeTerms reader and writer

laimis:

This one focuses on BlocktreeTermsReader and Writer. We are providing a way to pass state by the subclasses and then ignoring the sonar error about virtual methods being called.

### DIFF
--- a/src/Lucene.Net.Codecs/Pulsing/Pulsing41PostingsFormat.cs
+++ b/src/Lucene.Net.Codecs/Pulsing/Pulsing41PostingsFormat.cs
@@ -35,7 +35,7 @@ namespace Lucene.Net.Codecs.Pulsing
 
         /// <summary>Inlines docFreq=<paramref name="freqCutoff"/> terms, otherwise uses the normal "Lucene41" format.</summary>
         public Pulsing41PostingsFormat(int freqCutoff) 
-            : this(freqCutoff, BlockTreeTermsWriter.DEFAULT_MIN_BLOCK_SIZE, BlockTreeTermsWriter.DEFAULT_MAX_BLOCK_SIZE)
+            : this(freqCutoff, BlockTreeTermsWriter<object>.DEFAULT_MIN_BLOCK_SIZE, BlockTreeTermsWriter<object>.DEFAULT_MAX_BLOCK_SIZE)
         {
         }
 
--- a/src/Lucene.Net.Codecs/Pulsing/Pulsing41PostingsFormat.cs
+++ b/src/Lucene.Net.Codecs/Pulsing/Pulsing41PostingsFormat.cs
@@ -35,7 +35,7 @@ namespace Lucene.Net.Codecs.Pulsing
 
         /// <summary>Inlines docFreq=<paramref name="freqCutoff"/> terms, otherwise uses the normal "Lucene41" format.</summary>
         public Pulsing41PostingsFormat(int freqCutoff) 
-            : this(freqCutoff, BlockTreeTermsWriter<object>.DEFAULT_MIN_BLOCK_SIZE, BlockTreeTermsWriter<object>.DEFAULT_MAX_BLOCK_SIZE)
+            : this(freqCutoff, BlockTreeTermsWriter.DEFAULT_MIN_BLOCK_SIZE, BlockTreeTermsWriter.DEFAULT_MAX_BLOCK_SIZE)
         {
         }
 
--- a/src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
+++ b/src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
@@ -131,7 +131,7 @@ namespace Lucene.Net.Codecs
         /// so the overrides of those methods won't specifically need to use this field (although they could for consistency).
         /// </param>
         [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
-        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Internal class")]
+        [SuppressMessage("CodeQuality", "S1699:Required for continuity with Lucene's design", Justification = "Internal class")]
         public BlockTreeTermsReader(Directory dir, FieldInfos fieldInfos, SegmentInfo info, PostingsReaderBase postingsReader, IOContext ioContext, string segmentSuffix, int indexDivisor, TSubclassState subclassState)
         {
             // LUCENENET specific - added state parameter that subclasses
--- a/src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
+++ b/src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
@@ -114,13 +114,18 @@ namespace Lucene.Net.Codecs
 
         private readonly int version;
 
-        protected readonly object subclassState;
+        protected readonly object m_subclassState;
 
         /// <summary>
         /// Sole constructor. </summary>
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Internal class")]
         public BlockTreeTermsReader(Directory dir, FieldInfos fieldInfos, SegmentInfo info, PostingsReaderBase postingsReader, IOContext ioContext, string segmentSuffix, int indexDivisor, object subclassState = default)
         {
-            this.subclassState = subclassState;
+            // LUCENENET specific - added state parameter that subclasses
+            // can use to keep track of state and use it in their own virtual
+            // methods that are called by this constructor
+            this.m_subclassState = subclassState;
 
             NO_OUTPUT = fstOutputs.NoOutput;
             this.postingsReader = postingsReader;
--- a/src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
+++ b/src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
@@ -86,7 +86,7 @@ namespace Lucene.Net.Codecs
     /// <para/>
     /// @lucene.experimental
     /// </summary>
-    public class BlockTreeTermsReader<T> : FieldsProducer
+    public class BlockTreeTermsReader<TSubclassState> : FieldsProducer
     {
         // Open input to the main terms dict file (_X.tib)
 #pragma warning disable CA2213 // Disposable fields should be disposed
@@ -114,13 +114,13 @@ namespace Lucene.Net.Codecs
 
         private readonly int version;
 
-        protected readonly T m_subclassState;
+        protected readonly TSubclassState m_subclassState;
 
         /// <summary>
         /// Sole constructor. </summary>
         [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
         [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Internal class")]
-        public BlockTreeTermsReader(Directory dir, FieldInfos fieldInfos, SegmentInfo info, PostingsReaderBase postingsReader, IOContext ioContext, string segmentSuffix, int indexDivisor, T subclassState)
+        public BlockTreeTermsReader(Directory dir, FieldInfos fieldInfos, SegmentInfo info, PostingsReaderBase postingsReader, IOContext ioContext, string segmentSuffix, int indexDivisor, TSubclassState subclassState)
         {
             // LUCENENET specific - added state parameter that subclasses
             // can use to keep track of state and use it in their own virtual
@@ -552,7 +552,7 @@ namespace Lucene.Net.Codecs
         /// BlockTree's implementation of <see cref="GetTerms(string)"/>. </summary>
         public sealed class FieldReader : Terms
         {
-            private readonly BlockTreeTermsReader<T> outerInstance;
+            private readonly BlockTreeTermsReader<TSubclassState> outerInstance;
 
             internal readonly long numTerms;
             internal readonly FieldInfo fieldInfo;
@@ -567,7 +567,7 @@ namespace Lucene.Net.Codecs
             internal readonly FST<BytesRef> index;
             //private boolean DEBUG;
 
-            internal FieldReader(BlockTreeTermsReader<T> outerInstance, FieldInfo fieldInfo, long numTerms, BytesRef rootCode, long sumTotalTermFreq, long sumDocFreq, int docCount, long indexStartFP, int longsSize, IndexInput indexIn)
+            internal FieldReader(BlockTreeTermsReader<TSubclassState> outerInstance, FieldInfo fieldInfo, long numTerms, BytesRef rootCode, long sumTotalTermFreq, long sumDocFreq, int docCount, long indexStartFP, int longsSize, IndexInput indexIn)
             {
                 this.outerInstance = outerInstance;
                 if (Debugging.AssertsEnabled) Debugging.Assert(numTerms > 0);
@@ -661,7 +661,7 @@ namespace Lucene.Net.Codecs
             // NOTE: cannot seek!
             private sealed class IntersectEnum : TermsEnum
             {
-                private readonly BlockTreeTermsReader<T>.FieldReader outerInstance;
+                private readonly BlockTreeTermsReader<TSubclassState>.FieldReader outerInstance;
 
                 private readonly IndexInput @in;
 
@@ -681,7 +681,7 @@ namespace Lucene.Net.Codecs
                 // TODO: can we share this with the frame in STE?
                 private sealed class Frame
                 {
-                    private readonly BlockTreeTermsReader<T>.FieldReader.IntersectEnum outerInstance;
+                    private readonly BlockTreeTermsReader<TSubclassState>.FieldReader.IntersectEnum outerInstance;
 
                     internal readonly int ord;
                     internal long fp;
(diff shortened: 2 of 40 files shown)


src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
```diff
@@ -114,10 +114,14 @@ public class BlockTreeTermsReader : FieldsProducer
 
         private readonly int version;
 
+        protected readonly object subclassState;
+
         /// <summary>
         /// Sole constructor. </summary>
-        public BlockTreeTermsReader(Directory dir, FieldInfos fieldInfos, SegmentInfo info, PostingsReaderBase postingsReader, IOContext ioContext, string segmentSuffix, int indexDivisor)
+        public BlockTreeTermsReader(Directory dir, FieldInfos fieldInfos, SegmentInfo info, PostingsReaderBase postingsReader, IOContext ioContext, string segmentSuffix, int indexDivisor, object subclassState = default)
```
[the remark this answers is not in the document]
NightOwl888: Also, it looks like we need to add this parameter to the XML docs for `Lucene40PostingsFormat` and `Lucene41PostingsFormat` to avoid lots of warnings.

NightOwl888 (changes requested):

src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
```diff
@@ -86,7 +86,7 @@ namespace Lucene.Net.Codecs
     /// <para/>
     /// @lucene.experimental
     /// </summary>
-    public class BlockTreeTermsReader : FieldsProducer
+    public class BlockTreeTermsReader<T> : FieldsProducer
```
NightOwl888, line 89: Let's make this more descriptive: `TSubclassState`.

NightOwl888: Let's also add some docs to explain that this `TSubclassState` argument is so a subclass can set state. Specifically, it is for optional use when overriding the `ReadHeader()`, `ReadIndexHeader()` and `SeekDir()` methods. This only matters in the case where the state is required inside of any of those methods that is passed in to the subclass constructor.

`new MyBlockTreeTermsReader(bool decideSomething)`

Which can then be passed down in the constructor to `BlockTreeTermsReader<bool>`. The docs should be clear that this gets set to the protected field `m_subclassState` before any of the above methods are called where it is available for reading when overriding the above methods.

If the user needs to pass more than one piece of data, they can create a class or struct to do so. All other virtual members of `BlockTreeTermsReader<TSubclassState>` are not called in the constructor, so the overrides of those methods won't specifically need to use this field (although they could for consistency).

NightOwl888: Looks like we have some XML doc comments that also need to be updated with the generic type to prevent warnings from happening. I am leaving this comment here, because it won't let me comment on the line where it is detected.

NightOwl888 (changes requested):
Looks pretty good. Just a few updates to finish up the `BlockTreeTermsReader`.

src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
laimis: Perfect, thank you for taking a look! I am adding those changes now.

src/Lucene.Net.Codecs/Pulsing/Pulsing41PostingsFormat.cs
```diff
@@ -35,7 +35,7 @@ public Pulsing41PostingsFormat()
 
         /// <summary>Inlines docFreq=<paramref name="freqCutoff"/> terms, otherwise uses the normal "Lucene41" format.</summary>
         public Pulsing41PostingsFormat(int freqCutoff) 
-            : this(freqCutoff, BlockTreeTermsWriter.DEFAULT_MIN_BLOCK_SIZE, BlockTreeTermsWriter.DEFAULT_MAX_BLOCK_SIZE)
+            : this(freqCutoff, BlockTreeTermsWriter<object>.DEFAULT_MIN_BLOCK_SIZE, BlockTreeTermsWriter<object>.DEFAULT_MAX_BLOCK_SIZE)
```
NightOwl888, line 38: Usually when I run into ambiguity like this due to generics, I move the constant into a non-generic static class with the same name so it can still be accessed via `BlockTreeTermsWriter.DEFAULT_MIN_BLOCK_SIZE`. It isn't as nice on the inside of the class, but everybody else sees it in the same place. IMO, it makes more sense than having an `<object>` argument that is completely meaningless, but it isn't 100% clear that it is meaningless when specifying these constants in the code.

```c#
public static class BlockTreeTermsWriter
{
        /// <summary>
        /// Suggested default value for the 
        /// <c>minItemsInBlock</c> parameter to 
        /// <see cref="BlockTreeTermsWriter(SegmentWriteState, PostingsWriterBase, int, int)"/>.
        /// </summary>
        public const int DEFAULT_MIN_BLOCK_SIZE = 25;

        /// <summary>
        /// Suggested default value for the 
        /// <c>maxItemsInBlock</c> parameter to 
        /// <see cref="BlockTreeTermsWriter(SegmentWriteState, PostingsWriterBase, int, int)"/>.
        /// </summary>
        public const int DEFAULT_MAX_BLOCK_SIZE = 48;
}
```

There may be other constants that would be useful to move to this static class to reduce the number of these ambiguous calls.

We should also move the 2 public nested classes `BlockTreeTermsReader.FieldReader` and `BlockTreeTermsReader.Stats` so they can be found in the same place on the public API. Be sure to drop a comment in the place where they were moved from so we understand when looking at the Java code file where they have gone.

```c#
// LUCENENET specific - moved Stats class inside static BlockTreeTermsReader class so it can be used without a generic closing type.
```

src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
```diff
@@ -82,11 +82,11 @@ namespace Lucene.Net.Codecs
     /// option to see summary statistics on the blocks in the
     /// dictionary.</para>
     ///
-    /// See <see cref="BlockTreeTermsWriter"/>.
+    /// See <see cref="BlockTreeTermsWriter{T}"/>.
```
NightOwl888, line 85: Please update this XML doc comment to match `TSubclassState`

src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
```diff
@@ -114,15 +114,36 @@ public class BlockTreeTermsReader : FieldsProducer
...
+        /// 
+        /// When passed to the constructor, it is set to the protected field m_subclassState before
+        /// any of the above methods are called where it is available for reading when overriding the above methods.
+        /// 
+        /// If your subclass needs to pass more than one piece of data, you can create a class or struct to do so.
+        /// All other virtual members of BlockTreeTermsReader are not called in the constructor, 
+        /// so the overrides of those methods won't specifically need to use this field (although they could for consistency).
+        /// </param>
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Internal class")]
```
NightOwl888, line 134: Let's change the justification here to "Required for continuity with Lucene's design".

src/Lucene.Net/Codecs/BlockTreeTermsWriter.cs
```diff
@@ -178,20 +179,20 @@ namespace Lucene.Net.Codecs
     /// <para/>
     /// @lucene.experimental
     /// </summary>
-    /// <seealso cref="BlockTreeTermsReader"/>
-    public class BlockTreeTermsWriter : FieldsConsumer
+    /// <seealso cref="BlockTreeTermsReader{T}"/>
```
NightOwl888, line 182: Please update this XML doc comment to match `TSubclassState`

src/Lucene.Net/Codecs/BlockTreeTermsWriter.cs
```diff
@@ -178,20 +179,20 @@ namespace Lucene.Net.Codecs
...
+    public class BlockTreeTermsWriter<TSubclassState> : FieldsConsumer
     {
         /// <summary>
         /// Suggested default value for the 
         /// <c>minItemsInBlock</c> parameter to 
-        /// <see cref="BlockTreeTermsWriter(SegmentWriteState, PostingsWriterBase, int, int)"/>.
+        /// <see cref="BlockTreeTermsWriter{T}(SegmentWriteState, PostingsWriterBase, int, int, T)"/>.
```
NightOwl888, line 188: Please update this XML doc comment to match `TSubclassState` in both places

src/Lucene.Net/Codecs/BlockTreeTermsWriter.cs
```diff
@@ -178,20 +179,20 @@ namespace Lucene.Net.Codecs
...
         /// </summary>
         public const int DEFAULT_MIN_BLOCK_SIZE = 25;
 
         /// <summary>
         /// Suggested default value for the 
         /// <c>maxItemsInBlock</c> parameter to 
-        /// <see cref="BlockTreeTermsWriter(SegmentWriteState, PostingsWriterBase, int, int)"/>.
+        /// <see cref="BlockTreeTermsWriter{T}(SegmentWriteState, PostingsWriterBase, int, int, T)"/>.
```
NightOwl888, line 195: Please update this XML doc comment to match `TSubclassState` in both places

src/Lucene.Net/Codecs/BlockTreeTermsWriter.cs
```diff
@@ -279,14 +280,33 @@ public FieldMetaData(FieldInfo fieldInfo, BytesRef rootCode, long numTerms, long
...
+        /// 
+        /// When passed to the constructor, it is set to the protected field m_subclassState before
+        /// any of the above methods are called where it is available for reading when overriding the above methods.
+        /// 
+        /// If your subclass needs to pass more than one piece of data, you can create a class or struct to do so.
+        /// All other virtual members of BlockTreeTermsWriter are not called in the constructor, 
+        /// so the overrides of those methods won't specifically need to use this field (although they could for consistency).
+        /// </param>
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Internal class")]
```
NightOwl888, line 302: Let's change the justification here to "Required for continuity with Lucene's design".

src/Lucene.Net/Codecs/Lucene40/Lucene40PostingsFormat.cs
```diff
@@ -42,7 +42,7 @@ namespace Lucene.Net.Codecs.Lucene40
     /// field along with per-term statistics (such as docfreq)
     /// and pointers to the frequencies, positions and
     /// skip data in the .frq and .prx files.
-    /// See <see cref="BlockTreeTermsWriter"/> for more details on the format.
+    /// See <see cref="BlockTreeTermsWriter{T}"/> for more details on the format.
```
NightOwl888, line 45: Please update this XML doc comment to match `TSubclassState`

src/Lucene.Net/Codecs/Lucene40/Lucene40PostingsFormat.cs
```diff
@@ -87,7 +87,7 @@ namespace Lucene.Net.Codecs.Lucene40
     /// <a name="Termindex" id="Termindex"></a>
     /// <h3>Term Index</h3>
     /// <para>The .tip file contains an index into the term dictionary, so that it can be
-    /// accessed randomly.  See <see cref="BlockTreeTermsWriter"/> for more details on the format.</para>
+    /// accessed randomly.  See <see cref="BlockTreeTermsWriter{T}"/> for more details on the format.</para>
```
NightOwl888, line 90: Please update this XML doc comment to match `TSubclassState`

src/Lucene.Net/Codecs/Lucene41/Lucene41PostingsFormat.cs
```diff
@@ -379,15 +379,15 @@ public sealed class Lucene41PostingsFormat : PostingsFormat
...
+            : this(BlockTreeTermsWriter<object>.DEFAULT_MIN_BLOCK_SIZE, BlockTreeTermsWriter<object>.DEFAULT_MAX_BLOCK_SIZE)
         {
         }
 
         /// <summary>
         /// Creates <see cref="Lucene41PostingsFormat"/> with custom
         /// values for <paramref name="minTermBlockSize"/> and 
         /// <paramref name="maxTermBlockSize"/> passed to block terms dictionary. </summary>
-        /// <seealso cref="BlockTreeTermsWriter.BlockTreeTermsWriter(SegmentWriteState,PostingsWriterBase,int,int)"/>
+        /// <seealso cref="BlockTreeTermsWriter{T}.BlockTreeTermsWriter(SegmentWriteState,PostingsWriterBase,int,int,T)"/>
```
NightOwl888, line 390: Please update this XML doc comment to match `TSubclassState` in both places

src/Lucene.Net/Codecs/PostingsReaderBase.cs
```diff
@@ -29,7 +29,7 @@ namespace Lucene.Net.Codecs
 
     /// <summary>
     /// The core terms dictionaries (BlockTermsReader,
-    /// <see cref="BlockTreeTermsReader"/>) interact with a single instance
+    /// <see cref="BlockTreeTermsReader{T}"/>) interact with a single instance
```
NightOwl888, line 32: Please update this XML doc comment to match `TSubclassState`

src/Lucene.Net/Codecs/Lucene41/Lucene41PostingsFormat.cs
```diff
@@ -161,7 +161,7 @@ namespace Lucene.Net.Codecs.Lucene41
     /// <dd>
     /// <b>Term Index</b>
     /// <para>The .tip file contains an index into the term dictionary, so that it can be
-    /// accessed randomly.  See <see cref="BlockTreeTermsWriter"/> for more details on the format.</para>
+    /// accessed randomly.  See <see cref="BlockTreeTermsWriter{T}"/> for more details on the format.</para>
```
NightOwl888, line 164: Please update this XML doc comment to match `TSubclassState`

src/Lucene.Net/Codecs/Lucene40/Lucene40PostingsFormat.cs
```diff
@@ -221,15 +221,15 @@ public class Lucene40PostingsFormat : PostingsFormat
...
+            : this(BlockTreeTermsWriter<object>.DEFAULT_MIN_BLOCK_SIZE, BlockTreeTermsWriter<object>.DEFAULT_MAX_BLOCK_SIZE)
         {
         }
 
         /// <summary>
         /// Creates <see cref="Lucene40PostingsFormat"/> with custom
         /// values for <paramref name="minBlockSize"/> and 
         /// <paramref name="maxBlockSize"/> passed to block terms dictionary. </summary>
-        ///  <seealso cref="BlockTreeTermsWriter.BlockTreeTermsWriter(SegmentWriteState,PostingsWriterBase,int,int)"/>
+        ///  <seealso cref="BlockTreeTermsWriter{T}.BlockTreeTermsWriter(SegmentWriteState,PostingsWriterBase,int,int,T)"/>
```
NightOwl888, line 232: Please update this XML doc comment to match `TSubclassState` in both places

NightOwl888 (changes requested):
I apologize for the repetitive comments. If there is a way to specify "apply this comment to these lines..." feature in GitHub, please let me know where it is.

src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
```diff
@@ -114,10 +114,31 @@ public class BlockTreeTermsReader : FieldsProducer
...
+        /// 
+        /// When passed to the constructor, it is set to the protected field m_subclassState before
+        /// any of the above methods are called where it is available for reading when overriding the above methods.
+        /// 
+        /// If your subclass needs to pass more than one piece of data, you can create a class or struct to do so.
+        /// All other virtual members of BlockTreeTermsReader are not called in the constructor, 
+        /// so the overrides of those methods won't specifically need to use this field (although they could for consistency).
+        /// </param>
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Required for continuity with Lucene's design", Justification = "Internal class")]
```
NightOwl888, line 134: Oops, I meant to put that into the Justification property.

```c#
[SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]
```

src/Lucene.Net/Codecs/BlockTreeTermsWriter.cs
```diff
@@ -279,14 +286,33 @@ public FieldMetaData(FieldInfo fieldInfo, BytesRef rootCode, long numTerms, long
...
+        /// 
+        /// When passed to the constructor, it is set to the protected field m_subclassState before
+        /// any of the above methods are called where it is available for reading when overriding the above methods.
+        /// 
+        /// If your subclass needs to pass more than one piece of data, you can create a class or struct to do so.
+        /// All other virtual members of BlockTreeTermsWriter are not called in the constructor, 
+        /// so the overrides of those methods won't specifically need to use this field (although they could for consistency).
+        /// </param>
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Required for continuity with Lucene's design", Justification = "Internal class")]
```
NightOwl888, line 308: Oops, I meant to put that into the Justification property.

```c#
[SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]
```

NightOwl888 (changes requested):
Looks good.

Just a minor change in 2 different places and we are good to merge.

src/Lucene.Net/Codecs/BlockTreeTermsReader.cs
laimis: Done!

PR #820, opened 2023-04-11
BREAKING: Lucene.Net.Util.PriorityQueue<T>: Replaced (int, bool) constructor and removed constructor call to virtual GetSentinelObject() method

NightOwl888:
This breaks API compatibility in 2 ways:

1. The `PriorityQueue<T>(int, bool)` constructor was replaced with `PriorityQueue<T>(int, ISentinelFactory<T>)`.
2. The virtual `GetSentinelObject()` method was removed from `PriorityQueue<T>`.

This also removes ambiguity since both "prepopulate" had to be specified as "true" in the constructor **AND** the `GetSentinelObject()` method had to be overridden in a subclass for it to work. Now there is a single signal. Pass `null` to **not** populate. Pass a `ISentinelFactory<T>` implementation to populate.

This PR also contains some related `null` guard clauses and nullable reference type updates to the API (so it is clear when a `null` is accepted for a given parameter.

### DIFF
--- a/src/Lucene.Net.Benchmark/ByTask/Tasks/ReadTask.cs
+++ b/src/Lucene.Net.Benchmark/ByTask/Tasks/ReadTask.cs
@@ -205,7 +205,7 @@ namespace Lucene.Net.Benchmarks.ByTask.Tasks
 
         protected virtual ICollector CreateCollector()
         {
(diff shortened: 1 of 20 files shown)


laimis (approved):
This turned out nicely!

PR #821, opened 2023-04-12
add ignore attributes

laimis:

This one focuses on the findings in the Benchmark project. Adding attributes to ignore the warnings as opposed to introducing a workaround or a breaking change.

### DIFF
--- a/src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
+++ b/src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
@@ -99,9 +99,15 @@ namespace Lucene.Net.Benchmarks.ByTask
         private readonly IDictionary<string, object> perfObjects = new Dictionary<string, object>();
 
         // constructor
+        public PerfRunData(Config config) : this(config, true)
+        {    
+        }
+
+        // LUCENENET specific - added performReinit parameter to allow subclasses to skip reinit
+        // since it's a virtual method. Subclass can call that method itself if needed.
         [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
         [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]
-        public PerfRunData(Config config)
+        protected PerfRunData(Config config, bool performReinit)
         {
             this.config = config;
             // analyzer (default is standard analyzer)
@@ -125,7 +131,10 @@ namespace Lucene.Net.Benchmarks.ByTask
             qmkrClass = Type.GetType(config.Get("query.maker", typeof(SimpleQueryMaker).AssemblyQualifiedName));
 
             // index stuff
-            Reinit(false);
+            if (performReinit)
+            {
+                Reinit(false);
+            }
 
             // statistic points
             points = new Points(config);
--- a/src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
+++ b/src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
@@ -11,6 +11,7 @@ using Lucene.Net.Support.Threading;
 using Lucene.Net.Util;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
 using Console = Lucene.Net.Util.SystemConsole;
@@ -98,6 +99,8 @@ namespace Lucene.Net.Benchmarks.ByTask
         private readonly IDictionary<string, object> perfObjects = new Dictionary<string, object>();
 
         // constructor
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]
         public PerfRunData(Config config)
         {
             this.config = config;
--- a/src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
+++ b/src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
@@ -113,7 +113,8 @@ namespace Lucene.Net.Benchmarks.ByTask
         [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]
         protected PerfRunData(Config config, bool performReinit, bool logQueries)
         {
-            this.config = config;
+            this.config = config ?? throw new ArgumentNullException(nameof(config));
+            
             // analyzer (default is standard analyzer)
             analyzer = NewAnalyzerTask.CreateAnalyzer(config.Get("analyzer",
                 typeof(Lucene.Net.Analysis.Standard.StandardAnalyzer).AssemblyQualifiedName));
--- a/src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
+++ b/src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
@@ -99,15 +99,19 @@ namespace Lucene.Net.Benchmarks.ByTask
         private readonly IDictionary<string, object> perfObjects = new Dictionary<string, object>();
 
         // constructor
-        public PerfRunData(Config config) : this(config, true)
+        public PerfRunData(Config config) : this(
+            config,
+            performReinit: true,
+            logQueries: string.Equals(config?.Get("log.queries", "false") ?? "false", "true", StringComparison.OrdinalIgnoreCase))
         {    
         }
 
         // LUCENENET specific - added performReinit parameter to allow subclasses to skip reinit
-        // since it's a virtual method. Subclass can call that method itself if needed.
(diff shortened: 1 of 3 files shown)


src/Lucene.Net.Benchmark/ByTask/Tasks/WriteLineDocTask.cs
```diff
@@ -92,7 +93,8 @@ public class WriteLineDocTask : PerfTask
 
         private readonly object lineFileLock = new object(); // LUCENENET specific - lock to ensure writes don't collide for this instance
 
-
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]
         public WriteLineDocTask(PerfRunData runData)
```
NightOwl888, line 98: This one can be fixed by changing the `lineFileOut` field:

```c#
protected readonly TextWriter m_lineFileOut;
```

And adding a constructor so the user can opt out of the call:

```c#
protected WriteLineDocTask(PerfRunData runData, bool performWriteHeader)
    : base(runData)
{
    Config config = runData.Config;
    m_fname = config.Get("line.file.out", null);
    if (m_fname is null)
    {
        throw new ArgumentException("line.file.out must be set");
    }
    Stream @out = StreamUtils.GetOutputStream(new FileInfo(m_fname));
    m_lineFileOut = new StreamWriter(@out, Encoding.UTF8);
    docMaker = runData.DocMaker;

    // init fields 
    string f2r = config.Get("line.fields", null);
    if (f2r is null)
    {
        fieldsToWrite = DEFAULT_FIELDS;
    }
    else
    {
        if (f2r.IndexOf(SEP) >= 0)
        {
            throw new ArgumentException("line.fields " + f2r + " should not contain the separator char: " + SEP);
        }
        fieldsToWrite = f2r.Split(',').TrimEnd();
    }

    // init sufficient fields
    sufficientFields = new bool[fieldsToWrite.Length];
    string suff = config.Get("sufficient.fields", DEFAULT_SUFFICIENT_FIELDS);
    if (",".Equals(suff, StringComparison.Ordinal))
    {
        checkSufficientFields = false;
    }
    else
    {
        checkSufficientFields = true;
        ISet<string> sf = new JCG.HashSet<string>(suff.Split(',').TrimEnd());
        for (int i = 0; i < fieldsToWrite.Length; i++)
        {
            if (sf.Contains(fieldsToWrite[i]))
            {
                sufficientFields[i] = true;
            }
        }
    }

    if (performWriteHeader)
    {
        WriteHeader(m_lineFileOut);
    }
}

public WriteLineDocTask(PerfRunData runData)
    : this(runData, performWriteHeader: true)
{
}
```

It can then be used like this:

```c#
public WriteEnwikiLineDocTask(PerfRunData runData)
    : base(runData, performWriteHeader: false)
{
    // Do my stuff here

    WriteHeader(m_lineFileOut);
   

    Stream @out = StreamUtils.GetOutputStream(CategoriesLineFile(new FileInfo(m_fname)));
    categoryLineFileOut = new StreamWriter(@out, Encoding.UTF8);
    WriteHeader(categoryLineFileOut);
}
```

The trick is to move the existing code to the new protected constructor with an opt out of the virtual call.

NightOwl888 (changes requested):
So, playing around with the extra constructor idea, it appears that since these have only 1 violation each and there is a logical constructor signature that can be added, these are pretty straightforward to fix. We still need the suppression, but at least there is a way out for inheritors if it ever comes up.

src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
```diff
@@ -98,6 +99,8 @@ public class PerfRunData : IDisposable
         private readonly IDictionary<string, object> perfObjects = new Dictionary<string, object>();
 
         // constructor
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]
         public PerfRunData(Config config)
```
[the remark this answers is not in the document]
laimis: Sounds good, will do that!

src/Lucene.Net.Benchmark/ByTask/Tasks/ReadTask.cs
```diff
@@ -48,8 +49,10 @@ public abstract class ReadTask : PerfTask
     {
         private readonly IQueryMaker queryMaker;
 
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]    
         protected ReadTask(PerfRunData runData) // LUCENENET: CA1012: Abstract types should not have constructors (marked protected)
-            : base(runData)
+            : this(runData, null)
```
NightOwl888, line 55: Let's add a label here to improve readability `: this(runData, queryMaker: null)`

src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
```diff
@@ -98,7 +99,15 @@ public class PerfRunData : IDisposable
         private readonly IDictionary<string, object> perfObjects = new Dictionary<string, object>();
 
         // constructor
-        public PerfRunData(Config config)
+        public PerfRunData(Config config) : this(config, true)
```
NightOwl888, line 102: Let's add a label to improve readability in the case of `true`

```c#
: this(config, performReinit: true,
    string.Equals("log. Queries", "false") ?? "false", "true", StringComparison.OrdinalIgnoreCase))
```

NightOwl888 (changes requested):

src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
```diff
@@ -98,7 +99,15 @@ public class PerfRunData : IDisposable
...
+        {    
+        }
+
+        // LUCENENET specific - added performReinit parameter to allow subclasses to skip reinit
+        // since it's a virtual method. Subclass can call that method itself if needed.
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]
+        protected PerfRunData(Config config, bool performReinit)
```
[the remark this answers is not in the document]
laimis: I will leave the ? nullable check out, looking at the original code, I don't think it expects config to be null:

-            if (bool.Parse(config.Get("log.queries", "false")))

src/Lucene.Net.Benchmark/ByTask/Tasks/ReadTask.cs
```diff
@@ -61,6 +64,14 @@ protected ReadTask(PerfRunData runData) // LUCENENET: CA1012: Abstract types sho
             }
         }
 
+        // LUCENENET specific - added this constructor to allow subclasses to initialize it
+        // without having to call constructor that makes a virtual method call
+        protected ReadTask(PerfRunData runData, IQueryMaker queryMaker)
```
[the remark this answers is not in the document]
laimis: sounds good

laimis:
@NightOwl888 ready for another pass

src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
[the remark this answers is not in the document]
NightOwl888: Actually, the `config` value will get a `NullReferenceException` anyway if it is `null`. Let's add a guard glause for that (and update the docs that it will throw `ArgumentNullException` in this case). This was to allow that `null` check to be performed later inside of the constructor, but I forgot to mention it previously.

But if you feel this is too much logic to put inside of a constructor cascade, we could make it into a private static method instead.

src/Lucene.Net.Benchmark/ByTask/PerfRunData.cs
```diff
@@ -98,7 +99,19 @@ public class PerfRunData : IDisposable
...
+        }
+
+        // LUCENENET specific - added performReinit parameter to allow subclasses to skip reinit
+        // since it's a virtual method. Also added logQueries parameter to allow to skip calling
+        // virtual GetQueryMarker. Subclass can call these methods itself if needed.
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]
+        protected PerfRunData(Config config, bool performReinit, bool logQueries)
         {
             this.config = config;
```
NightOwl888, line 116: Let's add a `null` coalesce guard clause here.

```c#
this.config = config ?? throw new ArgumentNullException(nameof(config));
```

laimis: Done!

NightOwl888 (approved):

PR #822, opened 2023-04-12
Ignore sonar cloud errors for DocTermOrds

laimis:

We are ignoring DocTermOrds errors as the Lucene team intended for the subclasses to use the protected constructor that does not call the virtual method, and they even specify that subclasses should call the virtual method themselves.


NightOwl888 (approved):
No brainer! Good work finding such a simple solution.

PR #823, opened 2023-04-12
Supress sonar cloud virtual method from constructor call error

laimis:

This one focuses on OpenBitSetDISI class. It gets deleted in later version of Lucene, so figured it's not worth addressing the issue here, especially since the fix does not appear to be trivial (e.g. we can't resort to private easily as it calls another virtual method in the logic, or introduce a different constructor).

### DIFF
--- a/src/Lucene.Net/Util/OpenBitSetDISI.cs
+++ b/src/Lucene.Net/Util/OpenBitSetDISI.cs
@@ -1,21 +1,22 @@
 namespace Lucene.Net.Util
 {
+    using System.Diagnostics.CodeAnalysis;
     /*
-     * Licensed to the Apache Software Foundation (ASF) under one or more
-     * contributor license agreements.  See the NOTICE file distributed with
-     * this work for additional information regarding copyright ownership.
-     * The ASF licenses this file to You under the Apache License, Version 2.0
-     * (the "License"); you may not use this file except in compliance with
-     * the License.  You may obtain a copy of the License at
-     *
-     *     http://www.apache.org/licenses/LICENSE-2.0
-     *
-     * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


src/Lucene.Net/Util/OpenBitSetDISI.cs
```diff
@@ -31,6 +32,8 @@ public class OpenBitSetDISI : OpenBitSet
         /// Also give a maximum size one larger than the largest doc id for which a
         /// bit may ever be set on this <see cref="OpenBitSetDISI"/>.
         /// </summary>
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "This class will get removed in later versions of Lucene, see LUCENE-6010")]
         public OpenBitSetDISI(DocIdSetIterator disi, int maxSize)
```
NightOwl888, line 37: Since we have a single call to a public virtual method, we could use the `InPlaceOrInternal` trick to get around this one.

But after searching the codebase, I see this is not referenced anywhere. Either one works, I guess.

laimis: InPlaceOr that is called makes a call to another virtual FastSet. I believe that would be harder to fix unless we made it have it call a private version too. Seems like not worth the effort?

NightOwl888: It is fine if we leave it as is. I was even thinking that maybe we should remove the class since it isn't being referenced, but since it was in 4.8.0 I guess we can keep it in case anyone started using it in the beta.

PR #824, opened 2023-04-12
add ignore attributes

laimis:

JaspellTernarySearchTrie gets removed in the later versions of Lucene: https://github.com/apache/lucene/commit/4213f9d3cd1742082ebac74fb6510dc8c46bfd95. It is not worth it for us to find a way to extend this class without virtual method calls being a potential issue.

### DIFF
--- a/src/Lucene.Net.Suggest/Suggest/Jaspell/JaspellTernarySearchTrie.cs
+++ b/src/Lucene.Net.Suggest/Suggest/Jaspell/JaspellTernarySearchTrie.cs
@@ -30,6 +30,7 @@ using J2N.Text;
 using Lucene.Net.Util;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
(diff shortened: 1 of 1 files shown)


NightOwl888 (changes requested):
While we will probably be stuck with this for awhile before we can deprecate and/or remove it, I am not seeing a straightforward way to fix it, being that the call is in a loop.

But then, there is a `public JaspellTernarySearchTrie(CultureInfo culture)` method that could be overridden and the entire contents of `public JaspellTernarySearchTrie(FileInfo file, bool compression, CultureInfo culture)` placed there with whatever changes the subclass needs.

Since we took the time to figure that out, we should add some notes to at least the `public JaspellTernarySearchTrie(FileInfo file, bool compression, CultureInfo culture)` overload explaining to override the `: base(CultureInfo)` overload if they need to override virtual `GetOrCreateNode()` to customize it.

NightOwl888 (approved):

Issue #307, opened 2020-07-07
Documentation/Analysis for "System Properties"


NightOwl888:
We completed the "System Properties" feature in #254/#287, but we are lacking documentation on how end users can utilize it, in particular for the test framework. Here are a list of the properties that exist and their defaults, however, not all of them apply to the .NET port.

```
{
    "smartcn:data:dir": "",
    "egothor:stemmer:charset": "UTF-8",
    "kuromoji:data:dir": "",
    "solr:test:leavetmpdir": null, // Doesn't apply to Lucene.NET
    "assert": false, // Turn on assertions (for testing/logging) in release build (.NET-only)
    "tests": {
        "asserts:gracious": "false", // Doesn't apply to Lucene.NET
        "cleanthreads:sysprop": "perMethod", // Doesn't apply to Lucene.NET
        "codec": "random",  // Specifies a specific Codec subclass to use for the tests
        "directory": "random", // Specifies a specific Directory subclass to use for the tests
        "docvaluesformat": "random",  // Specifies a specific DocValuesFormat subclass to use for the tests
        "infostream": "${verbose}", // Boolean that specifies whether to log output to console
        "jvms": "auto", // Doesn't apply to Lucene.NET
        "jvms:override": "${jvms}", // Doesn't apply to Lucene.NET
        "leaveTemporary": null, // Skips deletion of files that were created during tests
        "leavetemporary": null, // Alias of above
        "leavetmpdir": null, // Alias of above
        "linedocsfile": "", // Specifies a specific LineDocsFile to use when running tests (may not be functional with external files)
        "locale": "random", // Specifies a specific culture to use when running tests (TODO: rename or alias "culture")
        "multiplier": "1", // Multiplier to scale random tests for more thorough (but longer) testing - used during nightly/weakly builds
        "nightly": "false", // TODO: Need to set up nightly build infrastructure
        "postingsformat": "random",  // Specifies a specific PostingsFormat subclass to use for the tests
        "seed": "", // Specifies a random seed to use during tests in hexadecimal format (not yet implemented - see #288)
        "slow": "true", // Specifies to run slow tests (not implemented, but we made a LongRunningTestAttribute that could be adapted)
        "timezone": "random",  // Specifies a specific time zone to use for the tests (not xplat because of different naming conventions)
        "verbose": "false", // Use verbose logging
        "weekly": "false" // TODO: Need to set up weekly build infrastructure
    },
    "user:timezone": "random" // Doesn't apply to Lucene.NET
[...]

PR #826, opened 2023-04-12
BREAKING: Introducing ValuePriorityQueue<T>

NightOwl888:
This adds a new ref struct to the project, `Lucene.Net.Util.ValuePriorityQueue<T>`. This struct contains the most of the same API and business logic as `PriorityQueue<T>`, but allows the memory used to be allocated elsewhere as `Span<T>` and passed in.

The amount to allocate can be calculated by calling `PriorityQueue.GetArrayHeapSize(maxSize)`, where `maxSize` is the external size of the `PriorityQueue` (representing the same value that is passed into the constructor of the current `PriorityQueue<T>`). The actual allocation amount is 1 greater than the priority queue size because it is 1-based rather than 0-based.

We get around the issue with structs not being inheritable by introducing a new `PriorityComparer<T>` abstract class with a single `bool LessThan(T a, T b)` method. This signature makes converting existing subclasses of `PriorityQueue<T>` to `PriorityComparer<T>` straightforward, as there are no business logic changes to make. This `IComparer<T>` is a required constructor argument of `ValuePriorityQueue<T>`. It can be either a `PriorityComparer<T>` or `IComparer<T>`, although if it returns a 0 (equal), that value is considered the same as 1 (greater than) by the priority queue.

This also modifies `SingleTaxonomyFacets`, `Int32TaxonomyFacets`, `TopDocs` merge and `FST` pack to use `ValuePriorityQueue<T>` and converts the item types they deal with from class to struct so they can be stack allocated if the size is below `Constants.MaxStackByteLimit`, which is set using the [system property](https://github.com/apache/lucenenet/issues/307) `maxStackByteLimit`.

Closes #774. Thanks @eladmarg for the suggestion.

There are a few more places in the codebase where this type can be useful, but unfortunately the types they use don't convert easily into structs so although we can eliminate the heap allocation of the priority queue itself, we will need some configuration and/or logic to decide whether to allocate the array directly or use an array pool.

## Breaking Changes

1. Changed `OrdAndValue<T>` from class to struct.
2. Added generic constraint to `OrdAndValue<T> where T : struct`.

### DIFF
--- a/src/Lucene.Net.Facet/Taxonomy/FloatTaxonomyFacets.cs
+++ b/src/Lucene.Net.Facet/Taxonomy/FloatTaxonomyFacets.cs
@@ -1,6 +1,8 @@
 ﻿// Lucene version compatibility level 4.8.1
 using Lucene.Net.Diagnostics;
+using Lucene.Net.Util;
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Diagnostics;
 
@@ -33,6 +35,8 @@ namespace Lucene.Net.Facet.Taxonomy
     /// </summary>
     public abstract class SingleTaxonomyFacets : TaxonomyFacets
     {
+        private const int OrdAndSingleByteSize = sizeof(int) + sizeof(float); // LUCENENET specific - so we can calculate stack size
+
         /// <summary>
         /// Per-ordinal value. </summary>
         protected readonly float[] m_values;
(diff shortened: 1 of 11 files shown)


laimis (approved):
This looks like a great change! Nicely done!

eladmarg:
fantastic! 
sometimes the priority queue size is so small, no need even to rent and allocate on the stack is much faster.

by the way, net 7 finally introduce [PriorityQueue](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.priorityqueue-2?view=net-7.0) implementation, however this one seems to be more specific and faster.

PR #828, opened 2023-04-13
Workaround for virtual call issue for WithinPrefixTreeFilter

laimis:

This one focuses on WithinPrefixTreeFilter. The fix was to add a protected constructor that does not call virtual methods, and the subclasses can use it to initialize and then call any methods they need inside their constructors.

### DIFF
--- a/src/Lucene.Net.Spatial/Prefix/WithinPrefixTreeFilter.cs
+++ b/src/Lucene.Net.Spatial/Prefix/WithinPrefixTreeFilter.cs
@@ -8,6 +8,7 @@ using Spatial4n.Distance;
 using Spatial4n.Shapes;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 
 namespace Lucene.Net.Spatial.Prefix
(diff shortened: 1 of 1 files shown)


src/Lucene.Net.Spatial/Prefix/WithinPrefixTreeFilter.cs
```diff
@@ -50,28 +51,40 @@ public class WithinPrefixTreeFilter : AbstractVisitingPrefixTreeFilter
...
             {
-                bufferedQueryShape = BufferShape(queryShape, queryBuffer);
+                m_bufferedQueryShape = BufferShape(queryShape, queryBuffer);
             }
         }
 
+        // LUCENENET specific - subclasses can use this class to override the behavior of
+        // BufferShape method in a way that's safe and called from the subclass constructor
+        protected WithinPrefixTreeFilter(IShape queryShape, string fieldName, SpatialPrefixTree grid, 
+                                      int detailLevel, int prefixGridScanLevel, IShape? bufferShape)
```
NightOwl888, line 82: Let's rename this to `bufferedQueryShape` to match the original field name.

NightOwl888 (changes requested):
Looks good! Just one small change and we are good to merge.

src/Lucene.Net.Spatial/Prefix/WithinPrefixTreeFilter.cs
laimis: done!

PR #829, opened 2023-04-13
Adjust QueryTemplateManager to allow subclasses avoid calling virtual methods from the constructors

laimis:

This one focuses on queryTemplateManager where we allow subclasses to use default constructor and expose m_defaultCompiledTemplates as protected for them to initialize it when needed.

### DIFF
--- a/src/Lucene.Net.QueryParser/Xml/QueryTemplateManager.cs
+++ b/src/Lucene.Net.QueryParser/Xml/QueryTemplateManager.cs
@@ -35,8 +35,7 @@ namespace Lucene.Net.QueryParsers.Xml
     public class QueryTemplateManager
     {
         private readonly IDictionary<string, XslCompiledTransform> compiledTemplatesCache = new Dictionary<string, XslCompiledTransform>(); // LUCENENET: marked readonly
-        // LUCENENET specific - made protected to allow subclasses to access and initialize
-        protected XslCompiledTransform m_defaultCompiledTemplates;
+        private XslCompiledTransform defaultCompiledTemplates;
 
         public QueryTemplateManager()
         {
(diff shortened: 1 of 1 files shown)


src/Lucene.Net.QueryParser/Xml/QueryTemplateManager.cs
```diff
@@ -34,20 +35,28 @@ namespace Lucene.Net.QueryParsers.Xml
     public class QueryTemplateManager
     {
         private readonly IDictionary<string, XslCompiledTransform> compiledTemplatesCache = new Dictionary<string, XslCompiledTransform>(); // LUCENENET: marked readonly
-        private XslCompiledTransform defaultCompiledTemplates;
+        // LUCENENET specific - made protected to allow subclasses to access and initialize
+        protected XslCompiledTransform m_defaultCompiledTemplates;
```
NightOwl888, line 39: It looks like this was meant to be a "default template" and the user could subclass to add their own template. To do this they would need to:

1. Override the `public QueryTemplateManager()` constructor.
2. Load their own template. However they want to.
3. Set their own private `XslCompiledTransform` to a private variable.
4. Override `GetQueryAsXmlString` and `GetQueryAsDOM` to supply their private variable.

Maybe I have missed something, but I don't think there is any advantage to making this variable protected as it is supposed to represent the "default" template, not the user specified one.

NightOwl888 (changes requested):

src/Lucene.Net.QueryParser/Xml/QueryTemplateManager.cs
laimis: oh ok, so basically no need to make this variable protected. Adjusting now...

NightOwl888 (approved):

PR #831, opened 2023-04-13
ComposedQuery fix for virtual call being made from constructor

laimis:

This one focuses on ComposedQuery. We are adding a protected constructor that subclasses can use.

### DIFF
--- a/src/Lucene.Net.QueryParser/Surround/Query/ComposedQuery.cs
+++ b/src/Lucene.Net.QueryParser/Surround/Query/ComposedQuery.cs
@@ -1,6 +1,5 @@
-﻿using Lucene.Net.Diagnostics;
-using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Text;
 using JCG = J2N.Collections.Generic;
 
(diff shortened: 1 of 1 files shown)


src/Lucene.Net.QueryParser/Surround/Query/ComposedQuery.cs
```diff
@@ -28,13 +27,22 @@ namespace Lucene.Net.QueryParsers.Surround.Query
...
 
+        [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
+        [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]
+        protected ComposedQuery(IList<SrndQuery> qs, bool operatorInfix, string opName) // LUCENENET: CA1012: Abstract types should not have constructors (marked protected)
+            : this(operatorInfix, opName)
+        {
+            Recompose(qs);
+        }
+
         protected virtual void Recompose(IList<SrndQuery> queries)
```
laimis, line 46: with the above change, the order of operations changes. We essentially will be first assigning the operatorInfix and opName and then calling Recompose. Where before we first called Recompose and assigned after. Looking at the Recompose logic, it seems a safe thing to do. But figured to highlight that.

NightOwl888 (approved):

PR #837, opened 2023-04-14
IndexSearcher fix for virtual calls from constructor

laimis:

This one focuses on IndexSearcher. Since we discovered a separate constructor approach as an elegant way to address this issue, let's use it for IndexSearcher.

### DIFF
--- a/src/Lucene.Net/Search/IndexSearcher.cs
+++ b/src/Lucene.Net/Search/IndexSearcher.cs
@@ -104,6 +104,7 @@ namespace Lucene.Net.Search
 
         /// <summary>
         /// Creates a searcher searching the provided index. </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="r"/> is <c>null</c>.</exception>
         public IndexSearcher(IndexReader r)
             : this(r, executor: null)
         {
@@ -117,8 +118,9 @@ namespace Lucene.Net.Search
         /// <para/>
         /// @lucene.experimental
         /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="r"/> is <c>null</c>.</exception>
         public IndexSearcher(IndexReader r, TaskScheduler? executor)
-            : this(r.Context, executor)
+            : this(r?.Context!, executor)
         {
         }
 
@@ -132,6 +134,7 @@ namespace Lucene.Net.Search
         /// <para/>
         /// @lucene.experimental 
         /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="context"/> is <c>null</c>.</exception>
         /// <seealso cref="IndexReaderContext"/>
         /// <seealso cref="IndexReader.Context"/>
         public IndexSearcher(IndexReaderContext context, TaskScheduler? executor)
@@ -154,6 +157,7 @@ namespace Lucene.Net.Search
         /// as a member field or property inside a custom override of
         /// <see cref="GetSlices(IList{AtomicReaderContext})"/>.
         /// </remarks>
+        /// <exception cref="ArgumentNullException"><paramref name="context"/> is <c>null</c>.</exception>
         [SuppressMessage("CodeQuality", "IDE0079:Remove unnecessary suppression", Justification = "This is a SonarCloud issue")]
         [SuppressMessage("CodeQuality", "S1699:Constructors should only call non-overridable methods", Justification = "Required for continuity with Lucene's design")]
         protected IndexSearcher(IndexReaderContext context, TaskScheduler? executor, bool allocateLeafSlices)
@@ -180,6 +184,7 @@ namespace Lucene.Net.Search
         /// <para/>
         /// @lucene.experimental
         /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="context"/> is <c>null</c>.</exception>
         /// <seealso cref="IndexReaderContext"/>
         /// <seealso cref="IndexReader.Context"/>
         public IndexSearcher(IndexReaderContext context)
@@ -192,8 +197,13 @@ namespace Lucene.Net.Search
         /// Each <see cref="LeafSlice"/> is executed in a single thread. By default there
         /// will be one <see cref="LeafSlice"/> per leaf (<see cref="AtomicReaderContext"/>).
         /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="leaves"/> is <c>null</c>.</exception>
         protected virtual LeafSlice[] GetSlices(IList<AtomicReaderContext> leaves)
         {
+            // LUCENENET: Added guard clause
+            if (leaves is null)
+                throw new ArgumentNullException(nameof(leaves));
+
             LeafSlice[] slices = new LeafSlice[leaves.Count];
             for (int i = 0; i < slices.Length; i++)
             {
@@ -217,6 +227,7 @@ namespace Lucene.Net.Search
         /// <summary>
         /// Sugar for <code>.IndexReader.Document(docID, fieldVisitor)</code> </summary>
         /// <seealso cref="IndexReader.Document(int, StoredFieldVisitor)"/>
+        /// <exception cref="ArgumentNullException"><paramref name="fieldVisitor"/> is <c>null</c>.</exception>
         public virtual void Doc(int docID, StoredFieldVisitor fieldVisitor)
         {
             reader.Document(docID, fieldVisitor);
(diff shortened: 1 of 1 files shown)


src/Lucene.Net/Search/IndexSearcher.cs
```diff
@@ -135,13 +134,31 @@ public IndexSearcher(IndexReader r, TaskScheduler executor)
...
+            : this(context, executor, allocateLeafSlices: executor is not null)
+        {
+        }
+
+        /// <summary>
+        /// LUCENENET specific constructor that can be used by the subclasses to
+        /// control whether or not the leaf slices are allocated.
+        /// If you choose to skip allocating the leaf slices here, you must 
+        /// initialize <see cref="m_leafSlices" /> in your subclass's constructor, either
+        /// by calling <see cref="GetSlices"/> or using your own logic
```
NightOwl888, line 146: Let's be a bit more specific:

```
/// <summary>
/// LUCENENET specific constructor that can be used by the subclasses to
/// control whether the leaf slices are allocated in the base class or subclass.
/// </summary>
/// <remarks>
/// If executor is non-<c>null</c> and you choose to skip allocating the leaf slices
/// (i.e. <paramref name="allocateLeafSlices"/> == <c>false</c>), you must
/// set the <see cref="m_leafSlices"/> field in your subclass constructor.
/// This is commonly done by calling <see cref="GetSlices(IList{AtomicReaderContext})"/>
/// and using the result to set <see cref="m_leafSlices"/>. You may wish to do this if you
/// have state to pass into your constructor and need to set it prior to the call to
/// <see cref="GetSlices(IList{AtomicReaderContext})"/> so it is available for use
/// as a member field or property inside a custom override of
/// <see cref="GetSlices(IList{AtomicReaderContext})"/>.
/// </remarks>
```

src/Lucene.Net/Search/IndexSearcher.cs
```diff
@@ -464,6 +471,9 @@ protected virtual TopDocs Search(Weight weight, ScoreDoc after, int nDocs)
                 ReentrantLock @lock = new ReentrantLock();
                 ExecutionHelper<TopDocs> runner = new ExecutionHelper<TopDocs>(executor);
 
+                if (m_leafSlices == null)
```
NightOwl888, line 474: Although it doesn't make any difference for arrays, all `null` equality checks have all been converted to use `is` in #617.

The `is` keyword doesn't take any overrides of the `==` operator into account, so it is not possible to circumvent our `null` check logic by passing a class that has one.

src/Lucene.Net/Search/IndexSearcher.cs
```diff
@@ -464,6 +471,9 @@ protected virtual TopDocs Search(Weight weight, ScoreDoc after, int nDocs)
...
+                    throw new InvalidOperationException("m_leafSlices must not be null and initialized in the constructor");
```
NightOwl888, line 475: Let's change this a bit:

```c#
throw new InvalidOperationException($"When the constructor is passed a non-null {nameof(TaskScheduler)}, {nameof(m_leafSlices)} must also be set to a non-null value in the constructor.");
```

src/Lucene.Net/Search/IndexSearcher.cs
```diff
@@ -560,6 +570,9 @@ protected virtual TopFieldDocs Search(Weight weight, FieldDoc after, int nDocs,
                 ReentrantLock @lock = new ReentrantLock();
                 ExecutionHelper<TopFieldDocs> runner = new ExecutionHelper<TopFieldDocs>(executor);
 
+                if (m_leafSlices == null)
+                    throw new InvalidOperationException("m_leafSlices must not be null and initialized in the constructor");
```
NightOwl888, line 574: Let's change this a bit:

```c#
throw new InvalidOperationException($"When the constructor is passed a non-null {nameof(TaskScheduler)}, {nameof(m_leafSlices)} must also be set to a non-null value in the constructor.");
```

NightOwl888 (changes requested):

src/Lucene.Net/Search/IndexSearcher.cs
```diff
@@ -135,13 +134,44 @@ public IndexSearcher(IndexReader r, TaskScheduler executor)
...
         {
-            if (Debugging.AssertsEnabled) Debugging.Assert(context.IsTopLevel,"IndexSearcher's ReaderContext must be topLevel for reader {0}", context.Reader);
+        }
+
+        /// <summary>
+        /// LUCENENET specific constructor that can be used by the subclasses to
+        /// control whether the leaf slices are allocated in the base class or subclass.
+        /// </summary>
+        /// <remarks>
+        /// If executor is non-<c>null</c> and you choose to skip allocating the leaf slices
```
NightOwl888, line 146: Oops. I missed that the second word here should be `<paramref name="executor"/>`.

NightOwl888 (changes requested):

src/Lucene.Net/Search/IndexSearcher.cs
laimis: Updated!

src/Lucene.Net/Search/IndexSearcher.cs
```diff
@@ -397,13 +432,9 @@ public virtual TopFieldDocs Search(Query query, int n, Sort sort)
         ///         <see cref="BooleanQuery.MaxClauseCount"/> clauses. </exception>
         public virtual TopDocs SearchAfter(ScoreDoc after, Query query, int n, Sort sort)
```
NightOwl888, line 433: after can be marked nullable

Also, let's update the docs to indicate `<exception cref="ArgumentNullException"><paramref name="query"> is <c>null</c></exception>`.

Not sure whether `sort` is nullable, that will take some digging.

src/Lucene.Net/Search/IndexSearcher.cs
```diff
@@ -424,13 +455,9 @@ public virtual TopDocs SearchAfter(ScoreDoc after, Query query, int n, Sort sort
         ///         <see cref="BooleanQuery.MaxClauseCount"/> clauses. </exception>
         public virtual TopDocs SearchAfter(ScoreDoc after, Query query, Filter filter, int n, Sort sort, bool doDocScores, bool doMaxScore)
```
NightOwl888, line 456: after can be marked nullable

src/Lucene.Net/Search/IndexSearcher.cs
```diff
@@ -587,7 +620,7 @@ protected virtual TopFieldDocs Search(Weight weight, FieldDoc after, int nDocs,
         /// whether or not the fields in the returned <see cref="FieldDoc"/> instances should
         /// be set by specifying <paramref name="fillFields"/>.
         /// </summary>
-        protected virtual TopFieldDocs Search(IList<AtomicReaderContext> leaves, Weight weight, FieldDoc after, int nDocs, Sort sort, bool fillFields, bool doDocScores, bool doMaxScore)
+        protected virtual TopFieldDocs Search(IList<AtomicReaderContext> leaves, Weight weight, FieldDoc? after, int nDocs, Sort sort, bool fillFields, bool doDocScores, bool doMaxScore)
         {
             // single thread
```
NightOwl888, line 625: We need a null guard clause for weight here, since we access properties of it.

NightOwl888 (changes requested):

NightOwl888 (approved):

src/Lucene.Net/Search/IndexSearcher.cs
```diff
@@ -192,8 +197,13 @@ public IndexSearcher(IndexReaderContext context)
...
         /// will be one <see cref="LeafSlice"/> per leaf (<see cref="AtomicReaderContext"/>).
         /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="leaves"/> is <c>null</c>.</exception>
         protected virtual LeafSlice[] GetSlices(IList<AtomicReaderContext> leaves)
         {
+            // LUCENENET: Added guard clause
+            if (leaves is null)
+                throw new ArgumentNullException(nameof(leaves));
+
             LeafSlice[] slices = new LeafSlice[leaves.Count];
```
laimis, line 206: I am not a fan of these type of additions. This is not present in the original source, not even in the latest version. We could add a lot of extra typing/code (that's not even interesting, just copy paste of exception docs and guar checks). This is why I wasn't sure what you meant by guard clauses as it seemed excessive.

Issue #836, opened 2023-04-14
Task: Investigate slow tests in Lucene.Net.Facet


NightOwl888:
This has been a known issue for quite a while. Some of the tests in Lucene.Net.Facet are running much slower than they do in Lucene 4.8.1.

While the tests are randomized so they vary quite a bit. These are consistently slower than they are in Java. These are some comparisons to demonstrate approximately what the gap is. These are results from Java SE 7 vs .NET 6.0.

| Test | Lucene Duration | Lucene.NET Duration   | 
|---|---|---|
| `TestDirectoryTaxonomyWriter.TestConcurrency()`  | 4.5 s  | 10.7 s  |
|  `TestAddTaxonomy.TestBig()` | 10.9 s  |  16.7 s |
|  `TestTaxonomyCombined.TestTaxonomyRefreshRaces()` | 16.1 s  |  46.2 s |

The other tests seem very reasonable in terms of performance. In the past, we have seen this sort of thing happen due to locking contention and we successfully overcame it by reviewing locks both in Lucene and in the JDK to make sure of several things:

1. That we are not missing a lock.
2. That we are correctly sharing locks between subclass and superclass.
3. That all collection types are being locked appropriately and that the implementation of the collection locks in a similar way in both environments. This is where some pretty big gaps are between Java and .NET.
4. If we still haven't found the issue, revert back to "plain old locks" (in our case `UninterruptableMonitor.Enter(syncLock)`) instead of using more advanced techniques that are available in C# to avoid locks, such as `ReaderWriterLockSlim`.

Lucene.Net.Facet has been refactored several times to try various ways to synchronize locks so it would be prudent to go back to basics to find a solution that is stable first. Once we have found a stable solution, we can begin looking at ways to avoid locks and ease into them to ensure concurrency is stable.

Jeevananthan-23:
@NightOwl888, as a part of investigating Lucene.Net.Facet and making sure the concurrency for DrillSideways class should implement as Lucene(Java) DrillSideways implementation with ConcurrentDrillSideways [here](https://github.com/apache/lucene/blob/0a525ce2aba572855ce045e77f3a8c9acac5a199/lucene/facet/src/java/org/apache/lucene/facet/DrillSideways.java#L123) 

 On large document set, the single threaded collection can be really slow #817

PR #838, opened 2023-04-21
Fix slow test: use different sleep method if resolution is low

laimis:
This fixes an issue with a slow test in TestTaxonomyCombined. The original issue description is here:

The issue has been brought up in https://github.com/apache/lucenenet/issues/836, under `TestTaxonomyCombined.TestTaxonomyRefreshRaces()` section

After debugging the issue, I discovered that the cause of the slow down are System.Thread.Sleep(1) calls being made thousands of times with intent of sleep taking 1ms. On windows however, that could take much more than 1ms, 15ms normally, confirmed locally and found a discussion about it here:

https://stackoverflow.com/questions/19066900/thread-sleep1-takes-longer-than-1ms

I setup Java version locally to compare the running times and after the fix, Java version and .NET version is running in basically identical amount of time. Locally I get it to run in 3-4s range, the same with Java version. Before the fix, it took 30-40s depending on what random generated loop variable values were.

I found some ways to influence System.Threading.Thread.Sleep behavior here https://stackoverflow.com/questions/15071359/how-to-set-timer-resolution-from-c-sharp-to-1-ms but it seemed like it's not worth the effort if the class is used just for that Facet test fixture.

### DIFF
--- a/src/Lucene.Net.TestFramework/Index/RandomCodec.cs
+++ b/src/Lucene.Net.TestFramework/Index/RandomCodec.cs
@@ -96,10 +96,10 @@ namespace Lucene.Net.Index
                 if (Debugging.AssertsEnabled) Debugging.Assert(previousMappings.Count < 10000, "test went insane");
             }
 
-            //if (LuceneTestCase.VERBOSE)
-            //{
-            Console.WriteLine("RandomCodec.GetPostingsFormatForField(\"" + name + "\") returned '" + codec.Name + "' with underlying type '" + codec.GetType().ToString() + "'.");
-            //}
+            if (LuceneTestCase.Verbose)
+            {
+                Console.WriteLine("RandomCodec.GetPostingsFormatForField(\"" + name + "\") returned '" + codec.Name + "' with underlying type '" + codec.GetType().ToString() + "'.");
+            }
 
             return codec;
         }
@@ -119,10 +119,10 @@ namespace Lucene.Net.Index
                 if (Debugging.AssertsEnabled) Debugging.Assert(previousDVMappings.Count < 10000, "test went insane");
             }
 
-            //if (LuceneTestCase.VERBOSE)
-            //{
+            if (LuceneTestCase.Verbose)
+            {
                 Console.WriteLine("RandomCodec.GetDocValuesFormatForField(\"" + name + "\") returned '" + codec.Name + "' with underlying type '" + codec.GetType().ToString() + "'.");
-            //}
+            }
 
             return codec;
         }
--- a/src/Lucene.Net.Tests.Facet/SlowRAMDirectory.cs
+++ b/src/Lucene.Net.Tests.Facet/SlowRAMDirectory.cs
@@ -1,7 +1,6 @@
 ﻿// Lucene version compatibility level 4.8.1
-using Lucene.Net.Support.Threading;
-using RandomizedTesting.Generators;
 using System;
+using System.Diagnostics;
 using System.Threading;
 
 namespace Lucene.Net.Facet
@@ -39,7 +38,7 @@ namespace Lucene.Net.Facet
 
         internal Random random;
         private int sleepMillis;
-
+        
         public virtual void SetSleepMillis(int sleepMillis)
         {
             this.sleepMillis = sleepMillis;
@@ -80,7 +79,20 @@ namespace Lucene.Net.Facet
 
             try
             {
-                Thread.Sleep(sTime);
+                // LUCENENET specific - timer resolution on windows by default is 15ms and small sleeps will end up taking
+                // 15ms instead causing certain tests to run slowly. Instead, we'll use a Stopwatch to get more accurate
+                if (sTime > 15)
+                {
+                    Thread.Sleep(sTime);
+                }
+                else
+                {
+                    var timer = Stopwatch.StartNew();
+                    while(timer.ElapsedMilliseconds < sTime)
+                    {
+                        continue;
+                    }
+                }
             }
             catch (Exception ie) when (ie.IsInterruptedException())
             {
@@ -188,7 +200,7 @@ namespace Lucene.Net.Facet
             private readonly IndexOutput io;
             private int numWrote;
             private readonly Random rand;
-
+            
             public SlowIndexOutput(SlowRAMDirectory outerInstance, IndexOutput io)
             {
                 this.outerInstance = outerInstance;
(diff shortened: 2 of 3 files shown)


src/Lucene.Net.TestFramework/Index/RandomCodec.cs
```diff
@@ -96,10 +96,10 @@ public override PostingsFormat GetPostingsFormatForField(string name)
...
             }
 
-            //if (LuceneTestCase.VERBOSE)
-            //{
-            Console.WriteLine("RandomCodec.GetPostingsFormatForField(\"" + name + "\") returned '" + codec.Name + "' with underlying type '" + codec.GetType().ToString() + "'.");
-            //}
+            if (LuceneTestCase.Verbose)
+            {
+                Console.WriteLine("RandomCodec.GetPostingsFormatForField(\"" + name + "\") returned '" + codec.Name + "' with underlying type '" + codec.GetType().ToString() + "'.");
+            }
```
laimis, line 103: Put this back as they were showing up in console and interfering with debugging.

laimis:
The failing tests do not appear to be there due to my changes.

return new J2N.Randomizer(random.NextInt64());

NextInt643= is not available on Random before .net 6. Not sure what the fix here is for that, let me know if you have any suggestions, @NightOwl888 . Thanks!

laimis:
I think I see what I can do here next, remove .net5 and .net46 targets for the facet project should make the issue go away. Me removing Slow attribute enabled a new code path. Should be able to adjust later today.

NightOwl888:
> Not sure how to proceed here. It feels like we either stop using NextInt64 or drop .net5?

[NextInt64()](https://github.com/NightOwl888/RandomizedTesting/blob/main/src/RandomizedTesting.Generators/Support/RandomExtensions.cs#L119) is an extension method in RandomizedTesting.Generators. However, I noticed that it doesn't cascade the call to the correct random implementation when the `Random` instance is a `J2N.Randomizer`, which I have submitted a fix for in https://github.com/NightOwl888/RandomizedTesting/pull/1. That won't affect what you are doing, though and we can upgrade RandomizedTesting.Generators later.

We most definitely do not want to drop tests for `net5.0` and `net461` as that would leave a gap in our test coverage. Namely, we wouldn't have any tests for the `netstandard2.1` or `netstandard2.0` assemblies.

src/Lucene.Net.Tests.Facet/Taxonomy/TestTaxonomyCombined.cs
```diff
@@ -784,7 +784,6 @@ public virtual void TestChildrenArraysGrowth()
 
         // Test that getParentArrays is valid when retrieved during refresh
         [Test]
-        [Slow]
```
NightOwl888, line 787: I have been using 4 seconds on my machine as a minimum for what is considered a slow test. This is actually right on the edge after the above updates, so this could go either way.

src/Lucene.Net.Tests.Facet/SlowRAMDirectory.cs
```diff
@@ -80,7 +80,20 @@ internal virtual void DoSleep(Random random, int length)
 
             try
             {
-                Thread.Sleep(sTime);
+                // LUCENENET specific - timer resolution on windows by default is 15ms and small sleeps will end up taking
+                // 15ms instead causing certain tests to run slowly. Instead, we'll use a Stopwatch to get more accurate
+                if (sTime > MINIMUM_SUPPORTED_THREAD_SLEEP_TIME)
```
NightOwl888, line 85: Good work finding this cause!

However, the resolution of `Stopwatch` varies depending on the OS, so we cannot use a constant for this. I used ChatGPT to come up with a solution that

1. Doesn't allocate a `Stopwatch` instance
2. Doesn't block the CPU
3. Correctly calculates the number of ticks to wait for the current platform
4. Can be used as a drop-in replacement for `Thread.Sleep()` in this instance

```c#
internal virtual void DoSleep(Random random, int length)
{
    int sTime = length < 10 ? sleepMillis : (int)(sleepMillis * Math.Log(length));
    if (random != null)
    {
        sTime = random.Next(sTime);
    }

    try
    {
        Sleep(sTime);
    }
    catch (Exception ie) when (ie.IsInterruptedException())
    {
        throw new Util.ThreadInterruptedException(ie);
    }
}

internal static void Sleep(int milliseconds)
{
    long ticks = (long)((Stopwatch.Frequency / (double)1000) * milliseconds); // ticks per millisecond * milliseconds = total delay ticks
    long initialTick = Stopwatch.GetTimestamp();
    long targetTick = initialTick + ticks;
    while (Stopwatch.GetTimestamp() < targetTick)
    {
        Thread.SpinWait(1);
    }
}
```

On my system this branch takes at least 6 seconds to run `TestTaxonomyReaderRefreshRaces()`, but with these changes it drops to 4 or 5 seconds.

I confirmed that the CPU is not pegged at 100% when `Sleep()` runs, but I haven't done any testing other than on Windows.

NightOwl888 (changes requested):

laimis:
Figured out what's up with the NextInt64 call, I removed the using for extension method as my IDE showed it as not being used, but as we can see, that gets used under different .net framework versions. I will have to keep that in mind.

laimis:
@NightOwl888 ready for your review again.

NightOwl888:
> Figured out what's up with the NextInt64 call, I removed the using for extension method as my IDE showed it as not being used, but as we can see, that gets used under different .net framework versions. I will have to keep that in mind.
Actually, since we consistently use J2N.Randomizer, it will call the overridden NextInt64() since it is a subclass of System.Random. The concrete method takes precedence over the extension method. But as previously mentioned, the extension method currently has a different implementation than the concrete method, so it will vary depending on whether dotnet supplies the method or not.

src/Lucene.Net.Tests.Facet/Taxonomy/TestTaxonomyCombined.cs
NightOwl888: Let's keep the `[Slow]` attribute since this is almost always over 4 seconds on my machine even after the improvement.

This can be used to get a faster run by setting the system property `tests: slow` to `false` to exclude the tests that are marked with this attribute. We currently only use that for the CLI installer test in GitHub actions, but it is also possible to set in Azure DevOps by passing in `RunSlowTests` as `false`.

NightOwl888 (changes requested):
Just 1 small change to make.

As we discussed in Slack, let's also open an [issue in J2N](https://github.com/NightOwl888/J2N/issues) to work on a more precise `ThreadJob.Sleep()` implementation for general use in Lucene.NET. This implementation (or one very similar) might work there, but we will need to use the `TestThreadInterruptDeadlock()` tests to confirm that it will throw a `ThreadInterruptedException` when `Thread.Interrupt()` is called. If it doesn't, we will need to find a way to throw one.

NightOwl888 (approved):

PR #844, opened 2023-04-27
BUG: Lucene.Net.Util.RandomizedContext: Create a separate instance of Randomizer() for each thread

### LINKED ISSUE #843: Random property used by the tests needs to return new instance of "Randomness generator" per thread
A random number generation discrepancy was discovered while troubleshooting slow running tests in Lucene.NET and comparing the behavior to Java Lucene. It appears that whenever random() was being called in Java, (e.g. here: ), Java version would create a new instance of random PER thread, initialized with the same seed. In our implementation, we return the same random instance.

Arguably our approach is better for generating more random data and multiple threads in the same test not generating the same numbers but it ends up resulting in .NET version tests having wildly different running times, making it hard to spot real issues.

.NET version should generate Random instance per thread.

NightOwl888:
Fixes #843.

### DIFF
--- a/src/Lucene.Net.TestFramework/Support/Util/RandomizedContext.cs
+++ b/src/Lucene.Net.TestFramework/Support/Util/RandomizedContext.cs
@@ -2,6 +2,7 @@
 using NUnit.Framework.Internal;
 using System;
 using System.Reflection;
+using System.Threading;
 
 namespace Lucene.Net.Util
 {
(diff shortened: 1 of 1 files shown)


laimis:
Applied the changes locally, can confirm that TestBig() looks better now.

laimis:
This also fixes #836. The randomization seed being reused across threads ends up triggering caching paths as well as generating less data during the tests, thus matching java Lucene behavior and running times.

PR #852, opened 2023-05-11
Upgrade Lucene.Net.Analysis.OpenNLP to target .net core 6, in addition to .net framework

### LINKED ISSUE #851: Add .NET Core support for Lucene.Net.Analysis.OpenNLP
As part of the 4.8 release, we should add .NET Core support for Lucene.Net.Analysis.OpenNLP project.

It seems like this will involve updating OpenNLP.NET to 1.9.4.1 and adjust target frameworks to include whatever .net verrsions we can support with that version of OpenNLP.NET.

laimis:
Closes #851

### DIFF
--- a/.build/TestReferences.Common.targets
+++ b/.build/TestReferences.Common.targets
@@ -25,8 +25,4 @@
     <PackageReference Include="NUnit" Version="$(NUnitPackageVersion)" />
     <PackageReference Include="NUnit3TestAdapter" Version="$(NUnit3TestAdapterPackageVersion)" />
   </ItemGroup>
-
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' ">
-    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="$(SystemRuntimeCompilerServicesUnsafePackageVersion)" />
-  </ItemGroup>
 </Project>
--- a/.build/azure-templates/publish-test-results-for-test-projects.yml
+++ b/.build/azure-templates/publish-test-results-for-test-projects.yml
@@ -102,7 +102,28 @@ steps:
     testResultsArtifactName: '${{ parameters.testResultsArtifactName }}'
     testResultsFileName: '${{ parameters.testResultsFileName }}'
 
-# Special case: Only supports net48
+# Special case: Only supports net7.0, net6.0 and net48
+
+- template: publish-test-results.yml
+  parameters:
+    testProjectName: 'Lucene.Net.Tests.Analysis.OpenNLP'
+    framework: 'net7.0' # Since condtions are not supported for templates, we check for the file existence within publish-test-results.yml
+    vsTestPlatform: '${{ parameters.vsTestPlatform }}'
+    osName: '${{ parameters.osName }}'
+    testResultsFormat: '${{ parameters.testResultsFormat }}'
+    testResultsArtifactName: '${{ parameters.testResultsArtifactName }}'
+    testResultsFileName: '${{ parameters.testResultsFileName }}'
+
+- template: publish-test-results.yml
+  parameters:
+    testProjectName: 'Lucene.Net.Tests.Analysis.OpenNLP'
+    framework: 'net6.0' # Since condtions are not supported for templates, we check for the file existence within publish-test-results.yml
+    vsTestPlatform: '${{ parameters.vsTestPlatform }}'
+    osName: '${{ parameters.osName }}'
+    testResultsFormat: '${{ parameters.testResultsFormat }}'
+    testResultsArtifactName: '${{ parameters.testResultsArtifactName }}'
+    testResultsFileName: '${{ parameters.testResultsFileName }}'
+
 - template: publish-test-results.yml
   parameters:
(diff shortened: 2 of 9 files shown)


NightOwl888:
@laimis - I have fixed the dependency conflict. But this still doesn't include the changes for  `<MavenReference>`. Since that basically builds on this, I will create a branch of this branch to add those changes. If it doesn't work, we can merge this branch instead.

NightOwl888:
Okay, looks like MavenReference is a bust for now: https://github.com/ikvmnet/ikvm-maven/issues/10#issuecomment-1550300017.

The fix for that is similar to how we do the file locking in Lucene.NET on macOS in the [NativeFSLockFactory](https://github.com/apache/lucenenet/blob/19cb0c1d806462aad2b6297e0885491062887456/src/Lucene.Net/Store/NativeFSLockFactory.cs). Unfortunately, IKVM requires nearly 100 GB to build, and I don't have that kind of disk space to spare. But if you want to have a go at it, you are welcome.

We have yet another issue with this addition - referencing IKVM adds about 4 minutes of build time to Azure DevOps and will almost certainly slow down the GitHub Actions builds as well. Unfortunately, the size of the IKVM NuGet package and dependencies has exploded. In both Azure DevOps and GitHub Actions, there are options for caching NuGet packages so they don't have to be downloaded per build. We used to have it set up on Azure DevOps, but it wasn't set up to invalidate the cache when `.build/dependencies.props` changes and I couldn't figure out how to get it cleared after versions of dependencies were updated. So, we will need to investigate this, also. The cache key we need to use should include all `*.props`, `*.targets` and `*.csproj` files.

laimis:
Nice, thanks for the changes there, it looks like it successfully completed on azure:

https://dev.azure.com/lucene-net-temp4/Main/_build?definitionId=1

Could we use the changes made in this branch as is to release Analysis.OpenNLP? I am not following why is IKVM needed. I definitely have the space/resources to build it, just not sure what is needed.

GAInTheHouse:
@laimis @NightOwl888 would y'all be able to help me understand what's the road map for OpenNLP to target .Net 6?

NightOwl888:
> @laimis @NightOwl888 would y'all be able to help me understand what's the road map for OpenNLP to target .Net 6?

Unfortunately, this is blocked by the fact that we don't really want to use OpenNLP.NET for this, but instead use `<MavenReference>`, which will transitively be added to everyone who depends on `Lucene.Net.Analysis.OpenNLP`. Basically, instead of including a compiled DLL in the package for opennlp-tools, it will compile it on your machine when you depend on `Lucene.Net.Analysis.OpenNLP` and it will resolve any other Maven dependencies that you pull into your project to fix versioning conflicts between common dependencies.

However, I have updated this branch (rebased against the master branch) and have built the packages [here](https://dev.azure.com/shad0962/Experiments/_build/results?buildId=2240&view=artifacts&pathAsName=false&type=publishedArtifacts). Download the `nuget` artifact and unzip it, and you can then put the `.nupkg` files into a local directory and add it as a package source. That should at least get you up and running for now.

The tests we have for the OpenNLP support work, but I am not sure the entire opennlp-tools package works on macOS or iOS. It won't if there is any file locking in the application, but I haven't checked the opennlp source to determine if they are doing so. It should work on Windows and Linux, though.

As for an official release, we will do so as soon as there is a version of IKVM that supports both macOS and .NET Framework, but so far there isn't one and the latest version has some annoying initialization exceptions if you don't explicitly call `new object()` before you call into an IKVM-converted library. IKVM has come a long way in the past 18 months, but it just isn't there yet.

NightOwl888:
@GAInTheHouse 

It turns out it was possible to overcome the issues with `MavenReference` with some due diligence. I have submitted #892, which should resolve this (even on macOS). You can download the latest update with these changes here:

https://dev.azure.com/shad0962/Experiments/_build/results?buildId=2242&view=artifacts&pathAsName=false&type=publishedArtifacts

A few things to note:

- Sometimes it may be necessary to physically delete your bin and obj folders in order to fix IKVM build issues, and these build issues (such as type initialization errors) often look like runtime issues.
- IKVM will download several GB of dependencies onto your build machine. So, it is important to ensure there is enough space.
- Nearly all issues you are likely to encounter will either be with [IKVM](https://github.com/ikvmnet/ikvm) or [OpenNLP](https://opennlp.apache.org/get-involved.html) (the Java project), so please report them to the appropriate party.  Of course, there could still be a problem with the Lucene.Net.Analysis.OpenNLP project, and we would like to hear feedback if you do find a problem with it.
- Since there will likely be a learning curve to get into the weeds with OpenNLP, it would be appreciated if you contribute what you learn as documentation back to this project so it can help others get up to speed more easily (as you pointed out in #890, there isn't much to go on). Documentation for OpenNLP can be provided via Markdown in the following file: https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Analysis.OpenNLP/overview.md. Additional documentation can be provided on class and members via inline XML doc comments. Documents can also be added as Markdown files and then added to the `/websites/apidocs` folder. There is some information about how to build the docs here: https://lucenenet.apache.org/contributing/documentation.html.

PR #863, opened 2023-06-14
Fix two typos in quick-start introduction.md

cucoreanu:


rclabo:
Thanks for these fixes.  They look good.  I wish merging them in were easier.  

Committing the PR is easy, of course, but the process to get the committed file published to the website is currently manual, and it always takes me a bit of studying to remember how to make that happen.  I don't have the time to dig into that right now, so it could be a few weeks before I can look into that.  

Then again, someone else on the committer team may jump in and handle it.

Shazwazza:
I thought the website build would be automatic for any commit to this path: https://github.com/apache/lucenenet/blob/51cc7418adf9da96cd37dd64ff5085d752a54074/.github/workflows/Lucene-Net-Website.yml#L37

But maybe, it is only automatic in combination with a release tag when we publish a release.

In any case, more docs are here https://github.com/apache/lucenenet-site

But I think for non-release site changes we can just manually run this Action against main branch: https://github.com/apache/lucenenet/actions/workflows/Lucene-Net-Website.yml

rclabo:
Thanks @Shazwazza 

 Yah, I put some docs here too: https://github.com/apache/lucenenet-site 

> I thought the website build would be automatic for any commit to this path: 'websites/site/**/*' 

I don't know, it hasn't been working that way. I've been having to do a manual PR to get the files over to https://github.com/apache/lucenenet-site/tree/asf-site and it always takes me a bit to research and re-remember the flow and how to run DocFX locally on my machine to try things out first.

> But maybe, it is only automatic in combination with a release tag when we publish a release.

Hmmm. That's an interesting clue.  Maybe.

NightOwl888:
The docs build works from GitHub Actions when triggered manually from the control panel. But for some reason it is not triggering automatically with the path: https://github.com/apache/lucenenet/blob/51cc7418adf9da96cd37dd64ff5085d752a54074/.github/workflows/Lucene-Net-Website.yml#L37.

I haven't been able to work out why that is happening. The path looks right and the pattern matches other workflow files that are working.

I haven't complained loudly about it because the manual trigger is a simple workaround. @rclabo - there is no need to pull the branch down to your dev box unless you want to debug or preview something.

NightOwl888 (approved):

PR #864, opened 2023-07-24
[backport] GITHUB-11838 Add api to allow concurrent query rewrite

sbiaudet:
To align with PR [#11840](https://github.com/apache/lucene/pull/11840). 

Change signature of rewrite to rewrite(IndexSearcher)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,4 +65,4 @@ websites/apidocs/api/**/*.manifest
 svn-*/
 
 # vscode files
-.vscode/
+.vscode/*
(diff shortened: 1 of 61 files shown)


NightOwl888:
Closing, as this is a feature that didn't exist in Lucene 4.8.0 and has breaking API changes that will be confusing to users.

We will gain this feature automatically when we do the upgrade. Getting out of sync with Lucene releases puts us in a situation where we have a library that is impossible to debug because there is nothing in Java to compare execution paths with.

PR #867, opened 2023-07-29
ci: Update action versions

nikcio:
I noticed that the action versions that are being used are a bit outdated so I've updated them to the newest versions. As far as I could see there shouldn't be any breaking changes that should break any of the actions.

image

This also updates the Sonar action based on this warning.
image

### DIFF
--- a/.github/workflows/Generate-TestWorkflows.ps1
+++ b/.github/workflows/Generate-TestWorkflows.ps1
@@ -274,7 +274,7 @@ jobs:
 
     steps:
       - name: Checkout Source Code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Disable .NET SDK Telemetry and Logo
         run: |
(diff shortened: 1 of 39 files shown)


NightOwl888 (approved):
Thanks for the PR!

NightOwl888 (commented):
It looks like checkout has incremented to V4: https://github.com/actions/checkout/releases. Would you like to change this before I merge? No worries if you don't.

nikcio:
@NightOwl888 Lets save that for a future update it don't think i'll have time to look at it soon so lets just get these updates in first :wink:

NightOwl888:
@nikcio - Hmm...although the scan shows no errors, the "Build and Analyze" step returns an exit code of 1 instead of 0 as it did previously.

```console
2023-10-14T14:09:59.9972498Z     17280 Warning(s)
2023-10-14T14:09:59.9972901Z     0 Error(s)
2023-10-14T14:09:59.9973138Z 
2023-10-14T14:09:59.9973316Z Time Elapsed 00:10:58.10
2023-10-14T14:09:59.9973810Z SonarScanner for MSBuild 5.9.2
2023-10-14T14:09:59.9974486Z Using the .NET Core version of the Scanner for MSBuild
2023-10-14T14:09:59.9975241Z Post-processing started.
2023-10-14T14:10:00.0027392Z ##[error]Process completed with exit code 1.
```
[logs_11749.zip](https://github.com/apache/lucenenet/files/12906857/logs_11749.zip)

Offhand, I don't see a change that could cause this. But since the scan is scheduled to run every day, it definitely didn't fail before this PR. Thoughts?

NightOwl888:
Okay, I figured out the issue. Somehow updating the cache versions caused something to go out of sync with the cache. This would have probably been fine, but [this line](https://github.com/apache/lucenenet/commit/179356fb485bba1ed94f386d754f8de0e3eb1293) had an incorrect path and 1 of the 2 caches wasn't being generated.

I also discovered that there is now a simple way to delete the cache via API or in the control panel: https://stackoverflow.com/a/64819132, which is very helpful.

PR #868, opened 2023-07-29
Use expression instead of file for NuGet license

mus65:
You can't set both and expressions are better because it allows tools like SBOM generators to automatically determine the license.

see also https://learn.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-a-license-expression-or-a-license-file

### DIFF
--- a/.build/nuget.props
+++ b/.build/nuget.props
@@ -25,7 +25,7 @@
     <Authors>The Apache Software Foundation</Authors>
     <PackageProjectUrl>https://lucenenet.apache.org</PackageProjectUrl>
     <PackageIcon>lucene-net-icon-128x128.png</PackageIcon>
-    <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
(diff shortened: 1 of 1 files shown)


NightOwl888:
Thanks for the PR.

However, this does not comply with [Apache's licensing policy](https://www.apache.org/legal/release-policy.html#licensing).

1. For source packages, LICENSE and NOTICE MUST be located at the root of the distribution. For additional packages, they MUST be located in the distribution format's customary location for licensing materials, such as the META-INF directory of Java "jar" files.
2. When a package bundles code under several licenses, the LICENSE file MUST contain details of all these licenses. For each component which is not Apache licensed, details of the component MUST be appended to the LICENSE file. The component license itself MUST either be appended or else stored elsewhere in the package with a pointer to it from the LICENSE file, e.g. if the license is long.

As parts of Lucene.NET are covered by more than once license, this licensing expression doesn't meet our needs. See: https://github.com/apache/lucenenet/blob/51cc7418adf9da96cd37dd64ff5085d752a54074/LICENSE.txt#L206-L751

PR #909, opened 2024-01-30
website: Added missing namespaces to tutorial

kant2002:
Port of https://github.com/apache/lucenenet-site/pull/46

Let's make sample copy-able


NightOwl888 (approved):
Thanks for the PR.

It looks like Lucene.Net.Index and Lucene.Net.Documents are already imported. But, it is possible there are namespaces with conflicting names, so this is probably a good thing to do to make the sample more robust.

NightOwl888 (approved):
Thanks for the PR.

It looks like Lucene.Net.Index and Lucene.Net.Documents are already imported. But, it is possible there are namespaces with conflicting names, so this is probably a good thing to do to make the sample more robust.

PR #926, opened 2024-03-10
Test review Core I-J

paulirwin:

Partial work for #259 for the core I-J tests assembly.

### DIFF
--- a/src/Lucene.Net.Tests/Index/TestIndexWriterExceptions.cs
+++ b/src/Lucene.Net.Tests/Index/TestIndexWriterExceptions.cs
@@ -12,6 +12,7 @@ using NUnit.Framework;
 using RandomizedTesting.Generators;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Threading;
 using Assert = Lucene.Net.TestFramework.Assert;
@@ -214,7 +215,7 @@ namespace Lucene.Net.Index
                         Console.WriteLine(Thread.CurrentThread.Name + ": TEST: IndexerThread: cycle");
                     }
                     outerInstance.doFail.Value = this.Instance;
-                    string id = "" + r.Next(50);
+                    string id = r.Next(50).ToString(CultureInfo.InvariantCulture); // LUCENENET: using InvariantCulture ToString overload instead of implicit `"" + r.Next(50)`
                     idField.SetStringValue(id);
                     Term idTerm = new Term("id", id);
                     try
@@ -2134,7 +2135,7 @@ namespace Lucene.Net.Index
                 for (int i = 0; i < numDocs; i++)
                 {
                     Document doc = new Document();
-                    doc.Add(new StringField("id", (docBase + i).ToString(), Field.Store.NO));
+                    doc.Add(new StringField("id", (docBase + i).ToString(CultureInfo.InvariantCulture), Field.Store.NO));
                     if (DefaultCodecSupportsDocValues)
                     {
                         doc.Add(new NumericDocValuesField("f", 1L));
@@ -2172,20 +2173,20 @@ namespace Lucene.Net.Index
                                 }
                                 if (Random.NextBoolean()) // update only numeric field
                                 {
-                                    w.UpdateNumericDocValue(new Term("id", (docBase + i).ToString()), "f", value);
-                                    w.UpdateNumericDocValue(new Term("id", (docBase + i).ToString()), "cf", value * 2);
+                                    w.UpdateNumericDocValue(new Term("id", (docBase + i).ToString(CultureInfo.InvariantCulture)), "f", value);
+                                    w.UpdateNumericDocValue(new Term("id", (docBase + i).ToString(CultureInfo.InvariantCulture)), "cf", value * 2);
                                 }
                                 else if (Random.NextBoolean())
                                 {
-                                    w.UpdateBinaryDocValue(new Term("id", (docBase + i).ToString()), "bf", TestBinaryDocValuesUpdates.ToBytes(value));
-                                    w.UpdateBinaryDocValue(new Term("id", (docBase + i).ToString()), "bcf", TestBinaryDocValuesUpdates.ToBytes(value * 2));
+                                    w.UpdateBinaryDocValue(new Term("id", (docBase + i).ToString(CultureInfo.InvariantCulture)), "bf", TestBinaryDocValuesUpdates.ToBytes(value));
+                                    w.UpdateBinaryDocValue(new Term("id", (docBase + i).ToString(CultureInfo.InvariantCulture)), "bcf", TestBinaryDocValuesUpdates.ToBytes(value * 2));
                                 }
                                 else
                                 {
-                                    w.UpdateNumericDocValue(new Term("id", (docBase + i).ToString()), "f", value);
-                                    w.UpdateNumericDocValue(new Term("id", (docBase + i).ToString()), "cf", value * 2);
-                                    w.UpdateBinaryDocValue(new Term("id", (docBase + i).ToString()), "bf", TestBinaryDocValuesUpdates.ToBytes(value));
-                                    w.UpdateBinaryDocValue(new Term("id", (docBase + i).ToString()), "bcf", TestBinaryDocValuesUpdates.ToBytes(value * 2));
+                                    w.UpdateNumericDocValue(new Term("id", (docBase + i).ToString(CultureInfo.InvariantCulture)), "f", value);
+                                    w.UpdateNumericDocValue(new Term("id", (docBase + i).ToString(CultureInfo.InvariantCulture)), "cf", value * 2);
+                                    w.UpdateBinaryDocValue(new Term("id", (docBase + i).ToString(CultureInfo.InvariantCulture)), "bf", TestBinaryDocValuesUpdates.ToBytes(value));
+                                    w.UpdateBinaryDocValue(new Term("id", (docBase + i).ToString(CultureInfo.InvariantCulture)), "bcf", TestBinaryDocValuesUpdates.ToBytes(value * 2));
                                 }
                             }
 
@@ -2194,10 +2195,10 @@ namespace Lucene.Net.Index
                             {
                                 if (Verbose)
                                 {
-                                    Console.WriteLine("  delete id=" + (docBase + i));
+                                    Console.WriteLine("  delete id=" + (docBase + i).ToString(CultureInfo.InvariantCulture));
                                 }
                                 deleteCount++;
-                                w.DeleteDocuments(new Term("id", "" + (docBase + i)));
+                                w.DeleteDocuments(new Term("id", (docBase + i).ToString(CultureInfo.InvariantCulture))); // LUCENENET: using InvariantCulture ToString overload instead of implicit `"" +` conversion
                             }
                         }
                     }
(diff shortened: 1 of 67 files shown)


src/Lucene.Net.Tests/Index/TestTermsEnum.cs
```diff
@@ -435,10 +435,16 @@ public virtual void TestIntersectRandom()
             dir.Dispose();
         }
 
-        private readonly string FIELD = "field";
+        // LUCENENET: passing directory in as parameter to MakeIndex and using returned IndexReader
```
NightOwl888, line 438: What is the reasoning behind this change?

NightOwl888 (changes requested):
Thanks for the PR.

There are a few requested changes as well as some other feedback.

NightOwl888:
It looks like there were a couple of omissions.

1. [This line](https://github.com/apache/lucenenet/blob/b1476aee4fe21131899c1f43b2e06e25971b3ebe/src/Lucene.Net/Index/IndexWriter.cs#L2607) will need to be reenabled to prevent the `NONE` option from being passed.
2. [This line](https://github.com/apache/lucenenet/blob/b1476aee4fe21131899c1f43b2e06e25971b3ebe/src/Lucene.Net.Tests/Index/TestNoMergeScheduler.cs#L37) will need to be fixed to exclude the `NONE` option from being tested. [Here](https://github.com/apache/lucenenet/blob/b1476aee4fe21131899c1f43b2e06e25971b3ebe/src/Lucene.Net.TestFramework/Index/BasePostingsFormatTestCase.cs#L489) is how we do it elsewhere.

I am still torn on whether to just hide the `NONE` option from users because it really isn't a thing anywhere but in the `FindMerges()` method and it isn't very likely users would enable asserts to find out that they are passing something that isn't valid to calls to other methods that accept `MergeTrigger`. We could add guard clauses to all of those methods to make it fail without enabling asserts, but it is much simpler just to make it invisible.

src/Lucene.Net/Index/MergeTrigger.cs
```diff
@@ -24,6 +24,11 @@ namespace Lucene.Net.Index
     /// </summary>
     public enum MergeTrigger
     {
+        /// <summary>
+        /// LUCENENET-specific value to be used instead of null.
+        /// </summary>
+        NONE = -1,
```
NightOwl888, line 30: We would need to set this to 0 (and number the others one higher) to match what Java does here. When you declare an unset MergeTrigger in Java, it is `null`. So, we would want to use `NONE` for an unset `MergeTrigger`. Hopefully nobody does it, but it is still better than having an unset `MergeTrigger` set to `SEGMENT_FLUSH`.

src/Lucene.Net.Tests/Index/TestIndexWriterExceptions.cs
```diff
@@ -694,7 +695,7 @@ public override void Eval(MockDirectoryWrapper dir)
                 {
                     // LUCENENET specific: for these to work in release mode, we have added [MethodImpl(MethodImplOptions.NoInlining)]
                     // to each possible target of the StackTraceHelper. If these change, so must the attribute on the target methods.
-                    bool sawAppend = StackTraceHelper.DoesStackTraceContainMethod(typeof(FreqProxTermsWriterPerField).Name, "Flush");
+                    bool sawAppend = StackTraceHelper.DoesStackTraceContainMethod(nameof(FreqProxTermsWriterPerField), "Flush");
```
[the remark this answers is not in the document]
paulirwin: Filed #931

src/Lucene.Net.Tests/Index/TestIndexWriterExceptions.cs
```diff
@@ -2185,10 +2194,10 @@ public virtual void TestNoLostDeletesOrUpdates()
                             {
                                 if (Verbose)
                                 {
-                                    Console.WriteLine("  delete id=" + (docBase + i).ToString());
+                                    Console.WriteLine("  delete id=" + (docBase + i));
                                 }
                                 deleteCount++;
-                                w.DeleteDocuments(new Term("id", "" + (docBase + i).ToString()));
+                                w.DeleteDocuments(new Term("id", "" + (docBase + i)));
```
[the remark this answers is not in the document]
paulirwin: While I think that's unnecessary in these tests given that an incorrect conversion would fail the test, I'll update them anyways as examples of handling the numbers in a culture-invariant way.

src/Lucene.Net.Tests/Index/TestIndexWriterExceptions.cs
```diff
@@ -2351,9 +2360,9 @@ public override void Eval(MockDirectoryWrapper dir)
                     if (Verbose)
                     {
                         Console.WriteLine("TEST: now fail; thread=" + Thread.CurrentThread.Name + " exc:");
-                        Console.WriteLine((new Exception()).StackTrace);
+                        Console.WriteLine(new Exception().StackTrace);
```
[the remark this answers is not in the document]
paulirwin: Filed #932

src/Lucene.Net.Tests/Index/TestPayloads.cs
```diff
@@ -85,9 +83,7 @@ public virtual void TestPayloadFieldBit()
             // enabled in only some documents
             d.Add(NewTextField("f3", "this field has payloads in some docs", Field.Store.NO));
             // only add payload data for field f2
-#pragma warning disable 612, 618
-            analyzer.SetPayloadData("f2", "somedata".GetBytes(IOUtils.CHARSET_UTF_8), 0, 1);
-#pragma warning restore 612, 618
+            analyzer.SetPayloadData("f2", "somedata".GetBytes(Encoding.UTF8), 0, 1);
```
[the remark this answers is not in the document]
paulirwin: I'll do the former for now, and revert the change to use the IOUtils field.

src/Lucene.Net.Tests/Index/TestTermsEnum.cs
paulirwin: There's a couple current and former changes here that you might be referring to:

1. Changed FIELD to be const, I assume that's self explanatory, it was only used as a constant string.
2. Added comment with the original fields that were previously moved (prior to this PR) in line to the methods that use them, for future line-by-line reference with the Java code.
3. The prior change that moved the fields to be in line in the methods... I am unsure of the reasoning, but I assumed that the statefulness of these fields on the test fixture caused some problems with the tests. Let me know your thoughts on if you think this should be reverted to be fields again.

src/Lucene.Net/Index/MergeTrigger.cs
paulirwin: In Java, SEGMENT_FLUSH at least has ordinal 0 even though it doesn't have an integer value like in .NET, so that was my reasoning, but I don't feel strongly about this so I'll go with your suggestion.

src/Lucene.Net.Tests/Index/TestIndexWriterExceptions.cs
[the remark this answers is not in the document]
NightOwl888: It is unnecessary to make the code work. But IMO, it is necessary to show future maintainers that we actually checked to ensure this conversion is correct. I suppose adding a comment after each one of these lines to indicate we checked the logic would also work, but explicitly passing the argument would be more concise and easier to make uniform.

src/Lucene.Net.Tests/Index/TestTermsEnum.cs
NightOwl888: No, I was referring to the part that diverged from Lucene, here. The fact that the `directory` is not being newed up inside of the `MakeIndex()` method and disposed afterward.

> The prior change that moved the fields to be in line in the methods... I am unsure of the reasoning, but I assumed that the statefulness of these fields on the test fixture caused some problems with the tests.

This was done during the attempts to convert the test framework to be compatible with xUnit, and thus be able to run the tests in parallel. Ultimately we found that using xUnit was possible, but impractical. Either we had to do major refactoring of several of the production types or we had to configure xUnit to run the tests serially. But even if we configured it to run serially, it required our users to implement constructors that pass through parameters from xUnit whenever they inherited `LuceneTestCase` or any subclass.

At this point, we have totally abandoned the approach and gone back to making things static as they were in Java. So, it is fine to revert any code you find that does this. We run the tests in parallel by dividing them up into multiple projects that run in parallel instead.

paulirwin: Alright my suspicion was correct; there was a problem when the code matches exactly, where it tries to clean up a directory that has files open. I was able to resolve this by moving the Close code in MakeIndex to the top of the method before the new directory is created. I added a comment noting that this is lucenenet specific.

The original Java code here, it should be noted, is quite sloppy with respect to these fields. It shadows them with variables of the same name in places, it relies on manually calling `close()` at the right time to clean it up, etc. But, at least this code better matches the upstream test now.

I think this is the last thing remaining on this PR, so if this looks good to you, let me know if this is approved to merge. Thanks!

paulirwin: I believe all feedback has been resolved here, so I'm going to go ahead and merge this, but feel free to let me know if there's anything else and I can do a follow-up PR. Thanks!

NightOwl888:
I setup a new repository with our Azure DevOps templates and discovered that there are some breaking changes.

### Breaking Change 1

There is a breaking change introduced by Microsoft in the .NET 8 SDK, which appends the Git commit hash to the `AssemblyInformationalVersion`: https://github.com/dotnet/sdk/issues/34568. We already add a shortened Git commit hash, so this will add it twice. The fix should definitely be in this PR, since this is where we introduce building with the .NET 8 SDK.

More info about it is here: https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/source-link. I have tested the solution and it seems to meet our needs. However, the first build after I changed it caused the NuGet packages to be built without any DLLs in them, but the second build succeeded. Not sure whether that was a fluke or a problem.

We can put the fix in the default `Directory.Build.targets` file of the solution (in each repository):

```xml
  <PropertyGroup Label="Versioning">
    <GenerateAssemblyVersionInfo>false</GenerateAssemblyVersionInfo>
  </PropertyGroup>
```

### Breaking Change 2

The default for a new pipeline in Azure DevOps is to create a shallow clone. Previously, it would create a deep clone by default.

This doesn't break anything on Lucene.Net, but all of the other repositories that we own use NerdBank.GitVersioning, and this is a problem for generating the correct version number. There is an issue about it here: https://github.com/dotnet/Nerdbank.GitVersioning/issues/837 and the fix is here: https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/doc/cloudbuild.md#azure-pipelines

We would need to apply that on the "Build" job of all of our repositories. On Lucene.NET, it is already done this way for performance reasons.

### Our Other Repositories

This is a list of all of the repositories that use the same Azure DevOps (and command line build) templates that will need to be patched to account for this.

- [J2N](https://github.com/NightOwl888/J2N)
- [ICU4N](https://github.com/NightOwl888/ICU4N)
- [RandomizedTesting](https://github.com/NightOwl888/RandomizedTesting)
- [Spatial4n](https://github.com/Spatial4n/Spatial4n)
- [Morfologik.Stemming](https://github.com/NightOwl888/Morfologik.Stemming)
- [NetFx.Polyfills](https://github.com/NightOwl888/NetFx.Polyfills) (fixes already applied)

While we could just fix the issue in this PR and add 2 issues to each of the other repositories, I think it might be simpler just to submit a PR to each repository with the patches and a link back to this PR.

paulirwin:
@NightOwl888 Thanks for the info and research! I think your comment was intended for PR #930 though?

NightOwl888:
> @NightOwl888 Thanks for the info and research! I think your comment was intended for PR #930 though?

@paulirwin - You are absolutely right. I goofed. I will re-post over there. Also, I realized I missed one.

PR #940, opened 2024-05-23
BUG: Fix for DocumentsWriter concurrency (fixes #935, closes #886)

### LINKED ISSUE #886: Support for Ordering of Indexing with SeqNo

`IndexWriter `can determine the order of concurrent index operations but does not provide this information to the user.

Currently, `IW `can determine the order in which operations were executed by updating the methods to return a long instead of `void`. If you are a caller who does not require this information, you can ignore the returned `long`. Additionally, this change enables the removal of the `TrackingIndexWriter `wrapper class with similar functionality by returning a `long` for each operation but with weaker guarantees.

### Additional context

For example, the below basic test returns the indexing in ordering with incremental `seqNo`
```cs
        [Test]
        public void TestBasic()
        {
            Directory dir = NewDirectory();
            IndexWriter w = new IndexWriter(dir, NewIndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random)));
            long a = w.AddDocument(new Document());
            long b = w.AddDocument(new Document());
            assertTrue(b > a);
            w.Dispose();
            dir.Dispose();
        }

### LINKED ISSUE #935: Poor multi-threaded indexing performance

### Describe the bug

I wrote a multithreaded indexer in Java for Lucene 4.8 and one for Lucene.net 4.8. I tried to do pretty much a line-for-line port. However, there is a substantial difference in the performance characteristics of the two. In Java, I used a ThreadPoolExecutor with CallerRunsPolicy, and in C# I used Parallel.ForEach.

1. The Java program scales more or less linearly to the number of CPUs.
2. The C# program doesn't seem to fundamentally speed up when increasing the number of threads. 
3. The Java program completes in about 8 minutes with 4 threads, the Lucene.net version takes about 65 minutes with 4, 10 or even 20 threads. 
4. It was very easy to max out CPU utilization in Java and pretty much impossible to do so in C#. 

I can provide both Java and .NET files, but before doing so, I wanted to understand if this is expected behavior and therefore not a bug. 

I am observing identical behavior in .NET 8 as well as .NET 6, so I don't think its related to https://github.com/apache/lucenenet/issues/933

### Expected Behavior

Multi-threaded indexing performance scales with the number of threads used.

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### Lucene.NET Version

4.8

### .NET Version

8.0.204 and 6.0.29

### Operating System

Windows 10

### Anything else?

_No response_

### PR DESCRIPTION
**Pull request by NightOwl888**

Lucene.Net.Support.Threading.ReentrantLock: Fixed the implementation so it prioritizes new threads obtaining the lock over waiting threads.

Fixes #935. Closes #886.

## Description

This has been a known issue for some time, however as we have had the `DocumentsWriter` working reliably on a single thread most users have not worried about it until it was reported in #935.

There were 2 issues that were causing test failures explained in https://github.com/apache/lucenenet/issues/935#issuecomment-2118228819. The second issue turned out to be much more involved to work out how to address even though the most recent solution is actually very simple and lightweight. Instead of calling `UninterruptableMonitor.Enter()` in the `Lock()` method, we call `UninterruptableMonitor.TryEnter()` in a loop that executes `Thread.Yield()`. This allows other threads to acquire the lock even though there are waiting threads.

Granted, while this approach seems to reliably pass the tests, it may be a bit naïve of an implementation. While it doesn't peg my CPU and seems to run well on Azure DevOps, I am not sure whether this is the appropriate solution for production scenarios. Suggestions are welcome as to how to improve this. Do note there were 2 prior attempts:

- https://github.com/apache/lucenenet/commit/6f2e1290ac067299c6e9b216649a852ade0a07f8 - Uses a timeout to ensure the lock is acquired, but it still has to wait until the queue schedules it until it runs.
- https://github.com/apache/lucenenet/commit/89b01e625116afd5ccba9457271a36524dc75ca1 - Uses `ManualResetEventSlim` and a `Queue<T>` to manage the waiting threads. This is a more complete implementation and even passed many of the Apache Harmony tests, but it comes at a pretty steep performance cost. Maybe there is a way to improve this, though.

### CODE REVIEW

File: src/Lucene.Net.Tests/Support/Threading/ReentrantLockTest.cs [hunk 1]
```diff
@@ -0,0 +1,1647 @@
...
+        {
+            ReentrantLock rl = new ReentrantLock();
+
+            try
+            {
+                rl.Unlock();
+                shouldThrow();
+            }
+#pragma warning disable CS0168 // Variable is declared but never used
+            catch (SynchronizationLockException success) { }
```
paulirwin, line 170: It might be cleaner to leave off the pragmas and remove the named catch variable. If you want to indicate that it's success a comment could work, like `catch (SynchronizationLockException /* success */) { }`

NightOwl888: Yep. And SonarCloud also warns against having an empty set of brackets, so I just moved it as a comment there instead.

pc-LMS:
@jeme @rclabo Hello, I am reaching out to see if you have some time to review the request.  I am hoping to begin using this fix if all goes well :)

rclabo:
I'll try to free up some time so I can look through this towards the end of the week.  I'm sure I'd learn a lot in the process, so looking forward to it.

rclabo:
This week, I made good headway toward setting up an environment where I can benchmark the current master vs this PR to see if the proposed improvements do, in fact, speed up indexing through support for DocumentsWriter concurrency.  

For benchmarking I'm using an approach similar to the code provided in [#935](https://github.com/apache/lucenenet/issues/935) to see the level of impact this PR has on the performance issue reported there.  

I will work on this more this next week and provide the results of the benchmarks once I have them.  Then I will review the code in light of those benchmarks.

That's my update for this week.

src/Lucene.Net.Tests/Support/Threading/JSR166TestCase.cs
rclabo, line 1: Many method names in this file are camelCase. They should be changed to PascalCase as is the norm in the .NET world.

rclabo:
I have reviewed each of the individual commits for this PR, but since many commits overlap and change the same areas of code I still want to review the aggregate effect on each changed file by comparing it to the Java Lucene 4.8.1 file.  I plan to do that on Monday.

I have also run performance tests on this PR vs the Master prior to the PR and the performance appears to be approximately the same.

src/Lucene.Net/Index/DocumentsWriterFlushControl.cs
rclabo, line 1: Line 505 contains "flushQueue.Count > 0 &&", which is not found in the corresponding Java Lucene 4.8.1 code. This, however, is a pre-existing deviation rather than a change submitted in this PR. Still, it may be worth investigating why this change was introduced and whether it is still needed.

src/Lucene.Net/Support/Threading/ReentrantLock.cs
rclabo, line 1: These changes to ReentrantLock overall seem like an elegant solution to get close to parity to how the reentrantLock class works in Java (as far as I understand it based on a bunch of googling). So, in general, this feels like a great step forward. It is, however, imperative that the code be original and not incorporate any OpenJDK code or code from any non-Apache source. Please ensure that is the case.

src/Lucene.Net/Support/Threading/UninterruptableMonitor.cs
rclabo, line 1: These method should ideally all contain doc comments to explain what they do.

rclabo (changes requested):

rclabo:
@pc-LMS
>We also ran a test and did not see any performance improvements - CPU remained at the same level and was not elevated in anyway. 

Today I made a private branch off of of this PR and I hacked in some concurrent logging to memory deep inside the indexing code at the dwpt level.  Specifically I logged to memory each time a thread entered the `documentWriterPerThread.UpdateDocument(...)` method and each time a thread left that method.  Through this code I was able to unequivocally verify that with this PR multiple threads are able to call `documentWriterPerThread.UpdateDocument(...)`  in parallel.  

In my observations during performance testing, I could never get my processor above 45% utilization, and then only for brief spikes, but that kinda makes sense to me given the fact that indexing is pretty clearly an IO bound operation not a CPU bound operation.

Just for fun I created a parallel threading test where I had up to 20 threads reading text files in one format and writing them in another format via an input and output stream per thread.  My local processor has 20 cores.  This 2nd setup doesn't use Lucene.NET in any way but I wanted to see if outside of Lucene.NET I could get high utilization of my CPU  when using lots of threads doing heavy IO work.  And the answer, as I suspected, is no.  My processor never got even above 35% utilization and that just for brief spikes.

src/Lucene.Net/Support/Threading/UninterruptableMonitor.cs
NightOwl888: I added some docs and eliminated the recursive methods that could overflow the stack (in theory).

src/Lucene.Net.Tests/Support/Threading/ReentrantLockTest.cs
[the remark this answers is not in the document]
rclabo: Thanks for the clarification an citing the source.  Great.

src/Lucene.Net/Support/Threading/ReentrantLock.cs
rclabo: Sweet.

src/Lucene.Net.Tests/Support/Threading/JSR166TestCase.cs
rclabo: > Typically, we haven't been strictly converting method names in test classes except for test method names

If that's the case then no worries.  I was just calling out the case difference since I was unaware of that history.  The casing is fine as submitted then.

rclabo (approved):
I appreciate your work on this @NightOwl888.  This improved locking implementation tracks MUCH more closely with the original Java implementation.  Super cool to see.

Issue #947, opened 2024-06-05
SegmentInfos replace doesn't update userData


tohidemyname:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Found that the replace method doesn't set userData with the new user data from other. 

### Expected Behavior

See my pull request

### Steps To Reproduce

See my pull request

paulirwin:
Fixed in #948 (did not auto-close due to formatting)

PR #948, opened 2024-06-05
Fix SegmentInfos replace doesn't update userData

tohidemyname:

Fixes #{bug number} (in this specific format)
https://github.com/apache/lucenenet/issues/947
## Description

{Detail}

### DIFF
--- a/src/Lucene.Net.Tests/Index/TestIndexWriter.cs
+++ b/src/Lucene.Net.Tests/Index/TestIndexWriter.cs
@@ -816,66 +816,6 @@ namespace Lucene.Net.Index
             }
         }
 
-        private Dictionary<String, String> GetCommitData(IndexWriter writer)
-        {
-            Dictionary<String, String> data = new Dictionary<String, String>();
-            foreach (var ent in writer.CommitData)
-            {
-                data.Put(ent.Key, ent.Value);
-            }
-            return data;
-        }
-
-        // LUCENENET-specific: backport fix and test from Lucene 9.9.0 (lucene#12626, lucene#12637)
-        [Test]
-        public void TestGetCommitDataFromOldSnapshot()
-        {
-            Directory dir = NewDirectory();
-            IndexWriterConfig conf = NewIndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random)).SetMaxBufferedDocs(2).SetMergePolicy(NewLogMergePolicy());
-            conf.SetIndexDeletionPolicy(new SnapshotDeletionPolicy(NoDeletionPolicy.INSTANCE));
-            IndexWriter writer = new IndexWriter(dir, conf);
-            writer.SetCommitData(
-            new Dictionary<String, String>()
-            {
-                { "key", "value" },
-            });
-            assertEquals("value", GetCommitData(writer).GetValueOrDefault("key"));
-            writer.Commit();
-            // Snapshot this commit to open later
-            IndexCommit indexCommit =
-                ((SnapshotDeletionPolicy)writer.Config.IndexDeletionPolicy).Snapshot();
-            writer.Dispose();
-
-            // Modify the commit data and commit on close so the most recent commit data is different
-            conf = NewIndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random)).SetMaxBufferedDocs(2).SetMergePolicy(NewLogMergePolicy());
-            conf.SetIndexDeletionPolicy(new SnapshotDeletionPolicy(NoDeletionPolicy.INSTANCE));
-            writer = new IndexWriter(dir, conf);
-            writer.SetCommitData(
-            new Dictionary<String, String>()
-              {
-                    {"key", "value2" },
-              });
-
-            assertEquals("value2", GetCommitData(writer).GetValueOrDefault("key"));
-            writer.Dispose();
-
-            // validate that when opening writer from older snapshotted index commit, the old commit data is
-            // visible
-            conf = NewIndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random)).SetMaxBufferedDocs(2).SetMergePolicy(NewLogMergePolicy());
-            conf.SetIndexDeletionPolicy(new SnapshotDeletionPolicy(NoDeletionPolicy.INSTANCE));
-            writer =
-                new IndexWriter(
-                    dir,
-                    conf
-                        .SetOpenMode(OpenMode.APPEND)
-                        .SetIndexCommit(indexCommit));
-            assertEquals("value", GetCommitData(writer).GetValueOrDefault("key"));
-            writer.Dispose();
-
-            dir.Dispose();
-         }
-
-
         [Test]
         public virtual void TestVariableSchema()
         {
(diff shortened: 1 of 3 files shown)


paulirwin:
@tohidemyname Thanks for this PR!

@NightOwl888 This appears to be a pretty straightforward port of https://github.com/apache/lucene/pull/12626 which was in Lucene 9.9.0. Given the single-line nature of the change, I think this is probably safe to backport. Let me know if you agree. I'll tidy it up with a comment explaining the backport so we're aware for future porting efforts.

NightOwl888:
@paulirwin - Thanks for finding that PR. The fact that a link or mention of it was missing from both the issue and PR was the primary thing holding this up.

I see a few issues that should be addressed before merging it.

- [ ] 1. The production change is missing a `LUCENENET` comment indicating where the change (which diverges from 4.8.0) came from. Ideally, there would be a link back to the PR or commit where the code was ported from.
- [ ] 2. The name of the test method should be Pascal case. NUnit doesn't automatically detect test methods based on a naming convention, so there is no need to stick with that casing in .NET.
- [ ] 3. The test method and `GetCommitData()` method are not in the same order in the file as they were in Lucene. In Lucene, `TestVariableSchema()` (the following test in this PR) is on [line 593](https://github.com/Shibi-bala/lucene/blob/e98cc333e2131ca9d39f81a2da9fed5ab0d1d5bb/lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java#L593) and `TestGetCommitDataFromOldSnapshot()` is on [line 2064](https://github.com/Shibi-bala/lucene/blob/e98cc333e2131ca9d39f81a2da9fed5ab0d1d5bb/lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java#L2064).
- [ ] 4. The original test used `getLiveCommitData()`, not `getCommitData()`. It's fine if `setLiveCommitData()` doesn't exist in 4.8.0, but we should `LUCENENET` comment to make it clear why we diverged (and that it has diverged) on each of the lines that call `GetCommitData()` and `SetCommitData()`, since these will need to be changed back when we upgrade.
- [ ] 5. The original test used [`newSnapshotIndexWriterConfig(null)`](https://github.com/Shibi-bala/lucene/blob/e98cc333e2131ca9d39f81a2da9fed5ab0d1d5bb/lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java#L935-L940). We should port that method in `LuceneTestCase` rather than inlining the configuration, and add a `LUCENENET` comment indicating where the diverging method came from.
- [ ] 6. The `TestGetCommitDataFromOldSnapshot()` and `GetCommitData()` methods also don't have a `LUCENENET` comment indicating where these diverging changes came from.

paulirwin:
@NightOwl888 My thoughts exactly, thanks for writing that all out. I'll make those changes to this PR.

src/Lucene.Net.Tests/Index/TestIndexWriter.cs
```diff
@@ -2556,12 +2556,26 @@ public virtual void TestCommitWithUserDataOnly()
             dir.Dispose();
         }
 
+        // LUCENENET-specific: backport fix and test from Lucene 9.9.0 (lucene#12626, lucene#12637)
+        private Dictionary<string, string> GetLiveCommitData(IndexWriter writer)
+        {
+            Dictionary<string, string> data = new Dictionary<string, string>();
+            // LUCENENET TODO: in a post-4.8 port, this should use LiveCommitData
```
NightOwl888, line 2563: `LUCENENET TODO` is generally reserved for work to do *prior to the 4.8.0 release* (except Ron also committed a bunch of them that will apply to an upgrade just like these). We should have a different convention for future work than for things that are pending for the current release. We specifically mention in the CONTRIBUTING section that `LUCENENET TODO` are up for grabs, but we definitely don't want this work done until later.

Maybe we should make it `LUCENENET UPGRADE TODO` to ensure it doesn't come up in a search for `LUCENENET TODO` (and change all of those that Ron added as well in a separate PR). It would then be easy to do a find and replace before we begin upgrading to change them to `LUCENENET TODO`.

src/Lucene.Net.Tests/Index/TestIndexWriter.cs
```diff
@@ -2575,6 +2589,49 @@ public virtual void TestGetCommitData()
             dir.Dispose();
         }
 
+        // LUCENENET-specific: backport fix and test from Lucene 9.9.0 (lucene#12626, lucene#12637)
+        [Test]
+        public void TestGetCommitDataFromOldSnapshot()
+        {
+            Directory dir = NewDirectory();
+            IndexWriter writer = new IndexWriter(dir, NewSnapshotIndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random)));
+            // LUCENENET TODO: in a post-4.8 port, this should use SetLiveCommitData
```
NightOwl888, line 2598: See my other comment about why `LUCENENET TODO` is not appropriate here.

src/Lucene.Net.Tests/Index/TestIndexWriter.cs
```diff
@@ -2575,6 +2589,49 @@ public virtual void TestGetCommitData()
...
+            });
+            assertEquals("value", GetLiveCommitData(writer)["key"]);
+            writer.Commit();
+            // Snapshot this commit to open later
+            IndexCommit indexCommit =
+                ((SnapshotDeletionPolicy)writer.Config.IndexDeletionPolicy).Snapshot();
+            writer.Dispose();
+
+            // Modify the commit data and commit on close so the most recent commit data is different
+            writer = new IndexWriter(dir, NewSnapshotIndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random)));
```
NightOwl888, line 2611: In the upstream code the analyzer was `null` rather than `new MockAnalyzer(Random)`. Looking at the `LiveIndexWriterConfig`, it should accept that, although I am not sure about the code that uses `IndexWriterConfig`. Did you try `null`, and if so, what happened?

If it doesn't work, then maybe we should patch this inside of `NewSnapshotIndexWriterConfig` for the time being, since that will come up as a new method that we need to port, anyway.

Either way, this diverges from upstream and we will need a `LUCENENET UPGRADE TODO` comment (unless `null` works).

src/Lucene.Net.Tests/Index/TestIndexWriter.cs
```diff
@@ -2575,6 +2589,49 @@ public virtual void TestGetCommitData()
...
+            // LUCENENET TODO: in a post-4.8 port, this should use SetLiveCommitData
```
NightOwl888, line 2612: See my other comment about why `LUCENENET TODO` is not appropriate here.

src/Lucene.Net.Tests/Index/TestIndexWriter.cs
```diff
@@ -2575,6 +2589,49 @@ public virtual void TestGetCommitData()
...
+
+            assertEquals("value2", GetLiveCommitData(writer)["key"]);
+            writer.Dispose();
+
+            // validate that when opening writer from older snapshotted index commit, the old commit data is
+            // visible
+            writer =
+                new IndexWriter(
+                    dir,
+                    NewSnapshotIndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random))
```
NightOwl888, line 2626: See my other comment about `null` vs `new MockAnalyzer(Random)`.

NightOwl888 (changes requested):

src/Lucene.Net.Tests/Index/TestIndexWriter.cs
paulirwin: Tests pass with null, so made that change.

NightOwl888 (approved):
Thanks @paulirwin and @tohidemyname, this looks good to merge.

PR #952, opened 2024-07-24
Leave debugStream as null on construction of parser token managers

### LINKED ISSUE #936: Lucene.Net.Util.SystemConsole throws not supported exception in .NET MAUI app running on android/iOS

### Describe the bug

I am simply trying to create a StandardQueryParser. I get the below exception only on Android and iOS. This makes sense as `System.Console.In` is not supported on these platforms, but I have no need for it in the context of our application. So, I think a way to ignore this would be very useful to prevent our app from crashing on these platforms.

Note that I am testing on Android and iOS in Android Studio and XCode respectively, not on physical devices. Though, I don't think this will make a difference.

It seems the issue is the way these properties are setup in `Lucene.Net.Util.SystemConsole`.

```
public static TextWriter Out { get; set; } = Console.Out;
public static TextWriter Error { get; set; } = Console.Error;
public static TextReader In { get; set; } = Console.In;
```

My application would work fine to simply return null here when Console.In is not supported.

I can't seem to find anything online about this regarding Lucene, so I think it is just specific to Lucene.NET and .NET MAUI. In any case, I believe it can possibly be addressed simply by doing the above.

From what I can tell, `SystemConsole` is only used in `StandardSyntaxParserTokenManager` to create the private `debugStream` property, which is then never used in this class. I believe this whole thing is the same when using the default QueryParser as well.

### Expected Behavior

Perhaps instead of defaulting to `Console.In`, the property can be set in a way that if a non-supported or any other exceptions that can be ignored is caught, this property returns null or a bogus TextReader by default.

### Steps To Reproduce

* In a .NET MAUI sample app, create a `Lucene.Net.QueryParsers.Flexible.Standard.StandardQueryParser` using the constructor.
    * Specifically we created our MAUI app using the default template in Rider.
* Run the app.
on Windows and macOS, this will run fine. On Android and iOS, the app will crash with a `System.PlatformNotSupportedException` when it attempts to construct the `StandardQueryParser`.

### Exceptions (if any)

```
Exception: System.TypeInitializationException: The type initializer for 'Lucene.Net.Util.SystemConsole' threw an exception.
 ---> System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.ConsolePal.GetOrCreateReader()
   at System.Console.<get_In>g__EnsureInitialized|14_0()
   at System.Console.get_In()
2024-04-29 15:55:32.051142-0700  at Lucene.Net.Util.SystemConsole..cctor()
   --- End of inner exception stack trace ---
   at Lucene.Net.QueryParsers.Flexible.Standard.Parser.StandardSyntaxParserTokenManager..ctor(ICharStream stream)
2024-04-29 15:55:32.051200-0700  at Lucene.Net.QueryParsers.Flexible.Standard.Parser.StandardSyntaxParser..ctor(ICharStream stream)
   at Lucene.Net.QueryParsers.Flexible.Standard.Parser.StandardSyntaxParser..ctor()
   at Lucene.Net.QueryParsers.Flexible.Standard.StandardQueryParser..ctor()
2024-04-29 15:55:32.051439-0700 at Lucene.Net.QueryParsers.Flexible.Standard.StandardQueryParser..ctor(Analyzer analyzer)
```

### Lucene.NET Version

4.8

### .NET Version

8.0.203

### Operating System

Windows 11

### Anything else?

_No response_

JayOfemi:

Leave debugStream as null on construction of parser token managers.

Fixes #{bug number} (in this specific format)
Fixes #936

## Description

{Detail}

* When `debugStream` is set to `Console.Out` by default, `StandardSyntaxParserTokenManager` throws an exception when constructing a `StandardQueryParser` on OSes that do not support System.Console, such as iOS and Android.
* `debugStream` can be set later using the setter, if needed.
* The above is the same for `QueryParserTokenManager` when constructing a `QueryParser`.
* See issue https://github.com/apache/lucenenet/issues/936 for details.

### DIFF
--- a/src/Lucene.Net.QueryParser/Classic/QueryParserTokenManager.cs
+++ b/src/Lucene.Net.QueryParser/Classic/QueryParserTokenManager.cs
@@ -1,4 +1,4 @@
-﻿using Lucene.Net.Support.IO;
+using Lucene.Net.Support.IO;
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
@@ -31,7 +31,7 @@ namespace Lucene.Net.QueryParsers.Classic
     {
         /// <summary>Debug output. </summary>
 #pragma warning disable IDE0052 // Remove unread private members
-        private TextWriter debugStream = Console.Out; // LUCENENET specific - made private, since we already have a setter
+        private TextWriter debugStream; // LUCENENET specific - made private, since we already have a setter
 #pragma warning restore IDE0052 // Remove unread private members
         /// <summary>Set debug output. </summary>
         public virtual void SetDebugStream(TextWriter ds)
--- a/src/Lucene.Net.QueryParser/Flexible/Standard/Parser/StandardSyntaxParserTokenManager.cs
+++ b/src/Lucene.Net.QueryParser/Flexible/Standard/Parser/StandardSyntaxParserTokenManager.cs
@@ -1,4 +1,4 @@
-﻿using Lucene.Net.Support.IO;
+using Lucene.Net.Support.IO;
 using System.Diagnostics.CodeAnalysis;
 using System;
 using System.IO;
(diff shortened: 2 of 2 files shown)


paulirwin:
@JayOfemi Thanks for this PR. There are several more places where SystemConsole is used, so this change alone is likely insufficient to address the problem and cover all cases. Also, we'd like to use `TextWriter.Null` instead of actual `null` if possible. I will make that change and figure out an approach that will work well for MAUI without regressing other platforms. I'll make sure to include your commit in the history though so you get credit. No need to do anything at this time. Thanks!

paulirwin:
This was changed to use `TextWriter.Null` by default which should resolve the issue.

PR #956, opened 2024-08-13
Locking performance optimizations for .NET 9.0+

### LINKED ISSUE #955: Make use of .NET 9.0's System.Threading.Lock which is more performant

### Task description

.NET 9.0's System.Threading.Lock provides a method for locking, instead of locking on an object, that is about 25% more performant.

MarkCiliaVincenti:

Summary of the changes
Targeting of .NET 9.0 and locking performance optimizations for .NET 9.0+

Fixes #955

## Description

Targeting of .NET 9.0 and locking performance optimizations for .NET 9.0+. Backported to all prior versions of .NET without performance degradation.


MarkCiliaVincenti:
Would need to be rewritten to harden against thread aborts. Let me know if interested.

MarkCiliaVincenti:
> Related [Reddit post](https://www.reddit.com/r/csharp/comments/1f6ari2/locking_with_net_90s_systemthreadinglock_even_on/) by @MarkCiliaVincenti which provides a nice background summary.
> 
> While in the longer term, this is, of course, very interesting, I can't see it being much of a priority given that there is no performance gain unless running on .NET 9. Just my thoughts.

That is correct. The performance gain is only on .NET 9 or greater, but there is no performance loss for .NET 8 or earlier.

More and more people will start adopting .NET 9, and later .NET 10, at which point the assumption is that .NET 8 would still be targeted by Lucene.NET. Therefore it stands to reason that one should ask whether or not it should be done. If the answer is yes, delaying it provides no value, except sensible ones such as waiting for .NET 9.0 to be officially released.

paulirwin:
@MarkCiliaVincenti Thank you for this PR, but there are a few reasons why we are going to close this at this time. First, as you might have noticed, it has been quite some time since our last release. At this time, to prevent having to continually shift our targets, we are not planning on targeting .NET 9 for this next release. (Additionally, if we were going to consider that, we would want that change done separate from locking changes.) We also have just gone through fixing a number of concurrency bugs for this release, and are hesitant to make any changes at this time. We will keep the issue open and reconsider this for a future release. Thanks again!

MarkCiliaVincenti:
@paulirwin meanwhile a new version of Backport.System.Threading.Lock has come out that acts as a source generator and basically dropping the DLL as a dependency.

paulirwin:
That is good to know about, thank you. At this time, however, we are still not planning on adding a .NET 9 target. Each new target requires quite a bit of effort and resources to support, and we haven't even launched with a LTS version yet. We can't have this be a moving target (literally).

We need the community's help on finishing up the 4.8 release items first, which is all the difficult work of fixing failing unit tests, doing API reviews to ensure compatibility, tracking down performance issues demonstrated with benchmarks, things like UTF-8 byte-order marks, etc. It's not the easy/fun stuff like this, but it's important work. If you'd like to contribute there, please check out the 4.8.0-beta00018 milestone in issues, and comment if you'd like to take one of the items so we can follow up on it. Thanks!

MarkCiliaVincenti:
OK @paulirwin, I'd appreciate if you'd let me know once .NET 9.0 is officially supported, so that we can revisit this.

PR #957, opened 2024-08-16
feat: #954 - TryDequeue and TryPeek Queue extension methods

### LINKED ISSUE #954: Add extension methods to patch Queue.TryDequeue() and Queue.TryPeek() on netstandard2.0 and .NET Framework

### Task description

Per https://github.com/apache/lucenenet/pull/940#discussion_r1643205431, it would make the code easier to understand if we didn't have to manually check `Queue.Count` every time we want to call `Dequeue()` or `Peek()`. Newer versions of .NET have the methods:

- [TryDequeue()](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.queue-1.trydequeue?view=net-8.0#system-collections-generic-queue-1-trydequeue(-0@))
- [TryPeek()](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.queue-1.trypeek?view=net-8.0#system-collections-generic-queue-1-trypeek(-0@))

but they are missing from `netstandard2.0` and .NET Framework. However, if we add them as extension methods, we can both solve the readability issue and also make the call a bit more performant in .NET Core.

## Specifics

- File Name: `QueueExtensions.cs`
- Location: `src/Lucene.Net/Support/`
- Namespace: `Lucene.Net.Support`
- Class Name: `QueueExtensions`
- Accessibility: `static internal`
- Documentation for class: `Extensions to <see cref="Queue"/>.`
- Method implementations: Search the codebase for `.Dequeue()` to see an example. We just want to wrap the check for `.Count` inside of the methods.
- Both methods should be marked `[MethodImpl(MethodImplOptions.AggressiveInlining)]`.
- Both methods should be fully documented - copy the documentation from the above links.
- Both extension methods should check for a `null` queue and if the check fails `throw new ArgumentNullException(nameof(queue))`.
- The methods should have basic tests in a corresponding file `src/Lucene.Net.Tests/Support/QueueExtensionsTests.cs`.
- Add a new feature named `FEATURE_QUEUE_TRYDEQUEUE_TRYPEEK` and set it up in the root `Directory.Build.targets` file to be enabled in `netstandard2.1` and all versions of .NET Core. We should conditionally compile out these extension methods in those cases by putting them inside `#if !FEATURE_QUEUE_TRYDEQUEUE_TRYPEEK` / `#endif` blocks.

In a separate Git commit, search the codebase for usage of `Queue` and replace all calls with the extension methods. We don't need any conditional compilation here because we will use the same method signatures.

devklick:

Add TryDequeue and TryPeek Queue extension methods

Fixes #954

## Description

New class created, `QueueExtensions`, with two new methods, `TryDequeue` and `TryPeek`. These are compiled based on the `FEATURE_QUEUE_TRYDEQUEUE_TRYPEEK` flag. 
- Inline documentation included
- Unit tests included 

**Notes:**

One of the requirements on the linked ticket was to update all references to use these new functions. I have updated _some_ of them, only where it will eliminate a check along the lines of `queue.count > 0 && queue.Dequeue/Peek()`. There are more calls to `queue.Dequeue()` and `queue.Peek()` where it did not seem suitable to replace with the Try* equivalent, so I have left them as-is.

The `QueueExtensionsTests` suite of tests will be testing the actual methods on the Queue object when testing on any framework other than netstandard2.0 and .NET Framework, rather than testing our extensions. Should these tests be configured to only run for certain frameworks? Seems redundant to have unit tests targeting a built-in feature.


NightOwl888 (changes requested):
Thanks for the PR and your willingness to help make our project better.

There are a few changes we would like you to make before we can accept this PR. First, please see the comments inline.

src/Lucene.Net/Support/QueueExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+#if !FEATURE_QUEUE_TRYDEQUEUE_TRYPEEK
+        /// <summary>
+        /// Removes the object at the beginning of the <see cref="Queue{T}"/>,
+        /// and copies it to the <paramref name="result"/> parameter.
+        /// </summary>
+        /// <typeparam name="T">The type of element in the <see cref="Queue{T}"/></typeparam>
+        /// <param name="queue">The <see cref="Queue{T}"/> to be checked</param>
+        /// <param name="result">The removed object</param>
+        /// <returns><c>true</c> if the object was successfully removed; <c>false</c> if the <see cref="Queue{T}"/> is empty.</returns>
+        /// <exception cref="ArgumentNullException"></exception>
```
NightOwl888, line 38: Please add "<paramref name="queue"/> is <c>null</>." within this element both here and in TryPeek().

devklick: Hi @NightOwl888, can you please clarify, do you mean the following?
```xml
<exception cref="ArgumentNullException">is null</exception>
```

src/Lucene.Net.Tests/Support/QueueExtensionsTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+    public class QueueExtensionsTests : LuceneTestCase
+    {
+#if !FEATURE_QUEUE_TRYDEQUEUE_TRYPEEK
+        [Test, LuceneNetSpecific]
+        public void TryDequeue_ThrowsWhenQueueNull()
+        {
+            Queue<int> queue = null;
+            var ex = Assert.Throws<ArgumentNullException>(() => queue.TryDequeue(out int _));
+            Assert.That(ex.ParamName, Is.EqualTo("queue"));
```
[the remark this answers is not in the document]
devklick: Apologies, I should have checked the existing tests to see what to use. I've updated to use Lucene Assert now. 

Note that since the Lucene equivalent doesnt cast the returned exception to `T` (which I guess might be for performance reasons), I'm casting it to `ArgumentNullException` within two of these tests to verify the `ParamName`. Please let me know if I should avoid doing do. I notice that the existing tests for `Assert.Throws<ArgumentNullException>` dont bother checking the `ParamName`, so maybe this is overkill?

src/Lucene.Net.Analysis.Phonetic/DoubleMetaphoneFilter.cs
```diff
@@ -52,12 +53,11 @@ public DoubleMetaphoneFilter(TokenStream input, int maxCodeLength, bool inject)
 
         public override bool IncrementToken()
         {
-            for (;;)
+            for (; ; )
```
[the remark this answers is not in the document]
devklick: Apologies - automatic editor linting. I've removed those spaces.

src/Lucene.Net/Support/QueueExtensions.cs
NightOwl888: My apologies. GitHub stripped the angle brackets.

```xml
<exception cref="ArgumentNullException"><paramref name="queue"/> is <c>null</>.</exception>
```

src/Lucene.Net.Tests/Support/QueueExtensionsTests.cs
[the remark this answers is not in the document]
NightOwl888: Yeah, usually just asserting the correct error type is enough, although some tests check error messages, as well.

src/Lucene.Net/Support/QueueExtensions.cs
```diff
@@ -0,0 +1,79 @@
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
```
NightOwl888, line 4: Please add the line `#nullable enable` here.

src/Lucene.Net/Support/QueueExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <summary>
+        /// Removes the object at the beginning of the <see cref="Queue{T}"/>,
+        /// and copies it to the <paramref name="result"/> parameter.
+        /// </summary>
+        /// <typeparam name="T">The type of element in the <see cref="Queue{T}"/></typeparam>
+        /// <param name="queue">The <see cref="Queue{T}"/> to be checked</param>
+        /// <param name="result">The removed object</param>
+        /// <returns><c>true</c> if the object was successfully removed; <c>false</c> if the <see cref="Queue{T}"/> is empty.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="queue"/> is <c>null</c>.</exception>
+        public static bool TryDequeue<T>(this Queue<T> queue, out T result)
```
NightOwl888, line 39: Please add the `[MaybeNullWhen(false)]` attribute before the `out` modifier.

src/Lucene.Net/Support/QueueExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// <summary>
+        /// Returns a value that indicates whether there is an object at the beginning of the <see cref="Queue{T}"/>,
+        /// and if one is present, copies it to the <paramref name="result"/> parameter. The object is not removed from the <see cref="Queue{T}"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of element in the <see cref="Queue{T}"/></typeparam>
+        /// <param name="queue">The <see cref="Queue{T}"/> to be checked</param>
+        /// <param name="result">If present, the object at the beginning of the <see cref="Queue{T}"/>; otherwise, the default value of <typeparamref name="T"/></param>
+        /// <returns><c>true</c> if there is an object at the beginning of the <see cref="Queue{T}"/>; <c>false</c> if the <see cref="Queue{T}"/> is empty.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="queue"/> is <c>null</c>.</exception>
+        public static bool TryPeek<T>(this Queue<T> queue, out T result)
```
NightOwl888, line 63: Please add the `[MaybeNullWhen(false)]` attribute before the `out` modifier.

NightOwl888 (changes requested):
I apologize for adding additional requirements, but could you please add nullable reference type support? See my inline comments.

src/Lucene.Net/Support/QueueExtensions.cs
devklick: Good thinking, a worthwhile addition for sure.

NightOwl888 (approved):

NightOwl888:
Thanks again for your contribution. Your help is appreciated.

Issue #911, opened 2024-02-03
Docs: DocFx Build Failure for API Docs


NightOwl888:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

The API documentation is failing to build. The website is generated with no URLs being resolved.

### Expected Behavior

The documentation will build and all links will be resolvable.

### Steps To Reproduce

1. Clone the repository locally
2. Run the powershell command `./websites/apidocs/docs.ps1 -ServeDocs -LuceneNetVersion 4.8.0-beta00016 -BaseUrl http://localhost:8080`

See the [API Docs build instructions](https://lucenenet.apache.org/contributing/documentation.html#api-docs).

### Exceptions (if any)

```console
Building api metadata for D:\a\lucenenet\lucenenet\main-repo\websites\apidocs\docfx.codecs.json...
[24-02-01 10:39:03.573]Warning:[ExtractMetadata](D:/a/lucenenet/lucenenet/main-repo/src/Lucene.Net.Codecs/Lucene.Net.Codecs.csproj)Workspace failed with: [Failure] Msbuild failed when processing the file 'D:\a\lucenenet\lucenenet\main-repo\src\Lucene.Net.Codecs\Lucene.Net.Codecs.csproj' with message: Method not found: 'System.ReadOnlySpan`1<Char> Microsoft.IO.Path.GetFileName(System.ReadOnlySpan`1<Char>)'.
[24-02-01 10:39:04.042]Warning:[ExtractMetadata]Project 'D:\a\lucenenet\lucenenet\main-repo\src\Lucene.Net.Codecs\Lucene.Net.Codecs.csproj' does not contain any documents.
[24-02-01 10:39:04.417]Warning:[ExtractMetadata]No metadata is generated for Lucene.Net.Codecs.
```

### Lucene.NET Version

Lucene.NET 4.8.0-beta00016

### Operating System

Windows 10 x64

### Anything else?

I did a search and discovered the following issue that appears to be the same problem:

https://github.com/dotnet/docfx/issues/8097

The solution was to [upgrade to 2.59.4](https://github.com/dotnet/docfx/issues/8097#issuecomment-1519557984). I attempted that next, and then got hit with the error that `System.Memory` had multiple versions referenced and could not load it similar to [this comment](https://github.com/dotnet/docfx/issues/7832).

[...]

Issue #793, opened 2023-03-24
When will the 4.8.0 version be released？


zhanghaixingxing:

rclabo:
This question was asked a few months ago and the answer hasn't likely changed much.  Please see [#778](https://github.com/apache/lucenenet/issues/778).  In that thread you will see references to [#437 Comment](https://github.com/apache/lucenenet/issues/437#issuecomment-992014580) which contains a wealth of information on the topic.

In all of this, one thing you should keep in mind is that many people (even Microsoft!) currently use portions of Lucene 4.8 in production.  So that is to say the product is already very stable.  

Companies using 4.8 in production just need to be awere there may be some minor API changes on the road to a final 4.8 release.  But honestly, that's not a lot different than using a Production release that has a major new release.

jeffreywstevens:
In my opinion, if you know people are using it, and you feel it is stable, you might as well call it a release and remove the beta designation. 

I am confident it works for most use cases. However, I can't get some developers to consider it. If it gets escalated, it will most likely get dropped.

I feel it is time to go forward.

Shazwazza:
I'm definitely in agreement with @jeffreywstevens here. I know there are some recent commits and changes done in the past several months which would warrant a new beta release, but after that I also think we should procure an RTM release. Any changes after that can just be patch versions. 

Would need buy in from @rclabo + @NightOwl888

rclabo:
I'm totally on board with that.  In the past two years I've had to make way more changes to my code base due to ASP.NET Core API changes than I have had to for Lucene.NET 4.8 API changes.  I feel that LuceneNET 4.8 is super stable (hats off to @NightOwl888!!!) and is worthy of an RTM release.  

Sure there are some aspects that may not be perfectly on par with Java Lucene 4.8 (OpenNLP comes to mind), but those areas tend to be auxiliary functionality that have no easy route for porting. 

The core functionality seems rock solid and awesome from my perspective.  And I think so many more devs will use this amazing project if it's RTM.

rclabo:
I know @NightOwl888 is under a lot of pressure right now due to a deadline on another project, so he might not be able to chime in for a while.  

That doesn't mean he isn't interested in this discussion; it just means he is juggling a lot at the moment.

Shazwazza:
@rclabo IIRC there were quite a lot of commits and fixes since the last release, do you think we should look to ship one more beta?

rclabo:
[...]

PR #958, opened 2024-09-13
[DRAFT] Partial docfx fix - Missing some plugins

nikcio:
This is a leg more on the fix @paulirwin have been trying out for solving #911 

This is based on the commits @paulirwin made here: https://github.com/paulirwin/lucene.net/tree/issue/911

Use this to finish off the what is left to have docfx fully working.

## What did I add?

1. Replaced `toc.yml`
As mentioned [here](https://github.com/apache/lucenenet/issues/793#issuecomment-2349070692) there was a problem with the `toc.yml` being locked this was fixed by changing the path to `"./toc.yml"` instead of `toc.yml"`.

luc-1

2. Added small delay when building site output.

While the first fix solved the problem I noticed in one case when running locally that the file got locked anyway therefore I added a 1 second delay between starting new site out builds and haven't seen the error since.

3. Fixed the links to the assets (`js` and `css`)

When starting the docs site locally the pages weren't showing the CSS or using the js files because the route was wrong due to the missing `_rel` variable, therefore, I've added a slash `/` before all uses of these files so it works locally and removed the ending slash from the `docfx.global.subsite.json`.

`docfx.vendor.css` was also changed to `docfx.vendor`.min.css`.
`docfx.vendor.js` was also changed to `docfx.vendor.min.js`

4. Added an environment variable replacement postprocessor

I'm not sure if it works in all cases that it should but I've created a `EnvironmentVariableProcessor` that makes sure to replace `EnvVar:ENVIRONMENT_VARIABLE` to the actual value of the environment variable. As far as I could tell this is only used on the index page but I have no idea if this is correct. I haven't replaced the other plugin modules in these changes so that is still missing.

image

## Next step

@paulirwin or @Shazwazza feel free to take these changes and finish them off 😄 

I think the only thing missing is the remaining commented out plugin files otherwise it looks to be working.

## How I've been testing

I've used the following command to test it.

```sh
./websites/apidocs/docs.ps1 -ServeDocs -LuceneNetVersion 4.8.0-beta00016 -BaseUrl http://localhost:8080
```

### DIFF
--- a/src/docs/.idea/.idea.DocumentationTools/.idea/.gitignore
+++ b/src/docs/.idea/.idea.DocumentationTools/.idea/.gitignore
@@ -1,0 +1,13 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Rider ignored files
+/contentModel.xml
+/projectSettingsUpdater.xml
+/.idea.DocumentationTools.iml
+/modules.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
--- a/src/docs/.idea/.idea.DocumentationTools/.idea/.name
+++ b/src/docs/.idea/.idea.DocumentationTools/.idea/.name
@@ -1,0 +1,1 @@
+DocumentationTools
--- a/src/docs/.idea/.idea.DocumentationTools/.idea/encodings.xml
+++ b/src/docs/.idea/.idea.DocumentationTools/.idea/encodings.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="Encoding" addBOMForNewFiles="with BOM under Windows, with no BOM otherwise" />
+</project>
--- a/src/docs/.idea/.idea.DocumentationTools/.idea/indexLayout.xml
+++ b/src/docs/.idea/.idea.DocumentationTools/.idea/indexLayout.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="UserContentModel">
+    <attachedFolders />
+    <explicitIncludes />
+    <explicitExcludes />
+  </component>
+</project>
--- a/src/docs/.idea/.idea.DocumentationTools/.idea/material_theme_project_new.xml
+++ b/src/docs/.idea/.idea.DocumentationTools/.idea/material_theme_project_new.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="MaterialThemeProjectNewConfig">
+    <option name="metadata">
+      <MTProjectMetadataState>
+        <option name="migrated" value="true" />
+        <option name="pristineConfig" value="false" />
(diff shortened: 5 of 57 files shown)


paulirwin:
Thanks for this! I'm going to merge this into an issue branch in this repo for collaboration. I was seemingly able to fix the toc locking by setting maxParallelism to 1. Obviously not ideal for performance, but reliability > performance right now, and it's still not too slow.

paulirwin:
I have merged this with my work into this issue branch: https://github.com/apache/lucenenet/tree/issue/911

We can now use this branch to finish up collaboration on this issue. Note though that I am getting an error building your plugins changes, not sure why. However, it's great that running with `-DisablePlugins` now works for the libraries! This is a huge boost to progress, thank you. I'll be digging into this more soon.

NightOwl888:
This is great. That would explain why I saw it choke when the 3rd part JSON parser tried to read the ToC.

As for the version numbers, we ended up resorting to using Powershell to update them here: https://github.com/apache/lucenenet/blob/0d7bba7a42a89a3e348c6c27370425b4db33b492/websites/apidocs/docs.ps1#L182-L192. We tried several times to get the plugins to do that for us, but they don't seem to work in any section of the page, only in certain sections (for example, we couldn't replace the version number inside of a code block using the plugin).

paulirwin:
I think we should be able to do that with plugins now as a post-processor! IIUC, anyways.

paulirwin:
I was able to fix the plugin build error by excluding some files that were using the old docfx SDK. I went ahead and excluded the rest of those files that were commented out, no reason in keeping them in the csproj. By the time we merge this we should remove those from source control (they will of course always be available in prior version history if needed).

I'm testing this now and will review the rest of the plugin code to see what all is needed to be implemented. This is very exciting to use the latest docfx!

NightOwl888:
Hmm...the whole thing is in Powershell, except for the DocFx bit :). I suggest you have a look at the [Docs and Website Documentation](https://lucenenet.apache.org/contributing/documentation.html) and the build automation for the [docs](https://github.com/apache/lucenenet/blob/0d7bba7a42a89a3e348c6c27370425b4db33b492/.github/workflows/Lucene-Net-Documentation.yml) and the [website](https://github.com/apache/lucenenet/blob/0d7bba7a42a89a3e348c6c27370425b4db33b492/.github/workflows/Lucene-Net-Website.yml). They are set up to build the docs and then submit a PR to the [lucenenet-site repo](https://github.com/apache/lucenenet-site) (although, the trigger isn't working and it requires starting manually).

Also, if you checkout the tag for beta00016, you should be able to run the build prior to the DLL conflict to see it work. Ideally, the UX and procedure won't have to change for the upgrade.

paulirwin:
Sorry, I was meaning replacing the "HACKs" you were referring to here: https://github.com/apache/lucenenet/pull/958#issuecomment-2351946840

We at least should be able to do that for the CLI index. I'll look into it.

paulirwin:
The CLI docs version replacement was an easy change, made that use `EnvVar:LuceneNetVersion` since that works in fenced code blocks now. Removed that from the ps1 script; the rest are TOC/breadcrumbs replacements that I think should stay for now.

Next up I'll be looking into why the header links aren't working, but once that is working we should be in good shape for the API docs site.

paulirwin:
@nikcio I have included your commit in #961 so I'll close this PR. Thanks!

PR #960, opened 2024-09-20
Use concrete ConcurrentHashSet instead of abstraction for performance

Shazwazza:

When it is a concrete type, this is avoided. Since this is an internal field there's no reason to use the abstracted type. 

See https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1859


paulirwin:
@Shazwazza I agree with this change. Even though the upstream Java code uses the abstraction for the field type, we're already deviated from the upstream code by using a set instead of a list. I think this field could use a `// LUCENENET-specific:` comment on it though to note for future porting work.

And I'm guessing you saw the build failed, so once that is fixed if you wouldn't mind adding that comment I'd be happy to approve.

Shazwazza:
Though, not sure why some tests fail, that seems very strange

paulirwin:
Looks like it was just a transient error downloading the NuGet package feed, I've seen that before. Re-ran the tests and they passed. Thanks!

PR #968, opened 2024-10-08
BREAKING: Backport serialization fix for replicator

paulirwin:

Backport of a serialization fix from Lucene.

## Description

This backports a serialization fix in the upstream Lucene project to avoid arbitrary type instantiation when errors occur.

Upstream commit: https://github.com/apache/lucene/commit/b4b153f64fb0420c6e28ddea4c442e119458756c

### DIFF
--- a/src/Lucene.Net.Replicator/Http/ReplicationService.cs
+++ b/src/Lucene.Net.Replicator/Http/ReplicationService.cs
@@ -188,19 +188,9 @@ namespace Lucene.Net.Replicator.Http
                         break;
                 }
             }
-            catch (Exception e)
+            catch (Exception)
             {
                 response.StatusCode = (int)HttpStatusCode.InternalServerError; // propagate the failure
-                try
-                {
-                    TextWriter writer = new StreamWriter(response.Body);
-                    JsonSerializer serializer = JsonSerializer.Create(JSON_SERIALIZER_SETTINGS);
-                    serializer.Serialize(writer, e, e.GetType());
-                }
-                catch (Exception e2) when (e2.IsException())
-                {
-                    throw new IOException("Could not serialize", e2);
-                }
             }
             finally
             {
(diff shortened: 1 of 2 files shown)


paulirwin:
Needs one more change, will reopen

src/Lucene.Net.Replicator/Http/ReplicationService.cs
```diff
@@ -188,24 +188,14 @@ public virtual void Perform(IReplicationRequest request, IReplicationResponse re
...
                 }
             }
-            catch (Exception e)
+            catch (Exception)
             {
                 response.StatusCode = (int)HttpStatusCode.InternalServerError; // propagate the failure
-                try
-                {
-                    TextWriter writer = new StreamWriter(response.Body);
-                    JsonSerializer serializer = JsonSerializer.Create(JSON_SERIALIZER_SETTINGS);
```
NightOwl888, line 197: Looks like the only caller left to `JSON_SERIALIZER_SETTINGS` is [`ExecutePost`](https://github.com/apache/lucenenet/blob/e6bac3109ffefff840d7abfac1d403fcebf26b93/src/Lucene.Net.Replicator/Http/HttpClientBase.cs#L184). This method is marked `@internal` in Lucene and has no callers in Lucene.NET.

It seems like we should mark both of these `internal`. Lucene never had a `JSON_SERIALIZER_SETTINGS` and it is not very useful (there is no parameter that accepts it anywhere), so I think that definitely needs to be `internal`. I don't like the idea of exposing 3rd party APIs on our API, as it couples us to that specific library (and `System.Text.Json` is a thing).

The only issue with marking `ExecutePost` internal is that somebody may be using it. If we do mark it `internal`, we should add a `LUCENENET specific` comment indicating why we diverged from Lucene.

paulirwin: I removed `JSON_SERIALIZER_SETTINGS` entirely because it's not needed if we're just using the default and removing TypeNameHandling.

I think we should leave `ExecutePost` as protected for now to match Java's semantics. It's a now-deprecated type (in Lucene 9.12) anyways, and I don't see a good reason to break anyone that has already implemented this method even though it's not being used. I think we can safely leave it as-is.

PR ready for re-review.

NightOwl888: Sounds good.

NightOwl888 (approved):
Thanks for the PR.

This looks good to merge.

PR #971, opened 2024-10-12
Pin ubuntu build to 22.04 due to .NET 5 issue

### LINKED ISSUE #970: Pin Ubuntu build to 22.04 due to .NET 5 build issue

### Task description

The .NET 5 builds on Ubuntu started failing due to the runner for ubuntu-latest being updated to 24.04, which breaks a dependency .NET 5 has on OpenSSL. We can pin the builds to ubuntu-22.04 for now which will remain a supported LTS release until 2027, by which time we should drop the unsupported .NET 5 for an alternative approach, such as using a newer .NET build that is forced to target .NET Standard 2.1.

See #961 for discussion of the build failure.

paulirwin:

Pins the ubuntu builds to Ubuntu 22.04 LTS due to a .NET 5 OpenSSL dependency issue. We should find a better approach sometime before 22.04 goes out of support in 2027.

Fixes #970 

## Description

Changes the build agent from `ubuntu-latest` to `ubuntu-22.04`.


NightOwl888 (changes requested):
> We should find a better approach sometime before 22.04 goes out of support in 2027.

We will have a better approach soon because we will be able to drop the `net6.0` target and use it instead of `net5.0` for testing `netstandard2.1`.

Looks like the `azure-pipelines.yml` wasn't included, but let's do that because Azure DevOps uses the same images. For that one, we can easily limit this change to `net5.0`.

NightOwl888 (approved):
Looks good. I confirmed the linux `net5.0` tests are producing results on both Azure DevOps and GitHub Actions.

PR #972, opened 2024-10-13
Upgrade TestFramework System.Text.Json reference to 6.0.10 for net462 target

paulirwin:

Upgrades System.Text.Json to 6.0.10 for net462 due to 6.0.6 advisory: https://github.com/advisories/GHSA-8g4q-xg66-9fp4

### DIFF
--- a/.build/dependencies.props
+++ b/.build/dependencies.props
@@ -82,7 +82,7 @@
     <SystemSecurityCryptographyXmlPackageVersion>6.0.1</SystemSecurityCryptographyXmlPackageVersion>
     <SystemTextEncodingCodePagesPackageVersion>4.3.0</SystemTextEncodingCodePagesPackageVersion>
     <SystemTextEncodingCodePagesPackageVersion Condition=" '$(TargetFramework)' == 'net472' ">5.0.0</SystemTextEncodingCodePagesPackageVersion>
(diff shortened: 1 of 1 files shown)


NightOwl888 (changes requested):
Yeah, let's do this only once for the release instead of upgrading dependencies one at a time and verifying compatibility each time. Since we will be bringing in new dependencies that may have transitive conflicts, we should wait until after J2N and ICU4N are integrated before upgrading the dependencies.

Also, it is important to test dependency changes on Azure DevOps because only that pipeline publishes the entire solution to a single directory. If there are conflicts because of transitive dependencies, those tests may fail, but the GitHub Actions workflows won't fail because they don't integration test all of the projects together.

According to Visual Studio Package Manager, the following dependencies are all deprecated, so we will also need to upgrade (or downgrade) them to a supported version.

- Microsoft.AspNetCore.Http.Abstractions 2.0.0
- Microsoft.AspNetCore.TestHost 5.0.0
- Microsoft.Extensions.DependencyInjection 2.0.0

There are also several other packages we are not using the latest version of that we should look into.

In theory, we should be using the lowest supported version to give users the option to use what we specify or upgrade. But, in practice many of these can probably be upgraded to the 6.x generation or higher now.

PR #981, opened 2024-10-21
lucene-cli: Removed target and tests for net6.0 (fixes #979)

### LINKED ISSUE #979: Task: Remove net6.0 target from lucene-cli

### Task description

Since we are now just days away from .NET 6 being out of support, it is no longer sensible to release lucene-cli with a `net6.0` target, as there won't be a supported runtime for it.

We can still have one more release with a `net6.0` target of our class libraries, though.

NightOwl888:

lucene-cli: Removed target and tests for `net6.0`

Fixes #979

## Description

This removes the `net6.0` target from lucene-cli. The runtime to use it will be out of support in a few days, so we don't need it for the next release. This reduces the total distribution size of .nupkg and .snupkg files from 127MB to 108MB, and will save us some disk space during publish, as well.

### DIFF
--- a/src/dotnet/tools/Lucene.Net.Tests.Cli/Lucene.Net.Tests.Cli.csproj
+++ b/src/dotnet/tools/Lucene.Net.Tests.Cli/Lucene.Net.Tests.Cli.csproj
@@ -25,11 +25,11 @@
 
   <PropertyGroup>
     <!-- Allow specific target framework to flow in from TestTargetFrameworks.props -->
-    <TargetFrameworks Condition=" '$(TargetFramework)' == '' ">net8.0;net6.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(TargetFramework)' == '' ">net8.0</TargetFrameworks>
     <AssemblyTitle>Lucene.Net.Tests.Cli</AssemblyTitle>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" $(TargetFramework.StartsWith('net4')) Or $(TargetFramework.StartsWith('net5.')) ">
+  <PropertyGroup Condition=" $(TargetFramework.StartsWith('net4')) Or $(TargetFramework.StartsWith('net5.')) Or $(TargetFramework.StartsWith('net6.')) ">
     <!-- For CI, we don't publish the projects and exclude them from the test run by setting IsPublishable=false and IsTestProject=false -->
     <IsPublishable>false</IsPublishable>
     <IsTestProject>false</IsTestProject>
--- a/src/dotnet/tools/lucene-cli/lucene-cli.csproj
+++ b/src/dotnet/tools/lucene-cli/lucene-cli.csproj
@@ -24,11 +24,11 @@
   <Import Project="$(SolutionDir).build/nuget.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>net8.0</TargetFrameworks>
     <RollForward Condition=" $(TargetFramework.StartsWith('net8.')) ">Major</RollForward>
 
     <IsPublishable>false</IsPublishable>
-    <IsPublishable Condition="'$(TargetFramework)' == 'net8.0' Or '$(TargetFramework)' == 'net6.0'">true</IsPublishable>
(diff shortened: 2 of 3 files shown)


src/dotnet/tools/lucene-cli/lucene-cli.csproj
```diff
@@ -24,11 +24,11 @@
   <Import Project="$(SolutionDir).build/nuget.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>net8.0</TargetFrameworks>
     <RollForward Condition=" $(TargetFramework.StartsWith('net8.')) ">Major</RollForward>
 
     <IsPublishable>false</IsPublishable>
-    <IsPublishable Condition="'$(TargetFramework)' == 'net8.0' Or '$(TargetFramework)' == 'net6.0'">true</IsPublishable>
```
paulirwin, line 31: Do we need these conditions if it's always just net8.0 above?

src/dotnet/tools/Lucene.Net.Tests.Cli/Lucene.Net.Tests.Cli.csproj
```diff
@@ -25,11 +25,11 @@
 
   <PropertyGroup>
     <!-- Allow specific target framework to flow in from TestTargetFrameworks.props -->
-    <TargetFrameworks Condition=" '$(TargetFramework)' == '' ">net8.0;net6.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(TargetFramework)' == '' ">net8.0</TargetFrameworks>
     <AssemblyTitle>Lucene.Net.Tests.Cli</AssemblyTitle>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" $(TargetFramework.StartsWith('net4')) Or $(TargetFramework.StartsWith('net5.')) ">
+  <PropertyGroup Condition=" $(TargetFramework.StartsWith('net4')) Or $(TargetFramework.StartsWith('net5.')) Or $(TargetFramework.StartsWith('net6.')) ">
```
paulirwin, line 32: Do we need these conditions if it's always just net8.0 above?

src/dotnet/tools/lucene-cli/lucene-cli.csproj
NightOwl888: The `<TargetFrameworks>` are specified above. `<TargetFramework>` has the condition. The solution level publish runs all of the targets, so we need the condition on `<TargetFramework>` to exclude the ones that don't apply in CI. This was recently updated in #975, which failed because `net5.0` was allowed to publish with the previous condition.

src/dotnet/tools/Lucene.Net.Tests.Cli/Lucene.Net.Tests.Cli.csproj
NightOwl888: In this case, `TestTargetFramework.props` takes priority over what is defined here *if* it is set to `TargetFrameworks`, which we do in CI. If it is a specific `TargetFramework` (which is what we normally do in the IDE), then we will either respect it or patch it up so there is a valid configuration in the IDE, since the IDE doesn't understand a test project with no targets in it.

src/dotnet/tools/lucene-cli/lucene-cli.csproj
paulirwin: If I understand correctly, does this mean that we are running the CLI tests for each target framework (.NET 6, .NET 8, etc) but always running it as .NET 8 no matter what? If so, that seems wasteful, and we could speed up our test times a little by only running it as part of the .NET 8 tests. But alas, I'm not going to fix what ain't broken here.

NightOwl888: No. This is just a quirk because we have some projects that target fewer and/or different targets than the "standard" set. So, the logic that we use is:

1. In CI, publish only the targets that are supported
2. In the IDE, allow the `TestTargetFrameworks.props` file to either run all targets or some of them.

Then the individual `.csproj` handle these special cases. The IDE needs something there for the selected target or it breaks (in some way I don't recall). The publish also breaks if the `.csproj` cannot handle one of the solution-level targets that is passed in. So this is a hack that handles both.

PR #982, opened 2024-10-21
Target .NET 8, #928

### LINKED ISSUE #928: Target .NET 8

Officially the latest release only supports .Net 6.0. With 6.0 going out of support this year it would be good to support the newer versions. (.Net 8.0) Are there any expected issues with the upgrade?

Added compatibility for .Net 8

### Additional context

_No response_

paulirwin:

Adds the .NET 8 target.

Fixes #928

## Description

This PR adds the .NET 8 target. Due to a conflict between extension methods in .NET 8 and J2N, we added new methods to Collections to create J2N read-only collections instead of using the extension method. The unit tests will use the net8.0 targets when run under .NET 8 now, instead of net6.0.

Additionally, this resolves a warning when running the .NET 5 tests, and ensures the `_artifacts/NuGetPackages` folder is created in the build script before restore, which is helpful when you add that folder as a local path in your NuGet sources for testing purposes.

### DIFF
--- a/src/Lucene.Net/Support/Collections.cs
+++ b/src/Lucene.Net/Support/Collections.cs
@@ -218,11 +218,6 @@ namespace Lucene.Net.Support
             return new ReadOnlyList<T>(list);
         }
 
-        public static ReadOnlyCollection<T> AsReadOnly<T>(ICollection<T> collection)
-        {
-            return new ReadOnlyCollection<T>(collection);
-        }
-
         public static ReadOnlyDictionary<TKey, TValue> AsReadOnly<TKey, TValue>(IDictionary<TKey, TValue> dictionary)
         {
             return new ReadOnlyDictionary<TKey, TValue>(dictionary);
--- a/src/Lucene.Net/Support/Collections.cs
+++ b/src/Lucene.Net/Support/Collections.cs
@@ -88,7 +88,7 @@ namespace Lucene.Net.Support
 
         public static IDictionary<TKey, TValue> SingletonMap<TKey, TValue>(TKey key, TValue value)
         {
-            return new Dictionary<TKey, TValue> { { key, value } }.AsReadOnly();
+            return AsReadOnly(new Dictionary<TKey, TValue> { { key, value } });
         }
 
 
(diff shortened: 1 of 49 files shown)


src/Lucene.Net/Support/Collections.cs
```diff
@@ -213,6 +213,21 @@ public static string ToString(object obj, CultureInfo culture)
             return ToString(obj);
         }
 
+        public static ReadOnlyList<T> AsReadOnly<T>(IList<T> list)
+        {
+            return new ReadOnlyList<T>(list);
+        }
+
+        public static ReadOnlyCollection<T> AsReadOnly<T>(ICollection<T> collection)
```
NightOwl888, line 221: A few things to note here:

1. This method has no callers according to code lens, so YAGNI.
2. In Java, Collections have undefined behavior, but Set, List, and Map all have a specific way they perform structural equality comparisons and do structural formatting. So, it would be best to avoid using `ReadOnlyCollection<T>`, since it has to do extra checks to work out if we are dealing with a `Set`, `List`, or `Dictionary` type before cascading the call to the code that knows how to deal with them, and fallback to a generic implementation if none of them match. So, if we are going for a complete set of collection methods, then `ReadOnlySet<T>` should be the choice here, not `ReadOnlyCollection<T>`. 
3. Each of the read only collections in J2N has a "Default" mode, which kicks in if the type being passed in is a J2N collection, where the rules of structural equality are defined. If it is a BCL or 3rd party collection, it will use "Aggressive" mode, which tries to patch the collection externally with the same behavior. But "Aggressive" mode is very costly in terms of performance.

Given number 3 above, we should only use "Aggressive" mode if we are dealing with a user-supplied collection that we have no control over. However, if we have control over the concrete collection type that is used (or default), we should always ensure that a J2N collection is used for each of these calls so we don't downgrade to "Aggressive" mode. So, it would be a good idea to review to make sure we didn't swap to any BCL collections where it could matter.

It would probably also be sensible to add warnings to our Roslyn code analyzer to warn users that there may be a performance downgrade if they use a BCL collection type in cases where there would be a performance downgrade. This only happens when comparing collection equality or formatting the contents of the collection, IIRC.

We probably should make the above 2 paragraphs into new issues.

NightOwl888 (approved):
Thanks for the PR.

The changes look good. 

This brings us from 93% to 95% of the allowed 10GB on Azure Pipelines, so we are good for now. After #981, this target only brought us from the original distribution size of 127MB up to 130MB.

There was a prior change that hard coded the net6.0 target framework in NativeFSLockFactory which triggered a build warning after the upgrade, which I have addressed in a way that is tolerant of future target framework changes.

src/Lucene.Net/Support/Collections.cs
paulirwin: Removed unused method. Split items #984 and #985.

PR #987, opened 2024-10-23
Replaces old docs project with modern 'no-targets' msbuild project

Shazwazza:
The old docs project in the sln was based on the old .net framework 'websites' since msbuild/.net didn't have a nice way to have a project that just contains files.

Now, with the no-targets SDK project type, this is possible.

This just replaces the old docs project with a new one that uses the no-targets SDK project type


paulirwin (approved):

NightOwl888:
Thanks Shannon. We also need this for the `.github` folder (as a separate project, of course). We often forget to update GitHub actions templates, so that would be helpful. It can be a separate PR, if preferred.

paulirwin:
Agreed, let's do a separate PR for that. This can be a template for how to do it if @Shazwazza is unavailable to finish that one.

rclabo:
Excellent.  I didn't even know about the modern 'no-targets' msbuild project, I had to research it.  It seems like a great fit for holding DocFx-based templates and associated files.

PR #988, opened 2024-10-23
Simplifies solution with no-targets projects for build folders

Shazwazza:
Simplifies solution with no-targets projects for build folders:

Solution now looks like this:

image

and removes a lot of hard coded values in the sln file

### DIFF
--- a/.build/build.msbuildproj
+++ b/.build/build.msbuildproj
@@ -5,6 +5,6 @@
     <OutputPath>bin\</OutputPath>
   </PropertyGroup>
   <ItemGroup>
-	  <None Include="**/*.*" Exclude="obj/**/*;bin/**/*;*.msbuildproj"/>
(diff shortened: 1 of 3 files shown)


NightOwl888:
Yeah, I should have mentioned not to touch the `.build` folder because we use the same config in other projects, but I guess we can deal with it.

The `TestTargetFramework.props` file, is not visible now, though. I suppose we could move the file up to the `.build` folder so it will be visible again, but also need to update all of the references to that physical file to move it there.

NightOwl888:
Or maybe just add it as a linked file. That should work.

Shazwazza:
Yeah I just left the root files in the Solution Items folder manually. I can add that one in.

Shazwazza:
I added that in now

NightOwl888 (approved):

PR #989, opened 2024-10-23
Use .NET 8 target for docs generation, #928

paulirwin:

Uses the .NET 8 target for docs generation.

Related: #928

## Description

The docs generation was using the net6.0 target as the latest available; as of #982 we now have the net8.0 target, so this updates the docfx json files to use that target instead.


NightOwl888 (approved):
I have confirmed that the build works locally, and it looks like all of the files have been updated with `net8.0`. This looks good to merge.

PR #991, opened 2024-10-23
.build/runbuild.ps1: Changed to edit the global.json file instead of overwriting it to account for any future changes to it.

NightOwl888:

.build/runbuild.ps1: Changed to edit the global.json file instead of overwriting it to account for any future changes to it.

This fixes the Azure DevOps and command line build, which was inadvertently broken by #987.

### DIFF
--- a/.build/runbuild.ps1
+++ b/.build/runbuild.ps1
@@ -67,7 +67,7 @@ task Clean -description "This task cleans up the build directory" {
 
 task UpdateLocalSDKVersion -description "Backs up the project.json file and pins the version to $minimumSdkVersion" {
     Backup-File $globalJsonFile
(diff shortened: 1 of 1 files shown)


.build/runbuild.ps1
```diff
@@ -67,7 +67,7 @@ task Clean -description "This task cleans up the build directory" {
 
 task UpdateLocalSDKVersion -description "Backs up the project.json file and pins the version to $minimumSdkVersion" {
```
paulirwin, line 68: Minor thing, but the description refers to the "project.json" file when I think it should be "global.json"

NightOwl888: Yeah. Leftovers from the dnx days, I guess.

PR #992, opened 2024-10-23
Update README.md for .NET support

paulirwin:

Updates the .NET support information in the README.

Related to #928 

## Description

Updates the README to add the .NET 8 target, specify that the CLI requires the .NET 8 runtime, and removes references to older unsupported frameworks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,11 @@ The Apache Lucene.NET website is at:
 
 ### Lucene.NET 4.8.0
 
+- [.NET 8.0](https://dotnet.microsoft.com/download/dotnet/8.0)
 - [.NET 6.0](https://dotnet.microsoft.com/download/dotnet/6.0)
 - [.NET Standard 2.1](https://docs.microsoft.com/en-us/dotnet/standard/net-standard)
 - [.NET Standard 2.0](https://docs.microsoft.com/en-us/dotnet/standard/net-standard)
-- .NET Framework 4.5
+- .NET Framework 4.6.2
 
 ## Status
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,9 +38,9 @@ Working toward Lucene.NET 4.8.0 (currently in BETA)
 
 * The beta version is extremely stable
 * Has more than 7800+ passing unit tests
-* Integrates well with .NET 6.0, .NET 5.0 and .NET Core 2+
+* Integrates well with .NET 8.0 and .NET 6.0 (as well as other unsupported versions)
```
NightOwl888, line 41: I was just reading a comment yesterday that I left about this. This tends to go out of date faster than we can update it. So, maybe we should just say something along the lines of "Tested daily on all [officially supported versions of .NET Core](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core)".

The targets have a loose relationship with the runtimes, but I suspect anyone reading this just wants to know if their preferred runtime will work. If anyone wants to see our supported target frameworks, it is all on the NuGet package page. Maybe we should just drop a link to: https://www.nuget.org/packages/Lucene.Net/absoluteLatest#supportedframeworks-body-tab, which will always be up to date.

Also, this information was copied over to our homepage, which will also need to be updated: https://lucenenet.apache.org/. At one point, these 2 descriptions were identical, but they don't remain in sync because they don't share the same source file.

NightOwl888 (commented):
Thanks for the PR.

I left an inline comment.

Please do not merge this to master until after the 4.8.0-beta00017 release.

NightOwl888 (approved):

Issue #264, opened 2019-10-10
Port RandomizedTesting Test Runner


clambertus:
<p>While the test framework can run with the standard NUnit test runner and we have managed to make it work, it is missing the ability to reproduce a random test failure, which is crucial for debugging some of the more complicated random tests. Lucene uses a custom JUnit test runner called <a href="https://github.com/randomizedtesting/randomizedtesting" class="external-link" rel="nofollow">RandomizedTesting</a> to accomplish this.</p>

<p>It also includes some other nice features</p>

<ol>
	<li>Ensure when a test fails, the random seeds are included in error messages and logs</li>
	<li>Code analysis to ensure the tests are setup properly</li>
	<li>Run tests in a random order to ensure they are not dependent upon each other</li>
</ol>

<p>Preliminary analysis shows that the API for NUnit allows building custom test runners and it is a close enough match to implement the functionality. Most likely, this will require a custom adapter as well, so the test runner can integrate into Visual Studio and dotnet test/mstest.</p>

<p>Do note that without doing some pretty heavy refactoring on the Codec, Directory, and LuceneTestCase classes, it is not possible to run tests in parallel within the same AppDomain because the codecs use a static variable to turn codec impersonation on/off. For now, it would probably be best to run tests serially.</p>
<i>JIRA link - <a href="https://issues.apache.org/jira/browse/LUCENENET-627">[LUCENENET-627]</a> created by nightowl888</i>

NightOwl888:
There may be some compelling reasons that make RandomizedTesting useful for Lucene.NET: https://chatgpt.com/share/677a832d-3424-8005-97d7-1c024c68302d.

There are a few features that Lucene uses that we don't have (such as running tests in random order and varying other parts of the test run). However, one benefit that makes a slightly stronger case for it: It would provide a level of indirection between NUnit's release schedule and Lucene.NET's schedule. In theory, we could roll out a new version of RandomizedTesting to keep up with NUnit changes without forcing us to also update Lucene.NET.

While at this point I still think it is not required for the release of Lucene.NET 4.8.0, even with the latest update on #1017, it does mean that everything that we import from RandomizedTesting should be made internal and not available for end users. This is a must if we are going to expose it publicly later in another library. Ideally, we would move everything into the RandomizedTesting namespace and physically separate it from the test framework to make it less of a burden to make public later.

[...]

PR #1000, opened 2024-10-26
Upgrade NUnit to 3.14 and update test adapter and VS test SDK packages

paulirwin:

Upgrade NUnit to 3.14.0, as well as the test adapter and VS test SDK packages.

Related #259 

## Description

This fixes an issue in Rider with test discovery, impacting the ability to see and individually run child tests of test classes.

### DIFF
--- a/.build/dependencies.props
+++ b/.build/dependencies.props
@@ -57,7 +57,7 @@
     <MicrosoftExtensionsDependencyInjectionAbstractionsPackageVersion>8.0.0</MicrosoftExtensionsDependencyInjectionAbstractionsPackageVersion>
     <MicrosoftExtensionsDependencyInjectionPackageVersion>8.0.1</MicrosoftExtensionsDependencyInjectionPackageVersion>
     <MicrosoftNETFrameworkReferenceAssembliesPackageReferenceVersion>1.0.3</MicrosoftNETFrameworkReferenceAssembliesPackageReferenceVersion>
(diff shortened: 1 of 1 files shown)


NightOwl888 (approved):
The good news is we can upgrade to 3.14.0 because the features that NUnit added were already present in 3.13.1, which is the version we were targeting when repeatable random test support was added.

The bad news is, until we do some more work to stabilize and add tests to the test framework (which is still experimental, see #547), I don't think upgrades are a thing we can support.

We are missing the scans for the [FixtureLifecycleAttribute](https://github.com/nunit/nunit/blob/v3.14.0/src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs#L79), and I don't recall whether that was an intentional design choice or if I was porting from an older version of NUnit (this line was added in 3.13.1). But as long as we own a chunk of NUnit code, we cannot guarantee new NUnit features will work if someone upgrades the NUnit version. IIRC, we also don't support all of NUnit's features (similar to how randomized-testing doesn't support all of JUnit's features).

I think we can leave this as-is for now, but we should probably consider whether to put a version constraint on this so upgrades to 4.x are not possible for users. At least until we can spec out some extensibility for NUnit to add so we don't have to own a chunk of their code. But before we do that, we really should finish #1017 so we can work out what other parts of NUnit will need extending to support our test framework and randomized testing (#264) going forward.

PR #945, opened 2024-06-05
FuzzyQuery produces a wrong result when prefix is equal to the term length

tohidemyname:
- [x] You've read the [Contributor Guide](https://github.com/apache/lucenenet/blob/main/CONTRIBUTING.md) and [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html).
- [x] You've included unit or integration tests for your change, where applicable.
- [x] You've included inline docs for your change, where applicable.
- [x] There's an open issue for the PR that you are making. If you'd like to propose a change, please [open an issue](https://github.com/apache/lucenenet/issues/new/choose) to discuss the change or find an existing issue.

Summary of the changes (Less than 80 chars)

Fixes #{bug number} (in this specific format)
https://github.com/apache/lucenenet/issues/941
## Description

{Detail}

PR #1002, opened 2024-10-29
FuzzyQuery produces a wrong result when prefix is equal to the term length

### LINKED ISSUE #941: FuzzyQuery produces a wrong result when prefix is equal to the term length

### Describe the bug

When using FuzzyQuery the search string `bba` does not match doc value `bbab` with an edit distance of 1 and prefix length of 3.

In FuzzyQuery an automaton is created for the "suffix" part of the search string which in this case is an empty string.

### Expected Behavior

In this scenario maybe the FuzzyQuery should rewrite to a WildcardQuery of the following form :

    searchString + "?" 
where there's an appropriate number of ? characters according to the edit distance.

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### Lucene.NET Version

_No response_

### .NET Version

_No response_

### Operating System

_No response_

### Anything else?

_No response_

paulirwin:

FuzzyQuery produces a wrong result when prefix is equal to the term length

Fixes #941 

## Description

See #941

(NOTE: This PR was originally opened as #945 but included an unexpected commit. This PR cherry-picks that commit.)

### DIFF
--- a/src/Lucene.Net.Tests/Search/TestFuzzyQuery.cs
+++ b/src/Lucene.Net.Tests/Search/TestFuzzyQuery.cs
@@ -195,6 +195,8 @@ namespace Lucene.Net.Search
             reader.Dispose();
             directory.Dispose();
         }
+
+        // LUCENENET-specific: backported fix from Lucene 9.0.0 (lucene@45611d0, LUCENE-9365)
         [Test]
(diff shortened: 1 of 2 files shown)


paulirwin:
@NightOwl888 Thanks for finding the PR, I only included the commit hash in the comment as that was what I had found at the time.

PR #1005, opened 2024-10-30
website/site/.htaccess - bug fix by removing BOM and update to beta0017 redirection

rclabo:

Fixes a website/site/.htaccess BOM bug, also updates to beta00017 redirects

## Description
Apache web server requires that the .htaccess file not be saved with a BOM.  The one previously committed accidentally contained a BOM, causing issues when deployed to the asf-site branch of the lucent-site repo.  This PR fixes that and also updates the related /docs/absolute-latest/ redirection to be to /docs/4.8.0-beta00017/

### DIFF
--- a/websites/site/.htaccess
+++ b/websites/site/.htaccess
@@ -1,4 +1,4 @@
-﻿# -----------------------------------------------------------------------------------
+# -----------------------------------------------------------------------------------
 #
 # Licensed to the Apache Software Foundation (ASF) under one or more
 # contributor license agreements.  See the NOTICE file distributed with
@@ -32,7 +32,8 @@
 # version, with [R=301] signaling permanent redirects and [L] preventing further
 # rule processing.
 #
-# Note: This file is line ending sensitive. MUST use LF line endings.
+# Note: This file is line ending and BOM sensitive. It MUST use LF line endings and 
+# MUST NOT have a BOM.
 
 
 RewriteEngine On
@@ -40,5 +41,5 @@ RewriteEngine On
 # Redirect /docs/latest/ to /docs/3.0.3/
 RewriteRule ^docs/latest/(.*)$ /docs/3.0.3/$1 [R=301,L]
 
-# Redirect /docs/absolute-latest/ to /docs/4.8.0-beta00016/
-RewriteRule ^docs/absolute-latest/(.*)$ /docs/4.8.0-beta00016/$1 [R=301,L]
+# Redirect /docs/absolute-latest/ to /docs/4.8.0-beta00017/
+RewriteRule ^docs/absolute-latest/(.*)$ /docs/4.8.0-beta00017/$1 [R=301,L]
--- a/websites/site/.gitattributes
+++ b/websites/site/.gitattributes
@@ -1,1 +1,1 @@
-.htaccess text eol=lf
+.htaccess text eol=lf working-tree-encoding=UTF-8


websites/site/.htaccess
```diff
@@ -32,13 +32,14 @@
...
 
 RewriteEngine On
 
 # Redirect /docs/latest/ to /docs/3.0.3/
 RewriteRule ^docs/latest/(.*)$ /docs/3.0.3/$1 [R=301,L]
 
-# Redirect /docs/absolute-latest/ to /docs/4.8.0-beta00016/
-RewriteRule ^docs/absolute-latest/(.*)$ /docs/4.8.0-beta00016/$1 [R=301,L]
+# Redirect /docs/absolute-latest/ to /docs/4.8.0-beta00017/
+RewriteRule ^docs/absolute-latest/(.*)$ /docs/4.8.0-beta00017/$1 [R=301,L]
```
NightOwl888, line 45: We should probably update the release policy so this update doesn't get skipped.

rclabo: Ok good idea. Is this the file for the release policy? https://lucenenet.apache.org/contributing/make-release.html

NightOwl888: Yeah. But, actually it probably would be better to automate it (if we can get the file format right). docs.ps1 already does some replacements based on regex. However, the approach used always adds a line to the end of the file. We probably should find a solution to that if we do it.

But I don't feel strongly whether we do the automation or put it into the checklist so it is done when we release.

rclabo: I could probably automate it by modifying the website/site/site.ps1 to get the value like '4.8.0-beta00017' from the last sorted doc in the /site/download/ folder ie version-4.8.0-beta00017.md and then updating the value in the .htaccess before copying it into the _site folder.  How does that sound?  Or do you have ideas on a better source to get the highest value like '4.8.0-beta00017'  value from?

NightOwl888: Yeah, this seems to be a hard problem. Ideally, the docs would be built as part of the deployment pipeline where the PackageVersion is available, but currently that is all manual except for the NuGet push.

I updated the procedure in: https://github.com/apache/lucenenet/pull/1010 so this can be dealt with manually for the time being. It definitely needs to be part of the checklist, or we risk skipping it during a release.

The website also failed to build again because the Powershell copy failed again (not sure why this time, I previously added the script to create the directory if it doesn't exist). Since this is so troublesome, I added the `.htaccess` and `.gitattributes` files to the `docfx.json` file as resources so they would get copied by docfx during the build rather than relying on Powershell to do it. We already were doing this for the DOAP file.

I will also cherry-pick this commit along with mine to the `docs/4.8.0-beta00017` branch, which is where we are supposed to be building the website and docs from (since that branch freezes the API at 4.8.0-beta00017). We definitely don't want to merge master into that branch again, though. Although it is fine to merge it into master.

rclabo: I think the way I proposed to automate this should have worked fine but I was waiting for your feedback so that I didn't have to do the work twice.  That approach would have been based on Powershell.  Switching to using docfx to copy the files is cool and all but it heads in the opposite direction of automating .htaccess modification to auto set the destination for /absolute-latest/.  (shrug)  It's odd to me that Powershell was deemed an unstable mechanism for copying files.

PR #1021, opened 2024-11-13
Improve message for unit test failures for lucene.testsettings.json, #999

### LINKED ISSUE #999: Change failed test message to use lucene.testsettings.json instead of .runsettings

### Task description

When a unit test fails, it prints a message instructing the user on how they can set the random seed and culture to reproduce the test. The output today looks like:

```

Option 1:

[assembly: Lucene.Net.Util.RandomSeed("0x9a2b7430d6d33f0d")]
[assembly: NUnit.Framework.SetCulture("en-IE")]

Option 2:

<RunSettings>
  <TestRunParameters>
    <Parameter name="tests:seed" value="0x9a2b7430d6d33f0d" />
    <Parameter name="tests:culture" value="en-IE" />
  </TestRunParameters>
</RunSettings>

See the .runsettings documentation at: https://docs.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file.
```

The .runsettings option is more verbose than the lucene.testsettings.json version, it requires structuring the XML properly to work, and has the extra documentation link that could be removed with the JSON approach. This message could instead look like:

```

Option 1:

[assembly: Lucene.Net.Util.RandomSeed("0x9a2b7430d6d33f0d")]
[assembly: NUnit.Framework.SetCulture("en-IE")]

Option 2:

 Create the following lucene.testsettings.json file somewhere between the test assembly and your root drive:

{
  "tests": {
     "seed": "0x9a2b7430d6d33f0d",
     "culture": "en-IE"
  }
}
```

paulirwin:

Improves the failed unit test message to provide an example lucene.testsettings.json.

Fixes #999

## Description

The lucene.testsettings.json seems to work more reliably and is a little more user-friendly than the runsettings file, so this improves the failed unit test message to provide a generated lucene.testsettings.json file. This also uses C# 11 multi-line strings instead of manual string concatenation.

### DIFF
--- a/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
+++ b/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
@@ -947,6 +947,17 @@ namespace Lucene.Net.Util
 
                       Option 2:
 
+                       Use the following .runsettings file:
+
+                      <RunSettings>
+                        <TestRunParameters>
+                          <Parameter name="tests:seed" value="{{RandomizedContext.CurrentContext.RandomSeedAsHex}}" />
+                          <Parameter name="tests:culture" value="{{Thread.CurrentThread.CurrentCulture.Name}}" />
+                        </TestRunParameters>
+                      </RunSettings>
+
+                      Option 3:
+
                        Create the following lucene.testsettings.json file somewhere between the test assembly and the root of your drive:
 
                       {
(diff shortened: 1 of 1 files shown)


src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
```diff
@@ -929,23 +929,35 @@ public virtual void TearDown()
...
+                      To reproduce this test result:
+
+                      Option 1:
+
+                       Apply the following assembly-level attributes:
+
+                      [assembly: Lucene.Net.Util.RandomSeed("{{RandomizedContext.CurrentContext.RandomSeedAsHex}}")]
+                      [assembly: NUnit.Framework.SetCulture("{{Thread.CurrentThread.CurrentCulture.Name}}")]
+
+                      Option 2:
```
NightOwl888, line 948: Could we just add this as Option 3? The syntax of the `.runsettings` file is hard to remember so I would rather not leave it out of the message entirely.

PR #1049, opened 2024-11-26
IAttribute and ICharTermAttribute method changes, #1038

### LINKED ISSUE #1038: Remove SetLength and SetEmpty from ICharTermAttribute
There doesn't seem to be a good reason to force implementations to define `SetLength()`, since it should just cascade the call to the setter of `Length`. We should remove this from the interface and make this method into an extension method on `ICharTermAttribute` (in a `Lucene.Net.Analysis.TokenAttributes.Extensions` namespace). It should also be made to return the type of `ICharTermAttribute` it is called upon rather than the interface type.

_Originally posted by @NightOwl888 in https://github.com/apache/lucenenet/pull/1028#discussion_r1848165854_

Update: We should also remove SetEmpty and make it an extension method as well, and add a Clear() method in its place for API consistency.

paulirwin:

Remove SetLength and SetEmpty from ICharTermAttribute and moves them to extension methods. Add `Clear()` in place of SetEmpty.

Fixes #1038

## Description

See the rationale in #1038.

This PR makes the following breaking changes:

- Removes SetLength and SetEmpty from ICharTermAttribute, moves them to extensions instead.
- Adds the Clear method to ICharTermAttribute.
- Removes the CopyTo method from IAttribute.


src/Lucene.Net.TestFramework/Index/BaseTermVectorsFormatTestCase.cs
```diff
@@ -360,7 +360,8 @@ public sealed override bool IncrementToken()
             {
                 if (i < terms.Length)
                 {
-                    termAtt.SetLength(0).Append(terms[i]);
+                    termAtt.Length = 0;
```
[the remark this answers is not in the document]
paulirwin: Done; added the extension method for SetLength, added Clear() to the interface (it was already on the concrete type via implementing Attribute's abstract method), moved SetEmpty to the extensions, and updated doc references.

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttribute.cs
```diff
@@ -62,35 +61,29 @@ public interface ICharTermAttribute : IAttribute, ICharSequence, IAppendable
...
+        // LUCENENET specific: Redefining this[] to make it settable
+        new char this[int index] { get; set; }
 
         /// <summary>
         /// Sets the length of the termBuffer to zero.
         /// Use this method before appending contents.
         /// </summary>
-        ICharTermAttribute SetEmpty();
+        /// <seealso cref="Lucene.Net.Analysis.TokenAttributes.Extensions.CharTermAttributeExtensions.SetEmpty(ICharTermAttribute)"/>
+        void Clear();
```
NightOwl888, line 86: We have already [deviated from upstream on `IAttribute`](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/util/Attribute.java) by [adding a `CopyTo()` method](https://github.com/apache/lucenenet/blob/bcd49d8c584cdcee480f80fa46c62495740b2bc4/src/Lucene.Net/Util/Attribute.cs#L23). Like `Clear()`, `CopyTo()` is also an abstract member of `Attribute`. I don't see any reason why we shouldn't treat `Clear()` the same way.

So, please move this declaration to `IAttribute` instead of here. Since `ICharTermAttribute` implements `IAttribute`, it will also provide a way to call `Clear()` from the extension method.

Granted, there may be a good reason not to do this, but I am struggling to find one. Since an attribute represents a dynamically added piece of data, the ability to clear that data seems like it will always be a requirement.

Adding `CopyTo()` to `IAttrbute` was a workaround because in .NET we would require a cast to do it the same way they did in Java, and this seems like a very similar scenario.

src/Lucene.Net/Analysis/Token.cs
```diff
@@ -64,15 +65,15 @@ namespace Lucene.Net.Analysis
     /// Failing that, to create a new <see cref="Token"/> you should first use
     /// one of the constructors that starts with null text.  To load
     /// the token from a char[] use <see cref="ICharTermAttribute.CopyBuffer(char[], int, int)"/>.
-    /// To load from a <see cref="string"/> use <see cref="ICharTermAttribute.SetEmpty()"/> followed by 
+    /// To load from a <see cref="string"/> use <see cref="ICharTermAttribute.SetEmpty()"/> followed by
```
NightOwl888, line 68: Please fix this reference, as it is not resolving.

src/Lucene.Net/Analysis/Token.cs
```diff
@@ -64,15 +65,15 @@ namespace Lucene.Net.Analysis
...
     /// <see cref="ICharTermAttribute.Append(string)"/> or <see cref="ICharTermAttribute.Append(string, int, int)"/>.
     /// Alternatively you can get the <see cref="Token"/>'s termBuffer by calling either <see cref="ICharTermAttribute.Buffer"/>,
     /// if you know that your text is shorter than the capacity of the termBuffer
     /// or <see cref="ICharTermAttribute.ResizeBuffer(int)"/>, if there is any possibility
     /// that you may need to grow the buffer. Fill in the characters of your term into this
     /// buffer, with <see cref="string.ToCharArray(int, int)"/> if loading from a string,
-    /// or with <see cref="System.Array.Copy(System.Array, int, System.Array, int, int)"/>, 
-    /// and finally call <see cref="ICharTermAttribute.SetLength(int)"/> to
+    /// or with <see cref="System.Array.Copy(System.Array, int, System.Array, int, int)"/>,
+    /// and finally assign to <see cref="ICharTermAttribute.Length"/> to
```
NightOwl888, line 76: Please change this to read `finally set the <see cref="ICharTermAttribute.Length"/> of the term text.`

src/Lucene.Net.Tests.Analysis.Common/Analysis/Position/PositionFilterTest.cs
```diff
@@ -105,7 +107,7 @@ public virtual void TestReset()
         /// <summary>
         /// Tests ShingleFilter up to six shingles against six terms.
         ///  Tests PositionFilter setting all but the first positionIncrement to zero. </summary> </exception>
-        /// <exception cref="java.io.IOException"> <seealso cref= Token#next(Token) </seealso>
+        /// <exception cref="IOException"> <seealso cref= Token#next(Token) </seealso>
```
NightOwl888, line 110: Please also fix the `<seealso>`  reference.

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttributeImpl.cs
```diff
@@ -489,9 +484,7 @@ public override void CopyTo(IAttribute target) // LUCENENET specific - intention
 
         char[] ICharTermAttribute.ResizeBuffer(int newSize) => ResizeBuffer(newSize);
 
-        ICharTermAttribute ICharTermAttribute.SetLength(int length) => SetLength(length);
-
-        ICharTermAttribute ICharTermAttribute.SetEmpty() => SetEmpty();
+        void ICharTermAttribute.Clear() => Clear();
```
NightOwl888, line 487: We technically only need these explicit declarations to change the return type of the public methods that implement the interface. Since this returns `void`, we don't need it.

NightOwl888 (changes requested):

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttribute.cs
paulirwin: I think this is a good idea, and most implementations seem to be using the `Attribute` abstract class anyways, which requires implementing it.

src/Lucene.Net.Tests.Analysis.Common/Analysis/Position/PositionFilterTest.cs
paulirwin: I am not sure what this is supposed to be pointing to, seems like an outdated reference. Token does not have a `Next(Token)` method in either .NET or Java. It looks from the upstream CHANGES doc that this hasn't been a thing since Lucene 2.9. PositionFilter also doesn't seem to throw IOException itself. Going to just remove this for now.

src/Lucene.Net.Analysis.Common/Analysis/NGram/NGramTokenFilter.cs
```diff
@@ -124,6 +124,10 @@ public int PositionIncrement
 
             // LUCENENET specific - The interface requires this to be implemented, since we added it to avoid casts.
             public void CopyTo(IAttribute target) => _ = target;
```
NightOwl888, line 126: Now that `CopyTo()` and `Clear()` are not defined on `IAttribute`, they can be removed to align with upstream.

src/Lucene.Net.Analysis.Common/Analysis/NGram/NGramTokenFilter.cs
```diff
@@ -136,6 +140,10 @@ public int PositionLength
...
```
NightOwl888, line 142: Now that `CopyTo()` and `Clear()` are not defined on `IAttribute`, they can be removed to align with upstream.

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttribute.cs
```diff
@@ -285,5 +272,16 @@ public interface ICharTermAttribute : IAttribute, ICharSequence, IAppendable
...
+        /// <summary>
+        /// Clears the values in this attribute and resets it to its
+        /// default value.
+        /// </summary>
+        /// <remarks>
+        /// LUCENENET specific - This method is not part of the Java Lucene API.
+        /// This was added to be a more consistent way to clear attributes than SetEmpty().
+        /// </remarks>
+        /// <seealso cref="CharTermAttributeExtensions.SetEmpty(ICharTermAttribute)"/>
+        void Clear();
```
NightOwl888, line 285: Since we are effectively replacing `SetEmpty()` with `Clear()` on this attribute, please move the declaration above the `Append()` methods so it can remain in the same order as upstream.

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttributeImpl.cs
```diff
@@ -107,32 +107,27 @@ private void GrowTermBuffer(int newSize)
             }
         }
 
-        int ICharTermAttribute.Length { get => Length; set => SetLength(value); }
+        int ICharTermAttribute.Length
```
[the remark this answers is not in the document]
paulirwin: It appears that also the explicit implementations for `ICharTermAttribute.Buffer`, `ICharSequence.this[int index]`, `ICharTermAttribute.this[int index]`, `ICharTermAttribute.CopyBuffer`, and `ICharTermAttribute.ResizeBuffer` are not needed either (i.e. for covariant return types). I'll go ahead and remove those as well since they just delegate to the same methods/properties/indexer that would be called if implemented implicitly.

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttribute.cs
```diff
@@ -62,37 +62,33 @@ public interface ICharTermAttribute : IAttribute, ICharSequence, IAppendable
...
-        /// difference is that this method returns a reference to the current object so it can be chained.
-        /// <code>
-        /// obj.SetLength(30).Append("hey you");
-        /// </code>
         /// </summary>
-        /// <param name="length"> the truncated length </param>
-        ICharTermAttribute SetLength(int length);
+        /// <remarks>
+        /// LUCENENET: To mimic StringBuilder, we allow this to be settable.
+        /// This replaces the chainable SetLength method in the Java version.
```
NightOwl888, line 74: It isn't really fair to say this replaces the chainable `SetLength()` method as to supplement it. Let's change this to read:

```c#
// The setter may be used as an alternative to <see cref="CharTermAttributeExtensions.SetLength{T}(T, int)"/> if chaining is not required.
```

NightOwl888 (changes requested):
Alright, I think this will be the last change. Sorry to go back and forth so much on this one, it is one of the primary APIs and it is good to be thorough.

NightOwl888 (approved):

PR #1054, opened 2024-12-02
Fix intermittent failed TestMaxClauseLimitations tests, #1050

### LINKED ISSUE #1050: Intermittent Failing Test: Lucene.Net.Search.TestMultiTermQueryRewrites.TestMaxClauseLimitations()

### Describe the bug

The `Lucene.Net.Search.TestMultiTermQueryRewrites.TestMaxClauseLimitations()` test fails under rare random conditions, but does not seem to repeat with the same random seed and culture.

### Expected Behavior

The `Lucene.Net.Search.TestMultiTermQueryRewrites.TestMaxClauseLimitations()` test should always succeed.

### Steps To Reproduce

Add the `[Repeat(100000)]` attribute to the test and it will fail frequently.

### Exceptions (if any)

```
TestMaxClauseLimitations
   Source: TestMultiTermQueryRewrites.cs line 295
   Duration: 2.6 sec

  Message: 
Expected: CheckMaxClauseCount, Actual: Collect

Option 1:

[assembly: Lucene.Net.Util.RandomSeed("0x933174838b8c2dfd")]
[assembly: NUnit.Framework.SetCulture("fr-BL")]

Option 2:

<RunSettings>
  <TestRunParameters>
    <Parameter name="tests:seed" value="0x933174838b8c2dfd" />
    <Parameter name="tests:culture" value="fr-BL" />
  </TestRunParameters>
</RunSettings>

Option 3:

{
  "tests": {
     "seed": "0x933174838b8c2dfd",
     "culture": "fr-BL"
  }
}


  Stack Trace: 
TestMultiTermQueryRewrites.CheckMaxClauseLimitation(RewriteMethod method, String memberName) line 268
TestMultiTermQueryRewrites.TestMaxClauseLimitations() line 297
InvokeStub_TestMultiTermQueryRewrites.TestMaxClauseLimitations(Object, Object, IntPtr*)
1)    at Lucene.Net.Search.TestMultiTermQueryRewrites.CheckMaxClauseLimitation(RewriteMethod method, String memberName) in F:\Projects\lucenenet\src\Lucene.Net.Tests\Search\TestMultiTermQueryRewrites.cs:line 262
TestMultiTermQueryRewrites.TestMaxClauseLimitations() line 297
InvokeStub_TestMultiTermQueryRewrites.TestMaxClauseLimitations(Object, Object, IntPtr*)

```

### Lucene.NET Version

4.8.0-beta00017

### .NET Version

net8.0

### Operating System

Windows

### Anything else?

It is known to fail on Windows x64 with net8.0. It may fail also fail under other configurations.

This test should be compared with the behavior in Lucene 4.8.0. For some reason we are sometimes seeing the exception thrown from the `Collect()` method when we are expecting it to always be thrown from the `CheckMaxClauseCount()` method according to the way the test is written. 

Note also that the [assert message is commented out](https://github.com/apache/lucenenet/blob/6d62a5765e9134440fa491b2358bc520b3330347/src/Lucene.Net.Tests/Search/TestMultiTermQueryRewrites.cs#L265), and it should not be.

paulirwin:

Fix intermittent failed TestMaxClauseLimitations tests

Fixes #1050

## Description

These tests occasionally fail on .NET 8 due to Dynamic PGO optimizing the code and inlining some methods, which changes the stack trace. Instead of harming performance by forcing NoInlining, since this case inspecting the stack trace was only for an assertion rather than changing behavior (as other cases of inspecting the stack trace for specific methods do), this assertion was removed. It should only matter here if the exception is thrown; not which specific method throws it. The `Assert.Fail` above this `catch` explains exactly that, and the original comment was even questioning whether it should be removed.

### DIFF
--- a/src/Lucene.Net.Tests/Search/TestMultiTermQueryRewrites.cs
+++ b/src/Lucene.Net.Tests/Search/TestMultiTermQueryRewrites.cs
@@ -1,8 +1,6 @@
 ﻿using Lucene.Net.Documents;
 using NUnit.Framework;
 using System;
-using System.Diagnostics;
-using System.Runtime.CompilerServices;
 using Assert = Lucene.Net.TestFramework.Assert;
 using Console = Lucene.Net.Util.SystemConsole;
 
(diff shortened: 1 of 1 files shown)


src/Lucene.Net.Tests/Search/TestMultiTermQueryRewrites.cs
```diff
@@ -243,36 +241,44 @@ public virtual void TestBoosts()
...
             {
+                // LUCENENET: The assertion below fails when Dynamic PGO is enabled (by default in .NET 8+) which can inline
+                // some methods at runtime. This causes the stack trace to be different than expected. Rather than forcing
+                // NoInlining on the method, we will just remove the assertion. It should only matter that the exception is
+                // thrown, not where it is thrown from.
+                // Original comment and assertion below:
+
                 //  Maybe remove this assert in later versions, when internal API changes:
-                Assert.AreEqual("CheckMaxClauseCount", new StackTrace(e, false).GetFrames()[0].GetMethod().Name); //, "Should throw BooleanQuery.TooManyClauses with a stacktrace containing checkMaxClauseCount()");
+                // Assert.AreEqual("CheckMaxClauseCount", new StackTrace(e, false).GetFrames()[0].GetMethod()?.Name, "Should throw BooleanQuery.TooManyClauses with a stacktrace containing checkMaxClauseCount()");
```
NightOwl888, line 266: Alright, this case we can let pass.

I wish this had used the `StackTraceHelper` so it would be clear that this is done in several places the same way (that is, declaring certain methods with `NoInlining` to ensure the tests can function). This one was missed during the sweep to apply `NoInlining` because `StackTraceHelper` was not used, which is why it intermittently failed.

NightOwl888 (approved):

PR #1056, opened 2024-12-04
Add operator overrides for public IComparable types, #683

### LINKED ISSUE #683: When implementing IComparable, you should also override ==, !=, <, <=, >, and >=
Mentioned in #648 

Issues found: https://sonarcloud.io/project/issues?resolved=false&rules=csharpsquid%3AS1210&id=apache_lucenenet

- [ ] Public classes - If the class is public, we should definitely implement these.
- [ ] Non-public classes - If not public, it probably doesn't matter. But we should do a review just in case.

paulirwin:

Add operator overrides for public IComparable types

Fixes #683

## Description

This is to resolve the Sonar warning about needing to override the comparison operators for any types that implement IComparable. We are only doing this here for the public types.

QualityQuery was missing Equals and GetHashCode, as well as had some possible NullReferenceExceptions, so those have been fixed as well.

### DIFF
--- a/src/Lucene.Net.Benchmark/Quality/QualityQuery.cs
+++ b/src/Lucene.Net.Benchmark/Quality/QualityQuery.cs
@@ -28,13 +28,13 @@ namespace Lucene.Net.Benchmarks.Quality
     /// <para/>
     /// The ID allows to map the quality query with its judgements.
     /// <para/>
-    /// The name-value pairs are used by a 
+    /// The name-value pairs are used by a
     /// <see cref="IQualityQueryParser"/>
     /// to create a Lucene <see cref="Search.Query"/>.
     /// <para/>
     /// It is very likely that name-value-pairs would be mapped into fields in a Lucene query,
     /// but it is up to the QualityQueryParser how to map - e.g. all values in a single field,
-    /// or each pair as its own field, etc., - and this of course must match the way the 
+    /// or each pair as its own field, etc., - and this of course must match the way the
     /// searched index was constructed.
     /// </summary>
     public class QualityQuery : IComparable<QualityQuery>
@@ -99,5 +99,28 @@ namespace Lucene.Net.Benchmarks.Quality
                 return queryID.CompareToOrdinal(other.queryID);
             }
         }
+
+        #region Operator overrides
+        // LUCENENET specific - per csharpsquid:S1210, IComparable<T> should override comparison operators
+
+        public static bool operator <(QualityQuery left, QualityQuery right)
+            => left is null ? right is not null : left.CompareTo(right) < 0;
+
+        public static bool operator <=(QualityQuery left, QualityQuery right)
+            => left is null || left.CompareTo(right) <= 0;
+
+        public static bool operator >(QualityQuery left, QualityQuery right)
+            => left is not null && left.CompareTo(right) > 0;
+
+        public static bool operator >=(QualityQuery left, QualityQuery right)
+            => left is null ? right is null : left.CompareTo(right) >= 0;
+
+        public static bool operator ==(QualityQuery left, QualityQuery right)
+            => left?.Equals(right) ?? right is null;
+
+        public static bool operator !=(QualityQuery left, QualityQuery right)
+            => !(left == right);
+
+        #endregion
     }
 }
--- a/src/Lucene.Net.Benchmark/Quality/QualityQuery.cs
+++ b/src/Lucene.Net.Benchmark/Quality/QualityQuery.cs
@@ -102,14 +102,12 @@ namespace Lucene.Net.Benchmarks.Quality
         }
 
         // LUCENENET specific - provide Equals and GetHashCode due to providing operator overrides
-        protected bool Equals(QualityQuery? other) => queryID == other?.queryID;
-
         public override bool Equals(object? obj)
         {
             if (obj is null) return false;
             if (ReferenceEquals(this, obj)) return true;
             if (obj.GetType() != GetType()) return false;
-            return Equals((QualityQuery)obj);
+            return queryID == ((QualityQuery)obj).queryID;
         }
 
         public override int GetHashCode() => queryID.GetHashCode();
--- a/src/Lucene.Net.Benchmark/Quality/QualityQuery.cs
+++ b/src/Lucene.Net.Benchmark/Quality/QualityQuery.cs
@@ -3,6 +3,7 @@ using J2N.Text;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+#nullable enable
 
 namespace Lucene.Net.Benchmarks.Quality
 {
@@ -49,8 +50,8 @@ namespace Lucene.Net.Benchmarks.Quality
         /// <param name="nameValPairs">The contents of this quality query.</param>
         public QualityQuery(string queryID, IDictionary<string, string> nameValPairs)
         {
-            this.queryID = queryID;
-            this.nameValPairs = nameValPairs;
+            this.queryID = queryID ?? throw new ArgumentNullException(nameof(queryID));
+            this.nameValPairs = nameValPairs ?? throw new ArgumentNullException(nameof(nameValPairs));
         }
 
         /// <summary>
@@ -66,10 +67,9 @@ namespace Lucene.Net.Benchmarks.Quality
         /// </summary>
         /// <param name="name">The name whose value should be returned.</param>
         /// <returns></returns>
-        public virtual string GetValue(string name)
+        public virtual string? GetValue(string name)
         {
-            nameValPairs.TryGetValue(name, out string result);
-            return result;
+            return nameValPairs.TryGetValue(name, out string? result) ? result : null;
         }
 
         /// <summary>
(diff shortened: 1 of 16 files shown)


NightOwl888:
> This is a draft PR, because I'm currently unsure about whether to override Equals and GetHashCode (and if so, how exactly to do that, i.e. what to compare) for QualityQuery. @NightOwl888 - please let me know your thoughts.
> 

Well, being that they didn't override the default behavior (which is reference equality), a direct port would be to override each and then cascade the call to the base class.

--------------------

But given the fact that the comparer uses `queryID` and it says "ID of this quality query" for a description, it would probably be best to use that to compare for equality/hash code. It would keep the equality and compare checks exactly in sync.

BTW - I noticed that the `QualityQuery.CompareTo()` implementation has an exception handler that we can remove because we can use `int.TryParse()` and fallback to string comparison if parsing either side returns `false`.

```c#
        public virtual int CompareTo(QualityQuery other)
        {
            try
            {
                // compare as ints when ids ints
                int n = int.Parse(queryID, CultureInfo.InvariantCulture);
                int nOther = int.Parse(other.queryID, CultureInfo.InvariantCulture);
                return n - nOther;
            }
            catch (Exception e) when (e.IsNumberFormatException())
            {
                // fall back to string comparison
                return queryID.CompareToOrdinal(other.queryID);
            }
        }
```

The constructor could also be improved by using guard clauses on `queryID` and `nameValPairs` since there will be NREs if either is passed `null`. If we do so, we don't need to change `CompareTo()` to account for the fact that `queryID` may be `null` (which it currently doesn't handle).

`#nullable enable` would catch this sort of thing.

src/Lucene.Net.Suggest/Suggest/Lookup.cs
```diff
@@ -122,14 +122,37 @@ public int CompareTo(LookupResult o)
             {
                 return CHARSEQUENCE_COMPARER.Compare(Key, o.Key);
```
NightOwl888, line 123: Please override `Equals(object)` and `GetHashCode()` to suppress the warnings.

I think the most logical thing to do here is to continue using reference equality by cascading the calls to the base class.

NightOwl888 (changes requested):
Thanks for the PR.

We are getting some build warnings in some types where `Equals(object)` and `GetHashCode()` have not been implemented.

Also, since these are all on nullable types, please use `#nullable enable` within the  `#region Operator overrides` sections and `#nullable restore` prior to `#endregion`. We don't necessarily have to do this on the entire file in this PR.

See the inline comments.

src/Lucene.Net.Benchmark/Quality/QualityQuery.cs
```diff
@@ -82,22 +82,59 @@ public virtual string GetValue(string name)
...
-                return queryID.CompareToOrdinal(other.queryID);
+                return n - nOther;
             }
+
+            // fall back to string comparison
+            return queryID.CompareToOrdinal(other.queryID);
+        }
+
+        // LUCENENET specific - provide Equals and GetHashCode due to providing operator overrides
+        protected bool Equals(QualityQuery? other) => queryID == other?.queryID;
```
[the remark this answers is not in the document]
paulirwin: That is how Rider generates it, so that subclasses (or any other private callers) don't have to do all the ReferenceEquals/GetType/etc. dance in the `object` version, but I'll remove it for now.

src/Lucene.Net.Suggest/Suggest/Fst/FSTCompletion.cs
```diff
@@ -67,6 +67,42 @@ public int CompareTo(Completion o)
             {
                 return this.Utf8.CompareTo(o.Utf8);
             }
+
+            // LUCENENET specific - per CS0660 and CS0661, we need to override Equals and GetHashCode
+            public override bool Equals(object obj) => ReferenceEquals(this, obj);
+
+            public override int GetHashCode()
+            {
+                unchecked
```
NightOwl888, line 76: Please cascade the call to the base class.

```c#
public override int GetHashCode() => base.GetHashCode();
```

src/Lucene.Net.Suggest/Suggest/Fst/FSTCompletion.cs
```diff
@@ -67,6 +67,42 @@ public int CompareTo(Completion o)
             {
                 return this.Utf8.CompareTo(o.Utf8);
             }
+
+            // LUCENENET specific - per CS0660 and CS0661, we need to override Equals and GetHashCode
+            public override bool Equals(object obj) => ReferenceEquals(this, obj);
```
NightOwl888, line 72: Please cascade the call to the base class.

```c#
public override bool Equals(object obj) => base.Equals(obj);
```

src/Lucene.Net.Suggest/Suggest/Lookup.cs
```diff
@@ -122,14 +122,53 @@ public int CompareTo(LookupResult o)
...
             }
+
+            // LUCENENET specific - per CS0660 and CS0661, we need to override Equals and GetHashCode
+            public override bool Equals(object obj) => ReferenceEquals(this, obj);
```
NightOwl888, line 127: Please cascade the call to the base class.

```c#
public override bool Equals(object obj) => base.Equals(obj);
```

src/Lucene.Net.Suggest/Suggest/Lookup.cs
```diff
@@ -122,14 +122,53 @@ public int CompareTo(LookupResult o)
...
+
+            public override int GetHashCode()
```
NightOwl888, line 129: Please cascade the call to the base class.

```c#
public override int GetHashCode() => base.GetHashCode();
```

src/Lucene.Net/Util/Automaton/State.cs
```diff
@@ -360,17 +360,44 @@ public virtual int CompareTo(State s)
...
 
-        // LUCENENET NOTE: DO NOT IMPLEMENT Equals()!!!
+        // LUCENENET NOTE: DO NOT IMPLEMENT Equals() with structural equality!!!
         // Although it doesn't match GetHashCode(), checking for
         // reference equality is by design.
         // Implementing Equals() causes difficult to diagnose
         // IndexOutOfRangeExceptions when using FuzzyTermsEnum.
         // See GH-296.
+        // Overriding here to prevent CS0660 warning due to defining the == operator.
+        public override bool Equals(object obj) => ReferenceEquals(this, obj);
```
NightOwl888, line 370: Please cascade the call to the base class.

```c#
public override bool Equals(object obj) => base.Equals(obj);
```

NightOwl888 (changes requested):

src/Lucene.Net.Suggest/Suggest/Fst/FSTCompletion.cs
paulirwin: Rider was giving a code analysis warning of having an override that just calls to the base class. If you prefer it that way, I'll suppress that warning for now. I figured it would be better to be explicit about the expectation of the behavior of Equals.

eladmarg:
there is a big performance penalty for using this inside try/catch block.
its much better to use this with TryParse and avoid this block for so hot-path function.

src/Lucene.Net.Highlighter/VectorHighlight/FieldPhraseList.cs
```diff
@@ -470,6 +470,31 @@ public override bool Equals(object obj)
                 return true;
             }
 
+            #region Operator overrides
+            #nullable enable
```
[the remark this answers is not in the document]
NightOwl888: Actually, it is sensible to have `#region` indented because that is the way Visual Studio formats it.

`#nullable` acts like `#pragma` in that they don't need to be paired and don't need to respect other code elements such as curly brackets. These are both compiler directives. `#region` is something that only an IDE understands and must be paired with `#endregion`.

Hopefully, we will eventually move the `#nullable` statements up to the top of the file or to an MSBuild file. It is helpful to have them left aligned to make them stand apart because eventually they may be removed, but we don't want to remove the `#region` when that happens.

NightOwl888 (changes requested):

src/Lucene.Net.Highlighter/VectorHighlight/FieldPhraseList.cs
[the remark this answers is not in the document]
paulirwin: I think this looks rather silly, but updated 😄

src/Lucene.Net.Suggest/Suggest/Lookup.cs
```diff
@@ -124,20 +124,13 @@ public int CompareTo(LookupResult o)
...
 
             // LUCENENET specific - per CS0660 and CS0661, we need to override Equals and GetHashCode
-            public override bool Equals(object obj) => ReferenceEquals(this, obj);
+            // ReSharper disable once BaseObjectEqualsIsObjectEquals
+            // ReSharper disable once RedundantOverriddenMember
+            public override bool Equals(object obj) => base.Equals(obj);
 
-            public override int GetHashCode()
-            {
-                unchecked
```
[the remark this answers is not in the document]
paulirwin: Created #1065

NightOwl888 (approved):

PR #1058, opened 2024-12-05
Overhaul TokenStream ICloseable/IDisposable Patterns, #271

### LINKED ISSUE #271: TokenStream.IncrementToken() is called after Dispose() is called

paulirwin:

Overhaul TokenStream ICloseable/IDisposable Patterns

Fixes #271

## Description

This pull request focuses on refactoring the codebase by replacing the `Dispose` method with the new `ICloseable` interface's `Close` method in all provided TokenStream implementations to make them reusable. None of the implementations in our codebase need a non-reusable cleanup method, however, in discussion in #271, it was determined that there are possible use cases in which a custom TokenStream implementation might need to have their instance Disposed when the owning Analyzer (with stored values set by a ReuseStrategy) is disposed, so we have retained the IDisposable pattern as well. This PR also ensures that those values are disposed by way of DisposableThreadLocal and TokenStreamComponents disposing of their values.

### DIFF
--- a/src/Lucene.Net.Memory/MemoryIndex.cs
+++ b/src/Lucene.Net.Memory/MemoryIndex.cs
@@ -288,7 +288,7 @@ namespace Lucene.Net.Index.Memory
         /// <typeparam name="T">The type of item in the collection.</typeparam>
         /// <remarks>
         /// LUCENENET specific - This class originally got an enumerator in the constructor and stored it to a field
-        /// that was never reset, which meant that it could not be reused (since most IEnumerator implementations can
+        /// that was never reset, which meant that it could not be reused (since many IEnumerator implementations can
         /// only be iterated once and throw on <see cref="System.Collections.IEnumerator.Reset()"/>). This class has
         /// been modified to initialize <see cref="iter"/> on <see cref="TokenStream.Reset()"/> instead, which allows
         /// it to be reused, per the TokenStream workflow contract of allowing <see cref="TokenStream.Reset()"/> after
--- a/src/Lucene.Net.Memory/MemoryIndex.cs
+++ b/src/Lucene.Net.Memory/MemoryIndex.cs
@@ -312,27 +312,16 @@ namespace Lucene.Net.Index.Memory
             }
 
             /// <summary>
-            /// Releases resources used by the <see cref="TokenStreamAnonymousClass{T}"/> and
-            /// if overridden in a derived class, optionally releases unmanaged resources.
+            /// Releases resources used by the <see cref="TokenStreamAnonymousClass{T}"/>.
             /// </summary>
-            /// <param name="disposing"><c>true</c> to release both managed and unmanaged resources;
-            /// <c>false</c> to release only unmanaged resources.</param>
-
-            // LUCENENET specific
-            protected override void Dispose(bool disposing)
+            /// <remarks>
+            /// LUCENENET specific
+            /// </remarks>
+            protected override void DoClose()
             {
-                try
-                {
-                    if (disposing)
-                    {
-                        iter?.Dispose(); // LUCENENET specific - dispose iter and set to null
-                        iter = null;
-                    }
-                }
-                finally
-                {
-                    base.Dispose(disposing);
-                }
+                iter?.Dispose(); // LUCENENET specific - dispose iter and set to null
+                iter = null;
+                base.DoClose();
             }
         }
 
--- a/src/Lucene.Net.Memory/MemoryIndex.cs
+++ b/src/Lucene.Net.Memory/MemoryIndex.cs
@@ -317,11 +317,15 @@ namespace Lucene.Net.Index.Memory
             /// <remarks>
             /// LUCENENET specific
             /// </remarks>
-            protected override void DoClose()
+            protected override void Dispose(bool disposing)
             {
-                iter?.Dispose(); // LUCENENET specific - dispose iter and set to null
-                iter = null;
-                base.DoClose();
+                if (disposing)
+                {
+                    iter?.Dispose(); // LUCENENET specific - dispose iter and set to null, can't be reused
+                    iter = null;
+                }
+
+                base.Dispose(disposing);
             }
         }
 
--- a/src/Lucene.Net.Memory/MemoryIndex.cs
+++ b/src/Lucene.Net.Memory/MemoryIndex.cs
@@ -269,7 +269,8 @@ namespace Lucene.Net.Index.Memory
         /// </summary>
         /// <param name="keywords"> the keywords to generate tokens for </param>
         /// <returns> the corresponding token stream </returns>
-        public virtual TokenStream KeywordTokenStream<T>(ICollection<T> keywords)
+        // LUCENENET: renamed from keywordTokenStream
+        public virtual TokenStream GetKeywordTokenStream<T>(ICollection<T> keywords)
         {
             // TODO: deprecate & move this method into AnalyzerUtil?
             if (keywords is null)
@@ -282,7 +283,7 @@ namespace Lucene.Net.Index.Memory
 
         /// <summary>
         /// An anonymous implementation of <see cref="TokenStream"/> for
-        /// <see cref="KeywordTokenStream{T}(ICollection{T})"/>.
+        /// <see cref="MemoryIndex.GetKeywordTokenStream{T}"/>.
         /// </summary>
         /// <typeparam name="T">The type of item in the collection.</typeparam>
         /// <remarks>
--- a/src/Lucene.Net.Memory/MemoryIndex.cs
+++ b/src/Lucene.Net.Memory/MemoryIndex.cs
@@ -284,16 +284,32 @@ namespace Lucene.Net.Index.Memory
             return new TokenStreamAnonymousClass<T>(keywords);
         }
 
+        /// <summary>
+        /// An anonymous implementation of <see cref="TokenStream"/> for
+        /// <see cref="KeywordTokenStream{T}(ICollection{T})"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of item in the collection.</typeparam>
+        /// <remarks>
+        /// LUCENENET specific - This class originally got an enumerator in the constructor and stored it to a field
+        /// that was never reset, which meant that it could not be reused (since most IEnumerator implementations can
+        /// only be iterated once and throw on <see cref="System.Collections.IEnumerator.Reset()"/>). This class has
+        /// been modified to initialize <see cref="iter"/> on <see cref="TokenStream.Reset()"/> instead, which allows
+        /// it to be reused, per the TokenStream workflow contract of allowing <see cref="TokenStream.Reset()"/> after
+        /// <see cref="TokenStream.Close()"/>.
+        /// </remarks>
         private sealed class TokenStreamAnonymousClass<T> : TokenStream
         {
             public TokenStreamAnonymousClass(ICollection<T> keywords)
             {
-                iter = keywords.GetEnumerator();
+                // LUCENENET specific - initializing iter in Reset() instead of here and storing keywords, see remarks above
+                // iter = keywords.GetEnumerator();
+                this.keywords = keywords;
                 start = 0;
                 termAtt = AddAttribute<ICharTermAttribute>();
                 offsetAtt = AddAttribute<IOffsetAttribute>();
             }
 
+            private ICollection<T> keywords; // LUCENENET specific - see remarks above
             private IEnumerator<T> iter;
             private int start;
             private readonly ICharTermAttribute termAtt;
@@ -301,6 +317,12 @@ namespace Lucene.Net.Index.Memory
 
             public override bool IncrementToken()
             {
+                // LUCENENET specific - check for null iter
+                if (iter is null)
+                {
+                    throw new InvalidOperationException("TokenStream is not properly initialized, IncrementToken() can only be called after Reset()");
+                }
+
                 if (!iter.MoveNext())
                 {
                     return false;
@@ -320,11 +342,21 @@ namespace Lucene.Net.Index.Memory
                 return true;
             }
 
+            // LUCENENET specific - added Close() method to clean up resources
             public override void Close()
             {
-                iter?.Dispose(); // LUCENENET specific - dispose iter and set to null, can't be reused
+                iter?.Dispose();
                 iter = null;
                 base.Close();
+            }
+
+            // LUCENENET specific - added Reset() method to allow reuse of the TokenStream
+            public override void Reset()
+            {
+                iter?.Dispose();
+                iter = keywords.GetEnumerator();
+                start = 0;
+                base.Reset();
             }
         }
 
--- a/src/Lucene.Net.Memory/MemoryIndex.cs
+++ b/src/Lucene.Net.Memory/MemoryIndex.cs
@@ -242,17 +242,26 @@ namespace Lucene.Net.Index.Memory
                 throw new ArgumentNullException(nameof(analyzer), "analyzer must not be null"); // LUCENENET specific - changed from IllegalArgumentException to ArgumentNullException (.NET convention)
             }
 
-            TokenStream stream;
+            // LUCENENET specific: dispose of the TokenStream when done here, instead of in AddField
+            TokenStream stream = null;
             try
             {
-                stream = analyzer.GetTokenStream(fieldName, text);
-            }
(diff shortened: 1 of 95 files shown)


paulirwin:
I feel confident enough in this now that it is ready for review.

This PR achieves the original goals of #271, of not only ensuring that the original issue is fixed, but that we have implemented a design that allows for TokenStreams to be "closed" in a reusable way (via this new `ICloseable` interface to mirror Java) while also still allowing for an extensibility point for them to be "disposed" in a non-reusable way, such as if your TokenStream implementation needs to dispose of native resources or file handles. 

Additionally, the code in BaseTokenStreamTestCase has been reviewed several times and is now much closer to the original implementation. I've left the try/finally calls, since they probably should have been in the original Java anyways, and it seems to possibly help cascading failures (although I haven't been able to reproduce those, possibly as a result of other previous fixes).

With all tests passing reliably, this design change implemented, and all concerns addressed in the original issue, I think it's time to wrap this up and call this 5+ year old issue complete.

src/Lucene.Net/Analysis/TokenStream.cs
```diff
@@ -36,13 +37,13 @@ namespace Lucene.Net.Analysis
     /// A new <see cref="TokenStream"/> API has been introduced with Lucene 2.9. this API
     /// has moved from being <see cref="Token"/>-based to <see cref="Util.IAttribute"/>-based. While
     /// <see cref="Token"/> still exists in 2.9 as a convenience class, the preferred way
-    /// to store the information of a <see cref="Token"/> is to use <see cref="Attribute"/>s.
+    /// to store the information of a <see cref="Token"/> is to use <see cref="System.Attribute"/>s.
```
NightOwl888, line 40: This reference is now incorrect. It should be `Lucene.Net.Util.Attribute`.

src/Lucene.Net/Analysis/TokenStream.cs
```diff
@@ -36,13 +37,13 @@ namespace Lucene.Net.Analysis
...
     /// <para/>
     /// <see cref="TokenStream"/> now extends <see cref="AttributeSource"/>, which provides
     /// access to all of the token <see cref="Util.IAttribute"/>s for the <see cref="TokenStream"/>.
-    /// Note that only one instance per <see cref="Attribute"/> is created and reused
+    /// Note that only one instance per <see cref="System.Attribute"/> is created and reused
```
NightOwl888, line 44: This reference is now incorrect. It should be `Lucene.Net.Util.Attribute`.

src/Lucene.Net/Analysis/TokenStream.cs
```diff
@@ -36,13 +37,13 @@ namespace Lucene.Net.Analysis
...
     /// for every token. This approach reduces object creation and allows local
-    /// caching of references to the <see cref="Attribute"/>s. See
+    /// caching of references to the <see cref="System.Attribute"/>s. See
```
NightOwl888, line 46: This reference is now incorrect. It should be `Lucene.Net.Util.Attribute`.

src/Lucene.Net/Analysis/Tokenizer.cs
```diff
@@ -57,29 +59,16 @@ protected Tokenizer(AttributeFactory factory, TextReader input)
...
         /// <summary>
-        /// Releases resources associated with this stream.
-        /// <para/>
-        /// If you override this method, always call <c>base.Dispose(disposing)</c>, otherwise
-        /// some internal state will not be correctly reset (e.g., <see cref="Tokenizer"/> will
-        /// throw <see cref="InvalidOperationException"/> on reuse).
+        /// <inheritdoc cref="TokenStream.Close()"/>
         /// </summary>
-        /// <remarks>
-        /// <b>NOTE:</b>
```
NightOwl888, line 67: Please add this note back, as was the case upstream.

https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/analysis/Tokenizer.java#L60-L62

NightOwl888 (changes requested):
Thanks for the PR.

Sorry it has been open for so long. This one required a significant block of time to go through.

I changed the `TokenStream`, `TokenStreamComponents`, and `DisposableThreadLocal` classes to allow `IDisposable` to be optionally implemented. This was done using a `DisposableTracker` class that is shared between all of the `TokenStream` instances. The `LinkedList<IDisposable>` is only instantiated if one or more of the `TokenStream`s implements `IDisposable`, otherwise it is `null`. Sharing it also allows a single call in `TokenStreamComponents` to clean it up (using `m_source`, since it is always set). `IOUtils.Dispose()` was used consistently to ensure all `Dispose()` methods are called without swallowing any exceptions that are thrown.

I noticed in later versions of Lucene, the `.close()` method is called in the tests that use a shared instance. But do note that virtually all test classes that use static fields rather than instance fields are likely doing it wrong. JUnit instantiates a new test class instance for each test, which is why they are doing it that way. Not only do we not need to put shared state in static fields, but doing so increases the memory footprint of the tests (although, there may be a benefit to not allowing the GC to collect these instances).

There are some other things I noticed, most of them very minor fixes. See the inline comments.

src/Lucene.Net/Analysis/TokenStream.cs
```diff
@@ -108,8 +130,17 @@ protected TokenStream(AttributeFactory factory)
...
+
+            // LUCENENET specific - track disposable TokenStreams that are part of the current chain
+            disposableTracker = new DisposableTracker(This);
         }
 
+        /// <summary>
+        /// LUCENENET: We cannot access <c>this</c> from the constructor, so this property is used
+        /// to cheat the compiler.
+        /// </summary>
+        private TokenStream This => this;
```
paulirwin, line 142: Are you sure this (no pun intended) is needed? I was curious when I saw it, and tried it out and the compiler doesn't have an issue with using `this` from the constructor. Or is there some non-compiler runtime failure doing that?

src/Lucene.Net/Analysis/TokenStream.cs
paulirwin: This was added by Rider as part of a refactoring, I should have caught it before submitting the PR. However, note that not only is it incorrect _now_, it was _previously_ incorrect as well, which is why Rider fully qualified the name here. You can see this yourself at the beta 17 docs, where that link goes to MSDN for `System.Attribute`: https://lucenenet.apache.org/docs/4.8.0-beta00017/api/core/Lucene.Net.Analysis.TokenStream.html

This problem has been there since the beta00005 docs! If it weren't for this accidental refactoring, we might not have caught it.

src/Lucene.Net/Analysis/TokenStream.cs
NightOwl888: If `this` works in the constructor, I would rather use it that way. When I tried it, it didn't compile, but perhaps I had another issue that was unrelated.

src/Lucene.Net.Memory/MemoryIndex.cs
```diff
@@ -275,23 +284,45 @@ public virtual TokenStream KeywordTokenStream<T>(ICollection<T> keywords)
...
 
+        /// <summary>
+        /// An anonymous implementation of <see cref="TokenStream"/> for
+        /// <see cref="KeywordTokenStream{T}(ICollection{T})"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of item in the collection.</typeparam>
+        /// <remarks>
+        /// LUCENENET specific - This class originally got an enumerator in the constructor and stored it to a field
+        /// that was never reset, which meant that it could not be reused (since most IEnumerator implementations can
+        /// only be iterated once and throw on <see cref="System.Collections.IEnumerator.Reset()"/>). This class has
```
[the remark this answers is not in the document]
paulirwin: First, I should have worded this differently; I should have said "many" instead of "most." I meant to communicate that we generally can't rely on using it in Reset, because many implementations do throw. For example, any enumerators created by LINQ or the Enumerable class throw on Reset, which are pretty commonly used. Note that [Microsoft's own docs on IEnumerator](https://learn.microsoft.com/en-us/dotnet/api/system.collections.ienumerator?view=net-9.0#remarks) say that Reset "is provided for COM interoperability and does not need to be fully implemented; instead, the implementer can throw a [NotSupportedException](https://learn.microsoft.com/en-us/dotnet/api/system.notsupportedexception?view=net-9.0)." I think it's likely enough that it will throw that alternative 1 is not a good idea.

Alternative 3 seems like it would be more costly than just allowing it to possibly allocate on reset. Enumerators for BCL types are highly optimized nowadays, and for some of the common ones you'd at most have a boxed struct. It seems to me like having try/catch unwinding would be more costly in runtime performance.

Alternative 2 is fine, if you really think we need it. I don't think we do. It is quite common to get and dispose of an enumerator in .NET. There are over 2,000 uses of `foreach` in our codebase which get a new enumerator each time that I doubt we are applying such scrutiny to. But, if you think it's important, I'd choose this option.

src/Lucene.Net/Analysis/Tokenizer.cs
paulirwin: I had removed this because it seemed duplicated from the inherited TokenStream docs, but it is subtly different, so I'm adding it back even though it is a little repetitive.

paulirwin:
@NightOwl888 I really like your changes to decouple TokenStream from IDisposable. I think this will be a great change that allows those that need non-reusable-disposable TokenStreams to do so without carrying that baggage all the way through the virtual call hierarchy of our many TokenStream implementations that don't need it.

src/Lucene.Net.Memory/MemoryIndex.cs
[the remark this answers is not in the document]
NightOwl888: Good points. I wish Microsoft made an `IResettable` so detection would be easy.

Given that this is a small single instance rather than an array and Lucene didn't even think to make it reusable in the first place, I think we can leave this as-is (although, let's fix the comments).

src/Lucene.Net.Spatial/Prefix/PrefixTreeStrategy.cs
```diff
@@ -157,17 +158,28 @@ public virtual Field[] CreateIndexableFields(IShape? shape, double distErr)
             IndexOptions = IndexOptions.DOCS_ONLY
         }.Freeze();
 
-        /// <summary>Outputs the tokenString of a cell, and if its a leaf, outputs it again with the leaf byte.</summary>
+        /// <summary>
+        /// Outputs the tokenString of a cell, and if its a leaf, outputs it again with the leaf byte.
+        /// </summary>
+        /// <remarks>
+        /// LUCENENET specific - This class originally took an enumerator, which meant that it could not
+        /// be reused (since most IEnumerator implementations can only be iterated once and throw on
```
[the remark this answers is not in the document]
NightOwl888: On another pass of this, I noticed that the `CellTokenStream` is internal. So, the question here is whether or not the `Field[]` returned from `CreateIndexableFields` is reusable. It is passed a `JCG.List<T>` from `SpatialPrefixTree.GetCells()`, which supports `Reset()`. The only way to change that is to create a custom subclass of `SpatialPrefixTree` to override it. It seems like a minor thing to restrict this to only implementations of `IEnumerable<T>` that support `Reset()` here, given that the user would have to go out of their way to change it to something that does not.

What's the downside? The user passes an `IList<T>` that doesn't support `Reset()` and it fails upon reuse. But in Lucene it would have failed upon reuse, anyway.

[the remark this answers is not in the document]
paulirwin: I also kept the lazy-initialization logic, as it helps keep it cleaner IMO, as well as supports `Close()` by disposing of the enumerator but in a reusable way. I think it makes sense to reallocate a new enumerator if you close that token stream and then reset it.

src/Lucene.Net.Spatial/Prefix/Tree/SpatialPrefixTree.cs
```diff
@@ -178,6 +178,11 @@ protected internal virtual Cell GetCell(IPoint p, int level)
         /// <para/>
         /// This implementation checks if shape is a <see cref="IPoint"/> and if so returns
         /// <see cref="GetCells(IPoint, int, bool)"/>.
+        /// <para />
+        /// LUCENENET NOTE: If overriding this method, make sure to return an implementation of
+        /// <c>IList&lt;Cell&gt;</c> whose <see cref="IList{T}.GetEnumerator()"/> returns an enumerator
+        /// that supports being <see cref="IEnumerator{T}.Reset"/> without throwing an exception.
```
NightOwl888, line 184: `Reset()` is on `IEnumerator`, not `IEnumerator<T>`. This XML doc doesn't compile.

NightOwl888 (changes requested):

src/Lucene.Net.Spatial/Prefix/PrefixTreeStrategy.cs
```diff
@@ -159,14 +159,23 @@ public virtual Field[] CreateIndexableFields(IShape? shape, double distErr)
...
-        /// workflow contract of allowing <see cref="TokenStream.Reset()"/> after <see cref="TokenStream.Close()"/>.
+        /// be reused after a call to <see cref="TokenStream.Close()"/> (which was not present in the Lucene code).
+        /// This class has been modified to take an <c>IEnumerable&lt;Cell&gt;</c> instead, which allows it to be
+        /// reused, per the TokenStream workflow contract of allowing <see cref="TokenStream.Reset()"/>
+        /// after <see cref="TokenStream.Close()"/>. To accomplish this, it takes an enumerable and lazy-initializes
+        /// its enumerator upon the first call to <see cref="TokenStream.Reset()"/>. Calling
+        /// <see cref="TokenStream.Close()"/> will dispose of the enumerator, but it can be reinitialized by calling
+        /// <see cref="TokenStream.Reset()"/> again.
+        /// <para />
+        /// This implementation also expects that the enumerator will support <see cref="IEnumerator{T}.Reset()"/>
```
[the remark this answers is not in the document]
paulirwin: Hm, I'm guessing you're referring to the docfx build. MSBuild doesn't have an issue with it since it's inherited, nor does Rider. I'll change it though.

NightOwl888:
> @NightOwl888 I really like your changes to decouple TokenStream from IDisposable. I think this will be a great change that allows those that need non-reusable-disposable TokenStreams to do so without carrying that baggage all the way through the virtual call hierarchy of our many TokenStream implementations that don't need it.

Yeah. I thought about doing it this way on `Close()` also, but that diverges from Lucene quite a bit.

Also, I noticed that `Close()` is executing the inner class prior to the current class, the same order as inherited constructors in Java. `Dispose()` is set up to execute the current class first followed by the inner class, which is what would be expected in .NET. It seems wrong to make the order of execution different between `Close()` and `Dispose()`, but it isn't clear which is the more sensible option. We could potentially break something important by changing `Close()`.

src/Lucene.Net.Spatial/Prefix/PrefixTreeStrategy.cs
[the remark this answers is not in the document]
paulirwin: Nevermind, I see the warning now. We should probably make those warnings be errors.

NightOwl888 (approved):
I went through the items in https://github.com/apache/lucenenet/issues/271#issuecomment-1306912515, and this satisfies everything except for the Roslyn code analyzers to verify that `base.Reset()`, `base.End()` and `base.Close()` are called. We should probably also have analyzers to determine whether these are called on `m_input`, as well. I checked and I didn't see an issue for these. So, if we open a new issue for the Roslyn code analyzers (which should be `Lucene` analyzers, not `LuceneDev` analyzers), then this is good to merge.

paulirwin:
Split out analyzers as #1146. Thanks for the collaboration and review on this big PR that fixes a longstanding issue!

PR #1060, opened 2024-12-05
SWEEP: Avoid unnecessary FileInfo/DirectoryInfo allocations, #832

### LINKED ISSUE #832: SWEEP: Add string overloads to eliminate unnecessary allocations of FileInfo and DirectoryInfo objects
In Java it is common to use the `File` or `Path` object as a parameter to methods, as they are accepted as parameters of stream constructors and many other places.

However, in .NET, the corresponding objects (`TextWriter`, `Stream`, `StreamWriter`, etc) don't accept a `FileInfo` as a parameter, they accept strings. As such, we are commonly newing up these `FileInfo` objects, passing them through our API and reading the original string back as `FileInfo.FullName`. The `FileInfo` object is not even used in many cases.

To keep the API the same as Lucene, we should keep the overloads that accept `FileInfo`, but add overloads that are identical in every way except that they accept a `string` for a file name instead of a `FileInfo`. The `FileInfo` overloads can then just pass `FileInfo.FullName` to the string overloads.

For example, we currently have

```c#
        /// <summary>
        /// Returns an <see cref="Stream"/> over the requested file, identifying
        /// the appropriate <see cref="Stream"/> instance similar to <see cref="GetInputStream(FileInfo)"/>.
        /// </summary>
        public static Stream GetOutputStream(FileInfo file)
        {
            // First, create a FileInputStream, as this will be required by all types.
            // Wrap with BufferedInputStream for better performance
            Stream os = new FileStream(file.FullName, FileMode.Create, FileAccess.ReadWrite, FileShare.ReadWrite);
            return GetFileType(file).GetOutputStream(os);
        }
```

And we should change it to

```c#
        /// <summary>
        /// Returns an <see cref="Stream"/> over the requested <paramref name="file"/>, identifying
        /// the appropriate <see cref="Stream"/> instance similar to <see cref="GetInputStream(FileInfo)"/>.
        /// </summary>
        public static Stream GetOutputStream(FileInfo file)
            => GetOutputStream(file?.FullName);

        /// <summary>
        /// Returns an <see cref="Stream"/> over the requested <paramref name="filePath"/>, identifying
        /// the appropriate <see cref="Stream"/> instance similar to <see cref="GetInputStream(string)"/>.
        /// </summary>
        public static Stream GetOutputStream(string filePath)
        {
            // First, create a FileInputStream, as this will be required by all types.
            // Wrap with BufferedInputStream for better performance
            Stream os = new FileStream(filePath, FileMode.Create, FileAccess.ReadWrite, FileShare.ReadWrite);
            return GetFileType(filePath).GetOutputStream(os);
        }
```

There are several places throughout the project where we are instantiating and then discarding `FileInfo` objects without even making use of their functionality ([example](https://github.com/apache/lucenenet/blob/87add0272d1da1eaee68ca95e37841e90ec07346/src/Lucene.Net/Store/MMapDirectory.cs#L185-L186)). Some could instead use static methods (i.e. `File.Exists`). In some other cases, these `FileInfo` and `DirectoryInfo` instances are appropriate.

paulirwin:

Add string-based overloads to some methods that take FileInfo/DirectoryInfo to avoid unnecessary allocations

Fixes #832

## Description

See #832 for context. This makes changes to add and prefer string-based overloads for methods that take FileInfo and DirectoryInfo primarily just for the FullName property, to avoid unnecessary allocations.

### DIFF
--- a/Lucene.Net.sln.DotSettings
+++ b/Lucene.Net.sln.DotSettings
@@ -1,4 +1,6 @@
 ﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Coord/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=LUCENENET/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=stopword/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=stopwords/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=testsettings/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
--- a/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/HyphenationTree.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/HyphenationTree.cs
@@ -127,9 +127,7 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
         /// <param name="filename"> the filename </param>
         /// <exception cref="IOException"> In case the parsing fails </exception>
         public virtual void LoadPatterns(string filename)
-        {
-            LoadPatterns(filename, Encoding.UTF8);
-        }
+            => LoadPatterns(filename, Encoding.UTF8);
 
         /// <summary>
         /// Read hyphenation patterns from an XML file.
@@ -149,9 +147,7 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
         /// <param name="f"> a <see cref="FileInfo"/> object representing the file </param>
         /// <exception cref="IOException"> In case the parsing fails </exception>
         public virtual void LoadPatterns(FileInfo f)
-        {
-            LoadPatterns(f, Encoding.UTF8);
-        }
+            => LoadPatterns(f.FullName, Encoding.UTF8);
 
         /// <summary>
         /// Read hyphenation patterns from an XML file.
@@ -160,10 +156,7 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
         /// <param name="encoding">The character encoding to use</param>
         /// <exception cref="IOException"> In case the parsing fails </exception>
         public virtual void LoadPatterns(FileInfo f, Encoding encoding)
-        {
-            var src = new FileStream(f.FullName, FileMode.Open, FileAccess.Read);
-            LoadPatterns(src, encoding);
-        }
+            => LoadPatterns(f.FullName, encoding);
 
         /// <summary>
         /// Read hyphenation patterns from an XML file.
@@ -171,9 +164,7 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
         /// <param name="source"> <see cref="Stream"/> input source for the file </param>
         /// <exception cref="IOException"> In case the parsing fails </exception>
         public virtual void LoadPatterns(Stream source)
-        {
-            LoadPatterns(source, Encoding.UTF8);
-        }
+            => LoadPatterns(source, Encoding.UTF8);
 
         /// <summary>
         /// Read hyphenation patterns from an XML file.
--- a/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/PatternParser.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/PatternParser.cs
@@ -17,9 +17,9 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
      * The ASF licenses this file to You under the Apache License, Version 2.0
      * (the "License"); you may not use this file except in compliance with
      * the License.  You may obtain a copy of the License at
-     * 
+     *
      *      http://www.apache.org/licenses/LICENSE-2.0
-     * 
+     *
      * Unless required by applicable law or agreed to in writing, software
      * distributed under the License is distributed on an "AS IS" BASIS,
      * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -62,7 +62,7 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
             hyphenChar = '-'; // default
         }
 
-        public PatternParser(IPatternConsumer consumer) 
+        public PatternParser(IPatternConsumer consumer)
             : this()
         {
             this.consumer = consumer;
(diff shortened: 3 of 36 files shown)


NightOwl888 (commented):
On our `DirectoryInfo` and `FileInfo` usage, we consistently use:

```c#
... = dir.FullPath
``
and 
``c#
... = file.FullPath
```
In both cases, this means the string that is being passed in by the user is being resolved first using `Path.GetFullPath(path)`, a static method.

https://github.com/dotnet/runtime/blob/v9.0.8/src/libraries/System.Private.CoreLib/src/System/IO/FileInfo.cs#L28
https://github.com/dotnet/runtime/blob/v9.0.8/src/libraries/System.Private.CoreLib/src/System/IO/DirectoryInfo.cs#L34

`Path.GetFullPath()` will resolve any relative path that is passed in by the user to an absolute path.

In both cases, there is an `isNormalized` field that gets passed around with this type so the  `GetFullPath()` method is never called again. So, just putting a `GetFullPath()` call in doesn't get us exactly the same behavior because the path may end up cascading through multiple public APIs and getting called multiple times (which is the incorrect behavor on Unix). So, it is actually better to leave these `GetFullPath()` calls off unless we address that issue everywhere.

Now, whether that is actually a problem is up for debate. On one hand, .NET users of System.IO types should already be aware that full paths must be resolved prior to passing them to `FileStream`, etc. to get them to resolve correctly. On the other hand these new APIs will act inconsistently with the existing `DirectoryInfo` and `FileInfo` overloads.

Thoughts?

NightOwl888:
I started a review, but got into the weeds a bit because I realized that the current port doesn't match the behavior of Java (at least according to ChatGPT - unconfirmed). The new string overload behavior should match Java, which is to pass the relative path down to the OS where it is resolved relative to the current application (the place where the jar lives). On the other hand, with `FileInfo` and `DirectoryInfo`, relative paths are resolved eagerly and not passed down to the OS. So, when we call `FullName` we don't necessarily have the same behavior.

Unfortunately, `FileInfo` and `DirectoryInfo` have no way to retrieve the original string that was passed into the constructor. So, there doesn't seem to be a reasonable fix,. Well, there is a protected `OriginalString` property that we could get to using the [`UnsafeAccessorAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute) on `net8.0`+, but we are probably better off accepting the differences of calling `FullName` in most cases.

Also, I suspect but haven't yet confirmed that the relative path behavior of our command line tools is broken. AFAIK, the expected behavior in a .NET console app is for a relative path to be resolved to the **environment's** current directory, but the behavior we have is relative to the directory of the executable. This behavior might be altered by wrapping these commands into a dotnet tool or by our (3rd party) wrapper project, though.

One other thing I noticed is that the `FileInfo.GetCanonicalPath()` method was not included here, probably because you didn't think to look for the common base class of `FileInfo` and `DirectoryInfo`: [`FileSystemInfo`](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo). This is the only place outside of the test framework and one test where we use it, though.

I will look into this more tomorrow. But we will need a clear idea of the *expected* behavior to get through this review.

paulirwin:
> I started a review, but got into the weeds a bit because I realized that the current port doesn't match the behavior of Java (at least according to ChatGPT - unconfirmed). The new string overload behavior should match Java, which is to pass the relative path down to the OS where it is resolved relative to the current application (the place where the jar lives).

I have analyzed this and determined that it largely behaves as you specify here: relative paths just work when passed down to the filesystem. Almost all .NET BCL methods that we use have predictable, correct behavior when relative paths are used. There is one exception, noted below.

| Function | Works with relative path? |
| --- | --- |
| Directory.Exists | Yes |
| FileStream constructor | Yes |
| Directory.GetParent | Yes |
| Directory.EnumerateFiles | Yes |
| File.Exists | Yes |
| FileInfo constructor | Yes |
| DirectoryInfo constructor | Yes |
| Path.GetDirectoryName | Somewhat* |
| Path.GetFileName | Yes |
| Directory.Delete | Yes |
| File.Delete | Yes |
| File.WriteAllText | Yes |
| Directory.CreateDirectory | Yes |

`Path.GetDirectoryName` (used here in a few places like Kuromoji, Benchmarks, and the CLI) does not itself error on a relative path, and works arguably correctly, but it returns a value that might not work correctly depending on how it was used previously. The function returns the directory component of a string containing a path, and does not resolve the directory of the path first. This means that for input values like `"."` and `".."`, it returns empty string (because those could be interpreted as the "file" component of a relative path, even though they have a special meaning). Then, for inputs like `"./foo.bar"`, it returns `"."`, because that is the directory component of the relative path. Methods like `WriteEnwikiLineDocTask.CategoriesLineFile` are fine with this behavior, because it's expected that you're passing in a full file name (and not a directory name). The only time when this is a problem is if you pass in those special relative directory names without a file name, which could happen in CompoundFileExtractor. So I've fixed that by reverting that change and leaving a comment.

Note that only the Path class methods can have a problem, because the others actually hit the filesystem where the paths are resolved. Path just operates on strings without caring whether the path exists on disk. `Path.GetFileName` is not a problem because even if it's a completely slash-less path like `"foo.bar"` or any form of relative path to a file, the file name will extract correctly.

Note also that `~` for a home directory shorthand does not work correctly in most BCL types, including FileInfo and DirectoryInfo anyways, so I do not think we need to worry about that. 

> On the other hand, with `FileInfo` and `DirectoryInfo`, relative paths are resolved eagerly and not passed down to the OS. So, when we call `FullName` we don't necessarily have the same behavior.

As noted above, this is fine in the cases in this PR.

> Unfortunately, `FileInfo` and `DirectoryInfo` have no way to retrieve the original string that was passed into the constructor. So, there doesn't seem to be a reasonable fix,. Well, there is a protected `OriginalString` property that we could get to using the [`UnsafeAccessorAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.unsafeaccessorattribute) on `net8.0`+, but we are probably better off accepting the differences of calling `FullName` in most cases.

Yeah, I don't think we need to worry about this, since relative paths work for I/O.

> Also, I suspect but haven't yet confirmed that the relative path behavior of our command line tools is broken. AFAIK, the expected behavior in a .NET console app is for a relative path to be resolved to the **environment's** current directory, but the behavior we have is relative to the directory of the executable. This behavior might be altered by wrapping these commands into a dotnet tool or by our (3rd party) wrapper project, though.

I'm not sure we need to worry about this unless someone reports it. Seems like worst case you'd get a file/directory not found error, fix it, and move on.

> One other thing I noticed is that the `FileInfo.GetCanonicalPath()` method was not included here, probably because you didn't think to look for the common base class of `FileInfo` and `DirectoryInfo`: [`FileSystemInfo`](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo). This is the only place outside of the test framework and one test where we use it, though.

I did see this, but I don't think it's in scope (at least not for having a string-based overload). The method docs say that it returns the path "with all references resolved" and the easiest way to do that is to allocate a FileInfo/DirectoryInfo and calling `FullName` like it does anyways. 

I'm not 100% convinced that this method truly needs to exist, but I am not going to worry about that at this time. (For example, using the lowercase path `"c:\\windows"` works fine in .NET operations with the lowercase drive letter.)

Now, there might be some possible analysis that could be done on whether or not code that _calls_ this method could avoid canonicalization and thus avoid FileInfo/DirectoryInfo... but that would be beyond the scope of just reducing allocations, as it possibly would result in non-canonical paths being used (assuming that is even important). I am not aware of a way to resolve a path without hitting the disk via I/O, and if you're doing that, the allocation of a FileInfo/DirectoryInfo is pretty negligible. The goal of this PR is only to help reduce allocations for cases where resolving the path is not necessary.

Also, it should be noted, _the original overloads are still there_. So if someone has a particular reason for using FileInfo/DirectoryInfo, they still can. This is not a breaking-change PR. So any new code that tries to use the string versions will need to confirm that it works for them with relative paths (it probably will), or just use an absolute path.

That said, I can add more unit tests to confirm this beyond my analysis and testing of the BCL. So I'll do that for what I can. Converting back to draft until I get that done. Feel free to review, though.

paulirwin:
@NightOwl888 Added unit tests to ensure relative paths work, as well as general testing of these string-based overloads.

PR #1061, opened 2024-12-06
SWEEP: Make anonymous class members default to private instead of internal, #715

### LINKED ISSUE #715: All methods, fields, and properties should be marked `private` instead of `internal` in all anonymous classes
Anonymous classes are suffixed `AnonymousClass`.

In Java, [anonymous classes](https://docs.oracle.com/javase/tutorial/java/javaOO/anonymousclasses.html) are declared inline, so there is no other code that can interact with them.

However, many of these classes (such as [this example](https://github.com/apache/lucenenet/blob/b6cf28b78c85f33e66d89abe399c4523dd41dc80/src/Lucene.Net.Misc/Index/Sorter/BlockJoinComparatorSource.cs#L102)) have methods, fields, and properties that are marked `internal`. `internal` was chosen because in Lucene it is common to leave off an accessibility modifier.

- In Java, if there is no accessibility modifier, the accessibility defaults to `package-private`, which is similar to `internal`.
- In .NET, if there is no accessibility modifier, the accessibility defaults to `private`.

However, in an anonymous class, `package-private` accessibility is not sensible. It has the same effect as declaring the member `private`.

Therefore, all methods, fields, and properties of anonymous classes can be changed to `private` and we should leave a comment indicating that the change was made.

```// LUCENENET: changed accessibility from internal to private```

There may be exceptions in cases where members were added to make testing easier, so those may need to remain `internal`. A compile can quickly identify members that should be left alone.

### PR DESCRIPTION
**Pull request by paulirwin**

Make anonymous class members default to private instead of internal

Fixes #715

## Description

See #715 for info and rationale. This makes "AnonymousClass" implementation members private instead of internal where possible (and readonly, where possible). This also takes the opportunity to remove redundant constructors and unused "outerInstance" fields where applicable.

### DIFF
--- a/src/Lucene.Net/Search/NumericRangeQuery.cs
+++ b/src/Lucene.Net/Search/NumericRangeQuery.cs
@@ -430,7 +430,7 @@ namespace Lucene.Net.Search
                     this.outerInstance = outerInstance;
                 }
 
-                public override sealed void AddRange(BytesRef minPrefixCoded, BytesRef maxPrefixCoded)
+                public override void AddRange(BytesRef minPrefixCoded, BytesRef maxPrefixCoded)
                 {
                     outerInstance.rangeBounds.Enqueue(minPrefixCoded);
                     outerInstance.rangeBounds.Enqueue(maxPrefixCoded);
@@ -446,7 +446,7 @@ namespace Lucene.Net.Search
                     this.outerInstance = outerInstance;
                 }
 
-                public override sealed void AddRange(BytesRef minPrefixCoded, BytesRef maxPrefixCoded)
+                public override void AddRange(BytesRef minPrefixCoded, BytesRef maxPrefixCoded)
                 {
                     outerInstance.rangeBounds.Enqueue(minPrefixCoded);
                     outerInstance.rangeBounds.Enqueue(maxPrefixCoded);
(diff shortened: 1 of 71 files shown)

### CODE REVIEW

File: src/Lucene.Net/Search/NumericRangeQuery.cs [hunk 1]
```diff
@@ -446,7 +446,7 @@ public Int32RangeBuilderAnonymousClass(NumericRangeTermsEnum outerInstance)
                     this.outerInstance = outerInstance;
                 }
 
-                public override sealed void AddRange(BytesRef minPrefixCoded, BytesRef maxPrefixCoded)
+                public override void AddRange(BytesRef minPrefixCoded, BytesRef maxPrefixCoded)
```
NightOwl888, line 449: `sealed` is how this is declared upstream. I realize it makes no difference (it isn't possible to inherit an anonymous class), but let's keep the declaration in sync.

https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/search/NumericRangeQuery.java#L472

src/Lucene.Net/Search/NumericRangeQuery.cs
```diff
@@ -430,7 +430,7 @@ public Int64RangeBuilderAnonymousClass(NumericRangeTermsEnum outerInstance)
...
```
NightOwl888, line 433: `sealed` is how this is declared upstream. I realize it makes no difference (it isn't possible to inherit an anonymous class), but let's keep the declaration in sync.

https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/search/NumericRangeQuery.java#L432

NightOwl888 (changes requested):

src/Lucene.Net/Search/NumericRangeQuery.cs
paulirwin: This was removed because it's redundant because the class is marked `sealed`.

NightOwl888 (approved):

PR #1064, opened 2024-12-12
Convert Lucene volatile long/double values to Atomic counterparts, #1063

### LINKED ISSUE #1063: Review for volatile 64-bit compatibility

### Task description

In C#, `volatile` is not an option on 64-bit types, such as `long` and `double`, however the following fields were marked as `volatile` in Lucene. In most cases, we simply removed the `volatile` keyword from the declaration without making them thread safe, but we should review to see whether these should be fixed.

- [ ] [org.apache.lucene.index.IndexWriter.changeCount](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L225)
- [ ] [org.apache.lucene.index.IndexWriter.lastCommitChangeCount](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L226)
- [ ] [org.apache.lucene.index.IndexWriter.pendingCommitChangeCount](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L231)
- [ ] [org.apache.lucene.index.LiveIndexWriterConfig.writeLockTimeout](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/index/LiveIndexWriterConfig.java#L68)
- [ ] [org.apache.lucene.index.LiveIndexWriterConfig.ramBufferSizeMB](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/core/src/java/org/apache/lucene/index/LiveIndexWriterConfig.java#L42)
- [ ] [org.apache.lucene.index.MergePolicy.OneMerge.estimatedMergeBytes](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java#L102)
- [ ] [org.apache.lucene.index.MergePolicy.OneMerge.totalMergeBytes](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java#L105)
- [ ] [org.apache.lucene.index.SegmentCommitInfo.sizeInBytes](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/index/SegmentCommitInfo.java#L61)
- [ ] [org.apache.lucene.search.ControlledRealTimeReopenThread.waitingGen](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/search/ControlledRealTimeReopenThread.java#L49)
- [ ] [org.apache.lucene.search.ControlledRealTimeReopenThread.searchingGen](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/search/ControlledRealTimeReopenThread.java#L50)
- [ ] [org.apache.lucene.search.TimeLimitingCollector.TimerThread.time](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/search/TimeLimitingCollector.java#L245)
- [ ] [org.apache.lucene.search.TimeLimitingCollector.TimerThread.resolution](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/search/TimeLimitingCollector.java#L247)
- [ ] [org.apache.lucene.store.RateLimiter.SimpleRateLimiter.lastNS](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/store/RateLimiter.java#L53)
- [ ] [org.apache.lucene.store.RateLimiter.SimpleRateLimiter.mbPerSec](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/store/RateLimiter.java#L51)
- [ ] [org.apache.lucene.store.RateLimiter.SimpleRateLimiter.nsPerByte](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/core/src/java/org/apache/lucene/store/RateLimiter.java#L52) (Note: this doesn't exist in 4.8.1 and is replaced with [`volatile long minPauseCheckBytes`](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/core/src/java/org/apache/lucene/store/RateLimiter.java#L59))
- [ ] [org.apache.lucene.replicator.LocalReplicator.ReplicationSession.lastAccessTime](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.0/lucene/replicator/src/java/org/apache/lucene/replicator/LocalReplicator.java#L85)

### Options

1. Use `AtomicInt64` for the `long` type
2. Use `Interlocked` for the `long` type
3. Use a local lock object

For `double` variables, we can convert to/from `long` using `J2N.BitConversion.Int64BitsToDouble` and `J2N.BitConversion.DoubleToRawInt64Bits`, making the backing field a `long` that uses `Interlocked` or `AtomicInt64`. Alternatively, we could wrap these operations in a new `AtomicDouble` class in J2N.

`Interlocked` would be the preferred option, since it can be used without adding heap allocations and does not use any locking, similar to `volatile`.

paulirwin:

Convert long/double values that are volatile in Lucene (but can't be in .NET) to AtomicInt64 or (new) AtomicDouble.

Fixes #1063

## Description

Since we can't use `volatile` on 64-bit values in .NET, this doesn't work for `long` or `double`. This PR changes `volatile long` fields to be `AtomicInt64` from J2N, as well as introduces a new internal `AtomicDouble` type in Support for those values, using bit conversion to/from long internally so that we can use Interlocked on it.

### DIFF
--- a/src/Lucene.Net/Index/IndexWriter.cs
+++ b/src/Lucene.Net/Index/IndexWriter.cs
@@ -228,13 +228,14 @@ namespace Lucene.Net.Index
         private readonly Directory directory; // where this index resides
         private readonly Analyzer analyzer; // how to analyze text
 
-        private long changeCount; // increments every time a change is completed
-        private long lastCommitChangeCount; // last changeCount that was committed
+        // LUCENENET specific - since we don't have `volatile long` in .NET, we use AtomicInt64
+        private readonly AtomicInt64 changeCount = new AtomicInt64(); // increments every time a change is completed
+        private readonly AtomicInt64 lastCommitChangeCount = new AtomicInt64(); // last changeCount that was committed
 
         private IList<SegmentCommitInfo> rollbackSegments; // list of segmentInfo we will fallback to if the commit fails
 
         internal volatile SegmentInfos pendingCommit; // set when a commit is pending (after prepareCommit() & before commit())
-        internal long pendingCommitChangeCount;
+        internal readonly AtomicInt64 pendingCommitChangeCount = new AtomicInt64(); // LUCENENET specific - since we don't have `volatile long` in .NET, we use AtomicInt64
 
         private ICollection<string> filesToCommit;
 
@@ -2197,7 +2198,7 @@ namespace Lucene.Net.Index
                 // could close, re-open and re-return the same segment
                 // name that was previously returned which can cause
                 // problems at least with ConcurrentMergeScheduler.
-                changeCount++;
+                _ = changeCount.IncrementAndGet();
                 segmentInfos.Changed();
                 return "_" + SegmentInfos.SegmentNumberToString(segmentInfos.Counter++, allowLegacyNames: false); // LUCENENET specific - we had this right thru all of the betas, so don't change if the legacy feature is enabled
             }
@@ -2821,7 +2822,7 @@ namespace Lucene.Net.Index
                     deleter.Checkpoint(segmentInfos, false);
                     deleter.Refresh();
 
-                    lastCommitChangeCount = changeCount;
+                    lastCommitChangeCount.Value = changeCount.Value;
 
                     deleter.Refresh();
                     deleter.Dispose();
@@ -2957,7 +2958,7 @@ namespace Lucene.Net.Index
                             // Don't bother saving any changes in our segmentInfos
                             readerPool.DropAll(false);
                             // Mark that the index has changed
-                            ++changeCount;
+                            _ = changeCount.IncrementAndGet();
                             segmentInfos.Changed();
                             globalFieldNumberMap.Clear();
                             success = true;
@@ -3128,7 +3129,7 @@ namespace Lucene.Net.Index
             UninterruptableMonitor.Enter(this);
             try
             {
-                changeCount++;
+                _ = changeCount.IncrementAndGet();
                 deleter.Checkpoint(segmentInfos, false);
             }
             finally
@@ -3144,7 +3145,7 @@ namespace Lucene.Net.Index
             UninterruptableMonitor.Enter(this);
             try
             {
-                changeCount++;
+                _ = changeCount.IncrementAndGet();
                 segmentInfos.Changed();
             }
             finally
@@ -3935,7 +3936,7 @@ namespace Lucene.Net.Index
                                 // sneak into the commit point:
                                 toCommit = (SegmentInfos)segmentInfos.Clone();
 
-                                pendingCommitChangeCount = changeCount;
+                                pendingCommitChangeCount.Value = changeCount.Value;
 
                                 // this protects the segmentInfos we are now going
                                 // to commit.  this is important in case, eg, while
@@ -4027,7 +4028,7 @@ namespace Lucene.Net.Index
             try
             {
                 segmentInfos.UserData = new Dictionary<string, string>(commitUserData);
-                ++changeCount;
+                _ = changeCount.IncrementAndGet();
             }
             finally
             {
@@ -4170,7 +4171,7 @@ namespace Lucene.Net.Index
                             infoStream.Message("IW", "commit: wrote segments file \"" + pendingCommit.GetSegmentsFileName() + "\"");
                         }
                         segmentInfos.UpdateGeneration(pendingCommit);
-                        lastCommitChangeCount = pendingCommitChangeCount;
+                        lastCommitChangeCount.Value = pendingCommitChangeCount.Value;
                         rollbackSegments = pendingCommit.CreateBackupSegmentInfos();
                         // NOTE: don't use this.checkpoint() here, because
                         // we do not want to increment changeCount:
@@ -5144,8 +5145,8 @@ namespace Lucene.Net.Index
                         int delCount = NumDeletedDocs(info);
                         if (Debugging.AssertsEnabled) Debugging.Assert(delCount <= info.Info.DocCount);
                         double delRatio = ((double)delCount) / info.Info.DocCount;
-                        merge.EstimatedMergeBytes += (long)(info.GetSizeInBytes() * (1.0 - delRatio));
-                        merge.totalMergeBytes += info.GetSizeInBytes();
+                        _ = merge.estimatedMergeBytes.AddAndGet((long)(info.GetSizeInBytes() * (1.0 - delRatio)));
+                        _ = merge.totalMergeBytes.AddAndGet(info.GetSizeInBytes());
                     }
                 }
 
--- a/src/Lucene.Net/Support/Threading/AtomicDouble.cs
+++ b/src/Lucene.Net/Support/Threading/AtomicDouble.cs
@@ -39,6 +39,10 @@ namespace Lucene.Net.Support.Threading
     /// It does not have the increment, decrement, and add methods because those operations are not atomic
     /// due to the conversion to/from <see cref="long"/>.
     /// </summary>
+    /// <remarks>
+    /// Note that this class is set up to mimic <c>double</c> in Java, rather than the J2N <see cref="J2N.Numerics.Double"/> class.
+    /// This may cause differences in comparing NaN values.
+    /// </remarks>
 #if FEATURE_SERIALIZABLE
     [Serializable]
 #endif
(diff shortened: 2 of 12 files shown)


src/Lucene.Net/Index/IndexWriter.cs
```diff
@@ -5144,8 +5145,8 @@ internal bool RegisterMerge(MergePolicy.OneMerge merge)
                         int delCount = NumDeletedDocs(info);
                         if (Debugging.AssertsEnabled) Debugging.Assert(delCount <= info.Info.DocCount);
                         double delRatio = ((double)delCount) / info.Info.DocCount;
-                        merge.EstimatedMergeBytes += (long)(info.GetSizeInBytes() * (1.0 - delRatio));
-                        merge.totalMergeBytes += info.GetSizeInBytes();
+                        _ = merge.estimatedMergeBytes.AddAndGet((long)(info.GetSizeInBytes() * (1.0 - delRatio)));
```
NightOwl888, line 5148: Are you getting warnings for these? It seems odd to have to use the discard character for ignoring the result of a method, which is a fairly common thing to do.

src/Lucene.Net/Index/IndexWriter.cs
```diff
@@ -4027,7 +4028,7 @@ public void SetCommitData(IDictionary<string, string> commitUserData)
             try
             {
                 segmentInfos.UserData = new Dictionary<string, string>(commitUserData);
-                ++changeCount;
+                _ = changeCount.IncrementAndGet();
```
NightOwl888, line 4031: Are you getting warnings for these? It seems odd to have to use the discard character for ignoring the result of a method, which is a fairly common thing to do.

src/Lucene.Net/Index/IndexWriter.cs
```diff
@@ -2197,7 +2198,7 @@ internal string NewSegmentName()
                 // could close, re-open and re-return the same segment
                 // name that was previously returned which can cause
                 // problems at least with ConcurrentMergeScheduler.
-                changeCount++;
+                _ = changeCount.IncrementAndGet();
```
NightOwl888, line 2201: Are you getting warnings for these? It seems odd to have to use the discard character for ignoring the result of a method, which is a fairly common thing to do.

src/Lucene.Net/Index/IndexWriter.cs
```diff
@@ -2957,7 +2958,7 @@ public virtual void DeleteAll()
                             // Don't bother saving any changes in our segmentInfos
                             readerPool.DropAll(false);
                             // Mark that the index has changed
-                            ++changeCount;
+                            _ = changeCount.IncrementAndGet();
```
NightOwl888, line 2961: Are you getting warnings for these? It seems odd to have to use the discard character for ignoring the result of a method, which is a fairly common thing to do.

src/Lucene.Net/Index/IndexWriter.cs
```diff
@@ -3128,7 +3129,7 @@ internal virtual void CheckpointNoSIS()
             UninterruptableMonitor.Enter(this);
             try
             {
-                changeCount++;
+                _ = changeCount.IncrementAndGet();
```
NightOwl888, line 3132: Are you getting warnings for these? It seems odd to have to use the discard character for ignoring the result of a method, which is a fairly common thing to do.

src/Lucene.Net/Index/IndexWriter.cs
```diff
@@ -3144,7 +3145,7 @@ internal void Changed()
...
```
NightOwl888, line 3148: Are you getting warnings for these? It seems odd to have to use the discard character for ignoring the result of a method, which is a fairly common thing to do.

NightOwl888 (changes requested):

src/Lucene.Net/Support/Threading/AtomicDouble.cs
```diff
@@ -0,0 +1,413 @@
...
+
+        /// <summary>
+        /// Compares <paramref name="a1"/> and <paramref name="a2"/> for value equality.
+        /// </summary>
+        /// <param name="a1">The first number.</param>
+        /// <param name="a2">The second number.</param>
+        /// <returns><c>true</c> if the given numbers are equal; otherwise, <c>false</c>.</returns>
+        public static bool operator ==(double a1, AtomicDouble a2)
+        {
+            return a1.Equals(a2.Value);
```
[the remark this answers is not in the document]
paulirwin: There is a separate override with AtomicDouble on the left below.

src/Lucene.Net/Support/Threading/AtomicDouble.cs
```diff
@@ -0,0 +1,413 @@
...
+
+        /// <summary>
+        /// Compares <paramref name="a1"/> and <paramref name="a2"/> for value equality.
+        /// </summary>
+        /// <param name="a1">The first number.</param>
+        /// <param name="a2">The second number.</param>
+        /// <returns><c>true</c> if the given numbers are equal; otherwise, <c>false</c>.</returns>
+        public static bool operator ==(double? a1, AtomicDouble a2)
+        {
+            return a1.GetValueOrDefault().Equals(a2.Value);
```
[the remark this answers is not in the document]
paulirwin: There is a separate override with AtomicDouble on the left above.

src/Lucene.Net/Support/Threading/AtomicDouble.cs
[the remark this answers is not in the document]
NightOwl888: What I mean is let's use the comparison rules from this class in case we move it to J2N and use the Java `Double` equality comparison rules in the future. If we use `AtomicDouble` on the left side of the comparison, we won't accidentally be tied to the .NET double equality comparison if we didn't intend that.

```c#
if (a1 is null)
    return a2 is null;
if (a2 is null)
    return false;

return a2.Equals(a1);
```

src/Lucene.Net/Index/IndexWriter.cs
paulirwin: Regardless of code analysis, given the name of the method `AddAndGet`, I feel it is most prudent and responsible to communicate to someone reading the code in the future that you're intentionally discarding the result. In this case, I don't think there was a CA warning about it (although there have been times I've added that to address warnings elsewhere), it's just a practice I use when methods clearly indicate they're trying to return a potentially useful value, and we want to not use that value.

src/Lucene.Net/Support/Threading/AtomicDouble.cs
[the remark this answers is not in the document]
paulirwin: In this particular one a1 can't be null, but I'll adopt that pattern for the other

src/Lucene.Net/Support/Threading/AtomicDouble.cs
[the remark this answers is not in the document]
paulirwin: Resolved per other comment

src/Lucene.Net/Support/Threading/AtomicDouble.cs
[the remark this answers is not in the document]
paulirwin: Updated, let me know if that's what you mean.

NightOwl888 (changes requested):

NightOwl888 (approved):
I found a couple more minor issues with parameter/variable naming and the fact that `TryFormat` wasn't overridden and pushed a couple of commits. I have opened https://github.com/NightOwl888/J2N/issues/125 and https://github.com/NightOwl888/J2N/issues/126 to address some of the issues that were discovered here in J2N.

PR #1067, opened 2024-12-17
Port Packed Producing Python Programs, #1008

### LINKED ISSUE #1008: Generate C# code for Lucene.Net.Util.Packed types

### Task description

Found when reviewing #1007. We should support generating the C# code files in Lucene.Net.Util.Packed like Java Lucene does with their Python scripts. See gen_BulkOperation.py, gen_Direct.py, gen_Packed64SingleBlock.py, and gen_PackedThreeBlocks.py in the Lucene codebase.

paulirwin:

Ports the Python scripts for generating the various Lucene.Net.Util.Packed files.

Fixes #1008

## Description

This ports the four Python scripts in the Packed package/namespace that generate the various files there, and updates them to produce C# code instead, as well as updates them to be compatible with Python 3 (as Python 2, which was used at the time, is now very out of support). This results in some minor syntax differences in the generated code. The README.md file in this folder explains how to run the scripts (which is simply to run them with Python 3, no virtual environment or dependencies needed).

Having these scripts ported will help us ensure that the code is being generated the same way as in Java, as well as to assist any necessary future changes to how the files are generated. It also just helps complete our 1:1 port of the Lucene code, be it Java or Python. 

This also corrects an out-of-date type name in the package.md documentation for this namespace.


src/Lucene.Net/Util/Packed/gen_PackedThreeBlocks.py
```diff
@@ -0,0 +1,192 @@
...
+# ## LUCENENET PORTING NOTES
+# This script was originally written for Python 2, but has been tested against Python 3.
+# No changes were necessary to run this script in Python 3.
+
+HEADER="""using Lucene.Net.Diagnostics;
+using Lucene.Net.Support;
+using System;
+using System.Runtime.CompilerServices;
+
+// This file has been automatically generated, DO NOT EDIT
```
NightOwl888, line 27: Maybe we should make this message more prominent and indicate how to generate the file so one doesn't have to search if they see the message.

```c#
//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by the gen_PackedThreeBlocks.py script.
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------
```

src/Lucene.Net/Util/Packed/gen_Packed64SingleBlock.py
```diff
@@ -0,0 +1,338 @@
...
+# No changes were necessary to run this script in Python 3.
+
+SUPPORTED_BITS_PER_VALUE = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12, 16, 21, 32]
+
+HEADER="""using Lucene.Net.Diagnostics;
+using Lucene.Net.Support;
+using System;
+using System.Runtime.CompilerServices;
+
+// This file has been automatically generated, DO NOT EDIT
```
NightOwl888, line 29: Maybe we should make this message more prominent and indicate how to generate the file so one doesn't have to search if they see the message.

```c#
//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by the gen_Packed64SingleBlock.py script.
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------
```

src/Lucene.Net/Util/Packed/gen_Direct.py
```diff
@@ -0,0 +1,224 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ## LUCENENET PORTING NOTES
+# This script was originally written for Python 2, but has been tested against Python 3.
+# No changes were necessary to run this script in Python 3.
+
+# LUCENENET specific - we write the usings before the header
+HEADER="""// This file has been automatically generated, DO NOT EDIT
```
NightOwl888, line 23: Maybe we should make this message more prominent and indicate how to generate the file so one doesn't have to search if they see the message.

```c#
//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by the gen_Direct.py script.
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------
```

src/Lucene.Net/Util/Packed/gen_BulkOperation.py
```diff
@@ -0,0 +1,414 @@
...
+
+"""Code generation for bulk operations"""
+
+MAX_SPECIALIZED_BITS_PER_VALUE = 24
+PACKED_64_SINGLE_BLOCK_BPV = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 12, 16, 21, 32]
+OUTPUT_FILE = "BulkOperation.cs"
+
+# LUCENENET specific - we write the usings before the header, since
+# the numbered files don't need them
+HEADER = """// This file has been automatically generated, DO NOT EDIT
```
NightOwl888, line 40: Maybe we should make this message more prominent and indicate how to generate the file so one doesn't have to search if they see the message.

```c#
//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by the gen_BulkOperation.py script.
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------
```

src/Lucene.Net/Util/Packed/README.md
```diff
@@ -0,0 +1,28 @@
...
+## Generating the files
+
+There are four Python scripts in this folder:
+[`gen_BulkOperation.py`](gen_BulkOperation.py),
+[`gen_Direct.py`](gen_Direct.py),
+[`gen_Packed64SingleBlock.py`](gen_Packed64SingleBlock.py), and
+[`gen_PackedThreeBlocks.py`](gen_PackedThreeBlocks.py).
+These scripts were ported from the original Lucene 4.8.1 code, and modified to generate the corresponding C# code instead of Java.
+
+To generate the files, run the following commands from this directory:
```
NightOwl888, line 12: While this documentation is useful to have, what is preventing us from adding these scripts to the build as was the case in Lucene?

https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/core/build.xml#L101-L127

Granted, we would need to add a Python3 dependency to Visual Studio, command line, and CI builds, but that doesn't seem like such a big deal. We would need a separate PR to update the main README so it can be applied after this is released.

NightOwl888 (changes requested):

src/Lucene.Net/Util/Packed/README.md
paulirwin: Are you thinking to have the csproj execute the scripts before build? We could certainly do that, although given the low frequency with which these scripts change, and that the generated files are already in the build, I don't know if it's worth the extra effort for users to have to make sure they have Python set up first in order to contribute. This would likely impact Windows users the most, as it's more common for macOS and Linux users (particularly developers) to already have python3 in their PATH. And given that admittedly most .NET developers are using Windows, I worry that it's just a possible speed bump in getting new contributors on board.

What we could do is only do this in the build scripts and CI, so that anyone that just opens up the solution in VS/Rider doesn't have to have Python installed to build, but we still get the benefit of being able to test it locally in an automated way as well as have it run as part of the CI builds for regression testing of the scripts.

However we do it, note that there is already a PR for the beta18 documentation changes including README changes that could be updated: https://github.com/apache/lucenenet/pull/1043

NightOwl888: Good point about installing Python3 on Windows. Although, this is exactly what you need to go through to setup dotnet/runtime to build (along with several other dependencies), it is just a matter of having decent documentation for the developer setup process.

There are also other code generation scripts to port, some of which use Python.

That being said, I think we could probably skip all of that for the time being and change the CI builds (especially GitHub Actions) so they generate the files and then use Git to determine if there are any changes to them. If a generated file doesn't match the repository, fail the build. This would catch both cases:

1. The script changed and the files weren't regenerated and committed
2. One or more of the generated files was manually changed

If we do that, the normal build for developers wouldn't need to change and we wouldn't have to manually check to ensure things didn't get out of sync.

paulirwin: Something else I just thought of, the upstream code has that `fixcrlf` task in the build.xml file, I need to determine if that's something we would need to do on Windows as well.

paulirwin: FYI, this seems to work just fine on Windows. It checks out the files as crlf, and running the scripts again does not result in any pending changes, and they remain crlf after running the script.

paulirwin: Added GitHub Actions workflow to run the python scripts and check the git diff for any changes.

NightOwl888 (approved):
That looks pretty good.

I have a couple of minor nit-picky things, but I will leave it up to you whether we should actually change them.

1. The name of the workflow `Lucene-Net-Packed-Python-Scripts.yml` doesn't reflect the namespace where packed is (`Util`).
2. In the IDE, it is more obvious that a file is generated by giving it a special name than having to rely on scrolling to the top of the file to see a warning that it is generated. In ICU4N, we are using the convention `.generated.cs` as a suffix for generated code. The tradeoff here is that we are also using a convention of keeping the file names 1-to-1 with the Java file names so it is easy to compare them lexographicallly in File Explorer so we can quickly tell when we have differences. Using the former approach may affect the latter to some extent, but I think it is something we can deal with (especially being that they are generated).

paulirwin:
I'm renaming the GitHub workflow file. I think we should keep the current names because it more closely matches the Java file names. This can always be revisited later in a separate PR if desired.

PR #1071, opened 2024-12-21
BREAKING: Replace SetOnce Get/Set with Value, #1069

### LINKED ISSUE #1069: Task: Refactor SetOnce<T> to .NETify

### Task description

`SetOnce<T>` has `Get()` and `Set()` methods without any overloads. This class can be improved by changing `Get()` and `Set()` methods to a `Value` property to match `Nullable<T>` in .NET.

`Get()` and `Set()` should be deprecated and moved to the `Support/ObsoleteAPI` (for removal in the rc1). This can be done by making `SetOnce<T>` into a partial class, which allows us to simply delete the whole class to remove the obsolete members.

> NOTE: In Java, there was no restriction on which types could be used in `SetOnce<T>`, but all custom types in Java are reference types. Adapting this class to support value types doesn't seem worth the trouble. Users that need to have this functionality with numeric types can either use reference types from `J2N.Numerics` or make their own `SetOnceInt32` type.

paulirwin:

Replaces Get and Set on the `SetOnce<T>` type with a new `Value` property, marks old methods as deprecated for removal.

Fixes #1069

## Description

In addition to the above, this also enables nullable in the file, and adds more meaningful exceptions in cases where otherwise it could possibly throw a NRE.

### DIFF
--- a/src/Lucene.Net.TestFramework/Index/MockRandomMergePolicy.cs
+++ b/src/Lucene.Net.TestFramework/Index/MockRandomMergePolicy.cs
@@ -47,7 +47,8 @@ namespace Lucene.Net.Index
             int numSegments/* = segmentInfos.Count*/; // LUCENENET: IDE0059: Remove unnecessary value assignment
 
             JCG.List<SegmentCommitInfo> segments = new JCG.List<SegmentCommitInfo>();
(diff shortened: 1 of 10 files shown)


NightOwl888 (approved):
Thanks for the PR. Looks good to merge.

PR #1075, opened 2024-12-24
BREAKING: Use BOM-less UTF-8 encoding for writes, #1027

### LINKED ISSUE #1027: Review all Encoding usage for BOM compatibility

### Task description

Java's `StandardCharsets.UTF_8` does not write a Byte-Order Mark (BOM), while .NET's `System.Text.Encoding.UTF8` does include a BOM by default. We have ensured that the `IOUtils.CHARSET_UTF_8` does not include a BOM to match Java, and as part of #1018 we've added an internal Support class to allow for using `StandardCharsets.UTF_8`, but we need to review all usage of `System.Text.Encoding.UTF8` to determine if it should be replaced with `StandardCharsets.UTF_8` or `IOUtils.CHARSET_UTF_8` (whatever best matches the corresponding Java Lucene code) to avoid BOM issues.

paulirwin:

Use BOM-less UTF-8 encoding for writes

Fixes #1027

## Description

See #1027 for details.

### DIFF
--- a/src/Lucene.Net.Benchmark/ByTask/Tasks/WriteEnwikiLineDocTask.cs
+++ b/src/Lucene.Net.Benchmark/ByTask/Tasks/WriteEnwikiLineDocTask.cs
@@ -2,6 +2,7 @@
 using Lucene.Net.Benchmarks.ByTask.Utils;
 using Lucene.Net.Documents;
 using Lucene.Net.Index;
+using Lucene.Net.Support;
 using System;
 using System.IO;
 using System.Text;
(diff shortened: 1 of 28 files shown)


paulirwin:
@NightOwl888 I assume this comment on the issue was really for this PR instead: https://github.com/apache/lucenenet/issues/1027#issuecomment-2562788290

I went ahead and took option 1 to remove the field, and renamed `CHARSET_UTF_8` as you noted as well. For consistency with use elsewhere and in Lucene, I left an underscore between "UTF" and "8" as `ENCODING_UTF_8_NO_BOM`.

NightOwl888 (approved):

PR #1080, opened 2024-12-31
BREAKING: Add cancellation support to IndexSearcher, #922

### LINKED ISSUE #922: Add cancellation support to IndexSearcher

_No response_

From @NightOwl888: 

> What I was wondering is whether it makes sense to be able to cancel an in progress search (and whether it would function as expected). We are using a .NET TaskScheduler here and CancellationToken is part of the tool set. This isn't something that exists in Lucene 4.8.0, but I think it would be useful to have in .NET if it can be implemented reliably. That is, if there isn't some other way to cancel a search that I am unaware of.

Context: https://github.com/apache/lucenenet/pull/914#discussion_r1501763220

### Additional context

_No response_

paulirwin:

Adds basic cancellation support to IndexSearcher.

Partial #922

## Description

This adds optional CancellationToken parameters to all `IndexSearcher.Search` and `SearchAfter` methods, which is a breaking change to the API.

If an executor is provided in the constructor for multithreaded search, then the provided CancellationToken is passed to the `Task.Wait(CancellationToken)` method. If an executor is not provided, then it will call `ThrowIfCancellationRequested()` in the synchronous case at the evaluation of each leaf reader context.

This will not help cancellation for i.e. single-leaf reader contexts that might take a while to complete (that would require a much larger and messier change to add CancellationToken support to Scorer and ICollector, and benchmarking to ensure we don't hurt performance by doing so), but should at least help with many scenarios that require cancellation.


NightOwl888:
Repeating this comment from https://github.com/apache/lucenenet/pull/1119#discussion_r1925579476:

After taking a step back from this and considering that we are working on support for `CancellationToken`s, I wonder if we should consider removing `LimitedConcurrencyLevelTaskScheduler.Shutdown()` and always use `CancellationToken` instead?

If we do keep the `LimitedConcurrencyLevelTaskScheduler.Shutdown()` method, there are some things to consider:

1. Should we make `LimitedConcurrencyLevelTaskScheduler` public? The BCL doesn't have one and it is provided in the docs for `TaskScheduler`. But there is no `ShutDown()` on that implementation.
2. There are some tests where `Shutdown()` was called in Lucene that would likely perform better if we use it, since those lines were often commented out in Lucene.NET. However, using `CancellationToken` could be an alternative way to handle those cases.
    -  https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/core/src/test/org/apache/lucene/search/TestBooleanQuery.java#L177-L178
    - https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/core/src/test/org/apache/lucene/search/TestSearcherManager.java#L291-L292
    - https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/suggest/src/test/org/apache/lucene/search/spell/TestSpellChecker.java#L435-L437
    - https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/test-framework/src/java/org/apache/lucene/index/ThreadedIndexingAndSearchingTestCase.java#L645-L646
    - https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java#L119
    - https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/test-framework/src/java/org/apache/lucene/util/LuceneTestCase.java#L1570

The upside of using `Shutdown()` is that it aligns more closely to Lucene.

The upside of using `CancellationToken` is that it is more fine-grained and can potentially shut down a process quicker, since it will be able to intervene inside of the running task instead of prior to queuing the task.

Allowing a user-defined `TaskScheduler` still seems like a good idea, but since Microsoft didn't allow them to be cancelled, it seems like we should follow the `CancellationToken` approach instead of trying to shoehorn a way to cancel a `TaskScheduler`. Using the same `CancellationToken` to make `LimitedConcurrencyLevelTaskScheduler` stop queuing new work is also something that might be worth considering.

paulirwin:
The LimitedConcurrencyLevelTaskScheduler proposed changes above are outside the scope of this PR, since these changes do not depend on LimitedConcurrencyLevelTaskScheduler (which is only used in tests currently). I agree that it could possibly be a good change to at least improve the tests, so I split that out as #1253.

paulirwin:
There are several callers of `IndexSearcher.Search` that will need cancellation support as well to fully realize this, including things like FacetsCollector, DrillSideways, JoinUtil, etc., even things you might not expect like `SimpleNaiveBayesClassifier.Train` (which seems like it could nicely benefit from cancellation, if you wanted to cancel the training because it was taking too long). I'll add those in a stacked PR after this one is merged. So I have updated this description to remove the word "Fixes" from the issue number, to prevent it from auto-closing once merged.

Meanwhile, I have a few more things to add to polish this up, so it's not quite ready to merge, but this is a good opportunity for the community to chime in if they have any concerns with adding cancellation support to these APIs. This PR has been open for well over a year so I assume there are no concerns, but now is a good time to chime in if so.

Issue #1017, opened 2024-11-11
Task: Review Lucene.Net.TestFramework for Missing Features, Incomplete Features and Inconsistencies


NightOwl888:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Task description

While it is presumed that we have enough of the test framework ported for the 4.8.0 release, we don't currently have an inventory of the features that are not ported or only partially working. For example, there are properties like `LuceneTestCase.SuiteFailureMarker` that are public but currently do not function that should either be made internal or functional before the release. There are also "after suite" cleanup features that are not ported (#898) that seem like they may be useful or even critical for fixing issues like #271.

For informational purposes, here are the open issues for the test framework at the time of this writing:

### Known Test Framework Incomplete Features

- #264
- #306
- #898

### Known Test Framework Bugs/Inconsistencies

- #739
- #923
- #999
- #1001
- #1016

-------------------

This task is to analyze the test framework to determine which features exist that we don't have. It will take some analysis to work out where a feature begins and ends, but once that is determined, we should use it to generate a list of missing features. From that list, we should then open an issue for each feature (assuming there isn't already an issue for it) so we can prioritize it and assign it a milestone.

We should also be on the lookout for features that are incomplete or inconsistent and open issues for those, as well.

If there are any notes that come from this analysis that could be helpful for us to decide on how difficult it is to port a feature or how valuable it is, please add them to the corresponding issue.

> **NOTE:** The main places to look are in `LuceneTestCase` and the missing classes in the `Lucene.Net.Util` namespace (i.e. "TestRule" classes). 
>
> **NOTE:** Some of the [randomized-testing runner](https://github.com/randomizedtesting/randomizedtesting) classes were ported to `Lucene.Net.TestFramework` (at least at one point), so if there are types that are from that library, they can be excluded from the features list because they are covered by #264. We have a substitute for the randomized-testing runner in the Support/Util folder and don't intend to port the entire library for the 4.8.0 release. It is a considerable amount of extra work to make those features general enough to work outside of the context of the test framework. Although, we may be missing tests from randomized-testing that would help to ensure our substitute is working as expected that could be considered "features".

NightOwl888:
Below is the list of features we are missing in the test framework, which is up for discussion about whether or not we need these  features. Some I thought were clear, but feel free to make a case if you disagree. There are a few additional features listed that I think would be useful, but are not necessarily required. There are some others that are probably the source of some of the bugs we are seeing.

[...]

PR #1084, opened 2025-01-02
Finish [SuppressTempFileChecks] attribute functionality, #898

### LINKED ISSUE #898: Task: Finish [SuppressTempFileChecks] attribute functionality

### Task description

This attribute exists in `LuceneTestCase`, but does not currently function. The task involves researching what the original attribute did in Lucene and finding a way to make it function.

Note that most of the other attributes we have ported are called automatically by NUnit because we implement [`IApplyToTest`](https://docs.nunit.org/articles/nunit/extending-nunit/IApplyToTest-Interface.html), [`IApplyToContext`](https://docs.nunit.org/articles/nunit/extending-nunit/IApplyToContext-Interface.html) and [`IWrapTestMethod`](https://docs.nunit.org/api/NUnit.Framework.Interfaces.IWrapTestMethod.html?q=IWrapTestMethod), as appropriate.

We also keep these tests nested within `LuceneTestCase` because these attributes don't function outside the context of `LuceneTestCase` anyway.

paulirwin:

Finishes implementation of the `[SuppressTempFileChecks]` attribute functionality.

Fixes #898

## Description

This attribute will cause the tear down logic to swallow any exceptions when cleaning up temporary files after the tests run. No tests actually need this attribute, but theoretically it is possible that a TestFramework user might want this. 

This also fixes an issue where the previous code was treating _all_ tests as if it had this attribute, and was not throwing if removing the temporary files failed. This change will make our tests more reliable by ensuring they fail if temp file handles are held during cleanup.

I'm also open to removing this attribute, since none of our tests need it and latest Lucene doesn't use it internally either. It could be viewed as a bad practice to just ignore these files being held too long.

### DIFF
--- a/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
+++ b/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
@@ -440,7 +440,6 @@ namespace Lucene.Net.Util
             public string[] Value { get; private set; }
         }
 
-        // LUCENENET TODO: Finish implementation
         /// <summary>
         /// Marks any suites which are known not to close all the temporary
         /// files. This may prevent temp files and folders from being cleaned
@@ -3086,10 +3085,11 @@ namespace Lucene.Net.Util
         [MethodImpl(MethodImplOptions.NoInlining)]
         protected string GetFullMethodName([CallerMemberName] string memberName = "")
         {
-            return string.Format("{0}+{1}", this.GetType().Name, memberName);
+            return $"{this.GetType().Name}+{memberName}";
         }
 
-        private static void CleanupTemporaryFiles()
+        // LUCENENET specific - this is equivalent to TemporaryFilesCleanupRule in Lucene
+        private void CleanupTemporaryFiles()
         {
             // Drain cleanup queue and clear it.
             FileSystemInfo[] everything;
@@ -3126,19 +3126,20 @@ namespace Lucene.Net.Util
                 }
                 catch (Exception e) when (e.IsIOException())
                 {
-                    //                    Type suiteClass = RandomizedContext.Current.GetTargetType;
-                    //                    if (suiteClass.IsAnnotationPresent(typeof(SuppressTempFileChecks)))
-                    //                    {
-                    Console.Error.WriteLine("WARNING: Leftover undeleted temporary files " + e.Message);
-                    return;
-                    //                    }
+                    Type suiteClass = this.GetType();
+                    if (suiteClass.GetCustomAttribute<SuppressTempFileChecksAttribute>(inherit: true) is { } suppressAttr)
+                    {
+                        Console.Error.WriteLine($"WARNING: Leftover undeleted temporary files (bugUrl: {suppressAttr.BugUrl}): {e.Message}");
+                        return;
+                    }
+                    throw;
                 }
             }
             else
             {
                 if (tempDirBasePath != null)
                 {
-                    Console.Error.WriteLine("NOTE: leaving temporary files on disk at: " + tempDirBasePath);
+                    Console.Error.WriteLine($"NOTE: leaving temporary files on disk at: {tempDirBasePath}");
                 }
             }
         }
--- a/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
+++ b/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
@@ -722,11 +722,11 @@ namespace Lucene.Net.Util
         /// </summary>
         internal static TestRuleSetupAndRestoreClassEnv ClassEnvRule { get; } = new TestRuleSetupAndRestoreClassEnv();
 
-        // LUCENENET TODO
+        // LUCENENET TODO: Implement these rules
         /// <summary>
         /// Suite failure marker (any error in the test or suite scope).
         /// </summary>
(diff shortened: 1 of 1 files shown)


NightOwl888 (changes requested):

src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
```diff
@@ -3126,19 +3126,20 @@ private static void CleanupTemporaryFiles()
...
                 catch (Exception e) when (e.IsIOException())
                 {
-                    //                    Type suiteClass = RandomizedContext.Current.GetTargetType;
-                    //                    if (suiteClass.IsAnnotationPresent(typeof(SuppressTempFileChecks)))
-                    //                    {
-                    Console.Error.WriteLine("WARNING: Leftover undeleted temporary files " + e.Message);
-                    return;
-                    //                    }
+                    Type suiteClass = this.GetType();
+                    if (suiteClass.GetCustomAttribute<SuppressTempFileChecksAttribute>(inherit: true) is { } suppressAttr)
```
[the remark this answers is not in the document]
paulirwin: Changed to use `RandomizedContext.CurrentContext.CurrentTest.TypeInfo?.Type.GetCustomAttribute<SuppressTempFileChecksAttribute>` to ensure we're getting that attribute from the class instead of the method. AttributeUsage should enforce that, but is a closer match to the Lucene logic of looking at the type instead of the method.

src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
```diff
@@ -3126,19 +3126,20 @@ private static void CleanupTemporaryFiles()
...
+                    {
+                        Console.Error.WriteLine($"WARNING: Leftover undeleted temporary files (bugUrl: {suppressAttr.BugUrl}): {e.Message}");
```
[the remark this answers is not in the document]
paulirwin: Split out implementing the full RuleChains as #1088. For now, I've changed TearDown to set that static field to false in the event of any test failure, which is all this logic cares about. Added a TODO to indicate that it can be removed when #1088 is implemented, which probably will require #1087 done first.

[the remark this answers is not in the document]
NightOwl888: I don't think the rule chain makes any sense for these particular features because NUnit manages these in static context so we don't have to re-invent the wheel. But, since these are all now listed in #1017 we can do those separately.

src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
```diff
@@ -938,6 +937,10 @@ public virtual void TearDown()
 
             if (result.ResultState == ResultState.Failure || result.ResultState == ResultState.Error)
             {
+                // LUCENENET specific - Track that at least one test in the suite failed.
+                // LUCENENET TODO: finish implementation of RuleChains so this is not needed (#1088)
+                SuiteFailureMarker = false;
```
NightOwl888, line 942: There is nothing resetting this back to `true`, so the first failure will cascade to the other tests. I think we were better off with this hard coded to `true` because this will leave lots of temp files on the disk for passing tests.

NightOwl888 (changes requested):

src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
paulirwin: That is indeed the point. It tracks whether there have been any failures. In Lucene, `suiteFailureMarker` is static and is flagged as failed if any tests fail.

NightOwl888: Oh, thanks for the clarification. I thought this was only per test.

NightOwl888: According to ChatGPT, that isn't how that works. It resets the state at the beginning of every test: https://chatgpt.com/share/677af21b-bcb8-8005-ad95-b070d0c6cdaa.

But whatever the case, I am hesitant to merge this to master when a test failure could leave all of the temporary files on disk for every successful test in the assembly that runs after it. There is a specific setting for that: `tests:leavetmpdir`.

> Side note: In NUnit, a failure isn't always an exception and `Assert.Pass()` throws a `SuccssException`. So, the `TestRuleMarkFailure` logic won't work as-is.

paulirwin: I really debated whether we should even be doing this anyways in the issue comments. Again, no Lucene tests even use this attribute. I'll let you take ownership of this one from here because I'm losing interest in the purity aspect of getting this working when there is no actual benefit to our code. You can feel free to use whatever of this PR might be useful in your other test framework changes and I'm just going to mark this as draft for now. I'd appreciate getting some of my other PRs reviewed ahead of this one, anyways.

NightOwl888 (approved):

src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
NightOwl888: I reverted the change to set `SuiteFailureMarker` to `false` and will do the implementation of `SuiteFailureMarker` in a separate PR.

PR #1086, opened 2025-01-03
Use full test name in NUnit tests, #923

### LINKED ISSUE #923: Fix test name reporting when test is in a base class

### Task description

Context: https://github.com/apache/lucenenet/pull/914#discussion_r1494919429

From @NightOwl888:

> we should have a new issue about fixing test reports so the correct superclass is reported for tests that are declared in a base class.

> NUnit reports the name of the class that the `[Test]` attribute it finds is declared in. We need it to report the name of the top level superclass that inherited the class that the `[Test]` attribute is declared in, since we have tests in test fixture classes that are designed specifically to be inherited so standardized tests can be reused. Since several of these classes are in the test framework, this issue affects end users of the test framework as well as internally.

> As a workaround, we currently override test methods, redefine the `[Test]` attribute on the superclass, and cascade the call to the base class. This gets NUnit to display the proper class name, but it obviously doesn't scale well.

Also see: https://github.com/apache/lucenenet/pull/914#discussion_r1494864427

paulirwin:

Uses the full test name in NUnit tests, so that it's clear where inherited tests come from.

Fixes #923

## Description

This outputs the full test name including namespace and class name in the GitHub, Azure DevOps, and local build scripts for the log output as well as the trx files.


NightOwl888:
A couple of things:

1. The `.runsettings` file is unnecessary because we can supply the settings on the command line (the same way the other runsettings are applied, i.e. `NUnit.DisplayName=FullName`.
2. This PR assumes the problem only applies to us, but it will also apply to all users of `Lucene.Net.TestFramework`. Arguably, the most useful functionality in `Lucene.Net.TestFramework` is the `BaseTokenStreamTestCase` class, which will have this problem for anyone who uses it. It would not be practical to tell all of our users that they must configure `FullName` in their CI environment to get useful debug info. See my comment about how we can report the class name for all users of the test framework: https://github.com/apache/lucenenet/issues/923#issuecomment-2568702773

paulirwin:
I'll add in the failure message to print the class name.

This fails currently on ADO, wrong path to the runsettings file. I'll replace this with command-line args which will solve the problem. I think it's worthwhile to have it in the test log output and trx file too, not just in the failure message.

NightOwl888:
> I'll add in the failure message to print the class name.
> 
> This fails currently on ADO, wrong path to the runsettings file. I'll replace this with command-line args which will solve the problem. I think it's worthwhile to have it in the test log output and trx file too, not just in the failure message.

Fair enough. I would like to see what the UX looks like without the extra info in the log file, though. Could we make this a parameter that defaults to enabled if not supplied in ADO that allows it to be disabled in a specific ADO pipeline? For example, the properties that are [documented at the top of the `azure-pipelines.yml` file](https://github.com/apache/lucenenet/blob/dbc602c7f872189516dc132045f1cc2b69d5ca7e/azure-pipelines.yml#L37-L52) can all be defined in the pipeline. Perhaps name it `DisplayFullName` and put it under the "Testing variables".

paulirwin:
Updated to remove the runsettings file; pass `NUnit.DisplayName=FullName` after the `--` in `dotnet test` in GitHub, ADO, and local build scripts; add ability to disable this in ADO; and also add to the GitHub workflows to trigger the workflow on workflow changes (so we don't have to do fake commits). I do like this approach better. It results in nice full test names in the GitHub and ADO logs/trx files by default, and should not interfere with the IDE or anyone's existing runsettings files.

NightOwl888 (approved):
The changes look good and I can confirm they work as expected.

There is a bit of extra time used to both log and upload the test results, but it is only around 5%. That being said, I will be configuring the environments I use to disable this feature. The logs are only the secondary way to view the test results and it isn't worth the extra time for me. There are still many things we can do to improve performance on ADO, though (like request more parallel agents, upload the test results in the background in parallel, and start testing NuGet packages so we can skip `dotnet publish`).

The effect on GitHub Actions is far less noticeable because it only uploads the files as artifacts without also pushing them to an API. Not to mention, there are far more build agents working in parallel to minimize the time it takes to run.

.build/azure-templates/run-tests-on-os.yml
```diff
@@ -172,6 +172,7 @@ steps:
     $tempDirectory = "$(Agent.TempDirectory)"
     $isNightly = if ($env:ISNIGHTLY -eq 'true') { 'true' } else { 'false' }
     $isWeekly = if ($env:ISWEEKLY -eq 'true') { 'true' } else { 'false' }
+    $displayFullName = if ($env:DISPLAYFULLNAME -eq 'false') { 'false' } else { 'true' }
```
NightOwl888, line 175: We should update the documentation to note this extra dependency. Normally, all dependencies are defined as parameters at the top of the file, which is why we have this comment:

```
# NOTE: Depends on environment variables $(IsNightly) and $(IsWeekly)
```

Please update it to include `$(DisplayFullName)`.

azure-pipelines.yml
```diff
@@ -119,6 +120,8 @@ stages:
         Write-Host "##vso[task.setvariable variable=IsWeekly;]$isWeekly"
         $runPack = if ($env:ISRELEASE -eq 'true' -or $env:ARTIFACTFEEDID -ne '' -or $env:GENERATEPACKAGES -ne 'false') { 'true' } else { 'false' }
         Write-Host "##vso[task.setvariable variable=RunPack;]$runPack"
+        $displayFullName = if ($env:DISPLAYFULLNAME -eq 'false') { 'false' } else { 'true' }
+        Write-Host "##vso[task.setvariable variable=DisplayFullName;]$displayFullName"
```
NightOwl888, line 124: Since this environment variable is not used in the build stage, we don't need to set its default here. This can be removed.

This could only make sense if we switch to using a `.runsettings` file instead of `lucene.testsettings.json` and build that file here for distribution among the test agents. We do it this way because we would have to write the settings to a file anyway to transport the settings across agent boundaries (which makes me miss TeamCity).

NightOwl888 (changes requested):

NightOwl888 (approved):

PR #1089, opened 2025-01-06
Use DecoderFallback.ExceptionFallback to match Java's CodingErrorAction.REPORT, #1076

### LINKED ISSUE #1076: Review use of Encoding fallback handling

### Task description

There are several places where Lucene uses Java's `CodingErrorAction.REPORT` to throw an exception on failing to decode data. The default in .NET's Encoding class is to use replacements rather than throw exceptions. We should review these usages to determine if we should be using `DecoderFallback.ExceptionFallback` in these places.

paulirwin:

Use DecoderFallback.ExceptionFallback to match Java's `CodingErrorAction.REPORT`

Fixes #1076

## Description

This adds an extension method to make it easy to match Java's `CodingErrorAction.REPORT` use in `CharSet` where it will throw an exception upon encountering invalid character sequences. All places where this is used in Lucene have been updated to use this extension method.

This should improve reliability of ensuring valid input in cases where Lucene would throw on invalid input, particularly with `IOUtils.GetDecodingReader` which, per its docs, was intended to throw on invalid input, but was doing a character replacement in Lucene.NET previously.

### DIFF
--- a/src/Lucene.Net/Support/ExceptionHandling/ExceptionExtensions.cs
+++ b/src/Lucene.Net/Support/ExceptionHandling/ExceptionExtensions.cs
@@ -6,6 +6,7 @@ using System.Reflection;
 using System.Resources;
 using System.Runtime.CompilerServices;
 using System.Security;
+using System.Text;
 using System.Threading;
 
 namespace Lucene
@@ -213,8 +214,11 @@ namespace Lucene
             if (e is null || e.IsAlwaysIgnored()) return false;
 
             return e is IOException ||
-                e.IsAlreadyClosedException() || // In Lucene, AlreadyClosedException subclass IOException instead of InvalidOperationException, so we need a special case here
-                e is UnauthorizedAccessException; // In Java, java.nio.file.AccessDeniedException subclasses IOException
+                   e.IsAlreadyClosedException() || // In Lucene, AlreadyClosedException subclass IOException instead of InvalidOperationException, so we need a special case here
+                   e is
+                       UnauthorizedAccessException // In Java, java.nio.file.AccessDeniedException subclasses IOException
+                       or DecoderFallbackException // In Java, CharacterCodingException subclasses IOException
+                       or EncoderFallbackException;
         }
 
         /// <summary>
@@ -368,9 +372,11 @@ namespace Lucene
             // LUCENENET: In production, there is a chance that we will upgrade to ArgumentNullExcpetion or ArgumentOutOfRangeException
             // and it is still important that those are caught. However, we have a copy of this method in the test environment
             // where this is done more strictly to catch ArgumentException without its known subclasses so we can be more explicit in tests.
-            return e is ArgumentException;
-                //!(e is ArgumentNullException) &&     // Corresponds to NullPointerException, so we don't catch it here.
-                //!(e is ArgumentOutOfRangeException); // Corresponds to IndexOutOfBoundsException (and subclasses), so we don't catch it here.
+            return e is ArgumentException
+                and not DecoderFallbackException // In Java, CharacterCodingException subclasses IOException, not ArgumentException
+                and not EncoderFallbackException;
+            //!(e is ArgumentNullException) &&     // Corresponds to NullPointerException, so we don't catch it here.
+            //!(e is ArgumentOutOfRangeException); // Corresponds to IndexOutOfBoundsException (and subclasses), so we don't catch it here.
         }
 
         /// <summary>
--- a/src/Lucene.Net/Support/Text/EncodingExtensions.cs
+++ b/src/Lucene.Net/Support/Text/EncodingExtensions.cs
@@ -35,8 +35,8 @@ namespace Lucene.Net.Support.Text
         /// Thus, to translate Java code that catches any of those exceptions, you can catch
         /// <see cref="DecoderFallbackException"/>.
         /// </summary>
-        /// <param name="encoding">The encoding to set the fallbacks on.</param>
-        /// <returns>A new <see cref="Encoding"/> instance with the fallbacks set to throw an exception.</returns>
+        /// <param name="encoding">The encoding to clone and set the fallback on.</param>
+        /// <returns>A new <see cref="Encoding"/> instance with the fallback set to throw an exception.</returns>
         /// <remarks>
         /// Note that it is necessary to return a new, cloned <see cref="Encoding"/> instance because
         /// the <see cref="Encoding.DecoderFallback"/> property is read-only without cloning.
--- a/src/Lucene.Net/Support/Text/EncodingExtensions.cs
+++ b/src/Lucene.Net/Support/Text/EncodingExtensions.cs
@@ -1,4 +1,6 @@
+using System.Collections.Concurrent;
 using System.Text;
+#nullable enable
 
 namespace Lucene.Net.Support.Text
 {
@@ -24,8 +26,10 @@ namespace Lucene.Net.Support.Text
     /// </summary>
     internal static class EncodingExtensions
     {
+        private static readonly ConcurrentDictionary<Encoding, Encoding> decoderExceptionFallbackCache = new();
+
         /// <summary>
-        /// Returns a new <see cref="Encoding"/> instance with the <see cref="DecoderFallback"/> set to throw
+        /// Returns an <see cref="Encoding"/> instance with the <see cref="DecoderFallback"/> set to throw
         /// an exception when an invalid byte sequence is encountered.
         /// <para />
         /// This is equivalent to Java's <c>CodingErrorAction.REPORT</c> for both <c>onMalformedInput</c> and
@@ -36,16 +40,19 @@ namespace Lucene.Net.Support.Text
         /// <see cref="DecoderFallbackException"/>.
         /// </summary>
         /// <param name="encoding">The encoding to clone and set the fallback on.</param>
-        /// <returns>A new <see cref="Encoding"/> instance with the fallback set to throw an exception.</returns>
+        /// <returns>An <see cref="Encoding"/> instance with the fallback set to throw an exception.</returns>
         /// <remarks>
-        /// Note that it is necessary to return a new, cloned <see cref="Encoding"/> instance because
+        /// Note that it is necessary to clone the <see cref="Encoding"/> instance because
         /// the <see cref="Encoding.DecoderFallback"/> property is read-only without cloning.
         /// </remarks>
         public static Encoding WithDecoderExceptionFallback(this Encoding encoding)
         {
-            Encoding newEncoding = (Encoding)encoding.Clone();
-            newEncoding.DecoderFallback = DecoderFallback.ExceptionFallback;
-            return newEncoding;
+            return decoderExceptionFallbackCache.GetOrAdd(encoding, static e =>
+            {
+                Encoding newEncoding = (Encoding)e.Clone();
+                newEncoding.DecoderFallback = DecoderFallback.ExceptionFallback;
+                return newEncoding;
+            });
         }
     }
 }
--- a/src/Lucene.Net/Support/Text/EncodingExtensions.cs
+++ b/src/Lucene.Net/Support/Text/EncodingExtensions.cs
@@ -1,0 +1,51 @@
+using System.Text;
+
+namespace Lucene.Net.Support.Text
+{
+    /*
+     * Licensed to the Apache Software Foundation (ASF) under one or more
+     * contributor license agreements.  See the NOTICE file distributed with
+     * this work for additional information regarding copyright ownership.
+     * The ASF licenses this file to You under the Apache License, Version 2.0
+     * (the "License"); you may not use this file except in compliance with
+     * the License.  You may obtain a copy of the License at
+     *
+     *     http://www.apache.org/licenses/LICENSE-2.0
+     *
+     * Unless required by applicable law or agreed to in writing, software
+     * distributed under the License is distributed on an "AS IS" BASIS,
+     * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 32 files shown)


NightOwl888 (changes requested):
Thanks for the PR.

I went ahead and pushed an optimization to the `Term.ToString()` method, since `System.Text.Unicode.Utf8` supports fallback without an exception.

I also discovered that the default fallback character for encodings in .NET doesn't match that in Java, so we need to adjust the value set to `IOUtils.ENCODING_UTF_8_NO_BOM` to:

```c#
    private static UTF8Encoding CreateUtf8Encoding()
    {
        // Create a UTF8Encoding instance
        UTF8Encoding encoding = new UTF8Encoding(
            encoderShouldEmitUTF8Identifier: false, // No BOM, similar to Java
            throwOnInvalidBytes: false             // Match Java's silent handling of invalid bytes
        );

        // Set the DecoderFallback to use '\uFFFD' as the replacement character
        encoding.DecoderFallback = new DecoderReplacementFallback("\uFFFD");

        // Set the EncoderFallback to use '\uFFFD' for invalid character encoding
        encoding.EncoderFallback = new EncoderReplacementFallback("\uFFFD");

        return encoding;
    }
```

For the `Encoding`-derived classes, the default fallback character is `?`, but in Java it is `\uFFFD`. The new `System.Text.Unicode.Utf8` class uses the same default fallback character as in Java.

Note that the fallback character has been defined in other places in the codebase as `?`, as well. Those should be reviewed.

Since this PR is encoding-related, I think it would be fine to put them here, but feel free to open separate issues if you want to keep them separate.

paulirwin:
@NightOwl888 In regards to:

> For the Encoding-derived classes, the default fallback character is ?, but in Java it is \uFFFD. The new System.Text.Unicode.Utf8 class uses the same default fallback character as in Java.

This is only true for `Encoding.ASCII` (at least amongst those defined on `Encoding`). `Encoding.UTF8` by default returns `\uFFFD` (from csharprepl):

```
> Encoding.ASCII.GetString(new byte[] { 0xc3 })
"?"
> Encoding.UTF8.GetString(new byte[] { 0xc3 })
"�"
> Encoding.UTF8.GetString(new byte[] { 0xc3 })[0] == '\ufffd'
true
```
I confirmed this is the case on net462 as well.

The only place in non-test code where `Encoding.ASCII` is used is in ConnectionCostsBuilder, and that doesn't matter because this PR changes it to throw on invalid characters anyways. 

Given that the rest of the places we use UTF8, I don't think we need to make that change.

src/Lucene.Net/Support/Text/EncodingExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// </summary>
+        /// <param name="encoding">The encoding to clone and set the fallback on.</param>
+        /// <returns>A new <see cref="Encoding"/> instance with the fallback set to throw an exception.</returns>
+        /// <remarks>
+        /// Note that it is necessary to return a new, cloned <see cref="Encoding"/> instance because
+        /// the <see cref="Encoding.DecoderFallback"/> property is read-only without cloning.
+        /// </remarks>
+        public static Encoding WithDecoderExceptionFallback(this Encoding encoding)
+        {
+            Encoding newEncoding = (Encoding)encoding.Clone();
```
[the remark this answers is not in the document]
paulirwin: I reviewed the GetHashCode .NET source code, and it seems fine for this use to me. In the case of the most common encoding we use, UTF8, it adds the hash codes of the encoder and decoder fallbacks (which are either the hashcode of the replacement string or a hardcoded number in the case of exception fallback), the UTF8 codepage constant 65001, and then either a 1 or 0 if it should use the BOM. If by "optimized" you mean optimized for good hash distribution, that is possibly true: the comment in the code says "Not great distribution, but this is relatively unlikely to be used as the key in a hashtable." Hm... well we are about to do just that.

But honestly, most apps are only going to have one record in this dictionary (UTF8) or at most a few, so hashtable distribution probably does not matter. I ran some benchmarks comparing no caching to concurrent dictionary (as well as just caching the common UTF8 case to a static field), in the case of just UTF8 as well as a few common encodings, and concurrent dictionary did result in a speed-up in the just-UTF8 case and reduced allocations in both, even with perhaps suboptimal hash distribution. It was slightly slower when multiple encodings were stored, but the allocations were still reduced. The happy-path caching case of just caching UTF8 was the fastest, and had the same allocations as the ConcurrentDictionary case. If you're interested in that, we could take that approach. We'd have special cases for Encoding.UTF8 as well as the IOUtils version, but anything else would allocate a new encoding on Clone. I think most apps would only be using the UTF8 case.

```
| Method                             | Mean     | Error    | StdDev   | Gen0   | Allocated |
|----------------------------------- |---------:|---------:|---------:|-------:|----------:|
| NoCaching_JustUTF8                 | 42.77 ns | 0.794 ns | 0.704 ns | 0.0255 |     160 B |
| NoCaching_AFewEncodings            | 77.53 ns | 0.379 ns | 0.355 ns | 0.0331 |     208 B |
| ConcurrentDictionary_JustUTF8      | 33.15 ns | 0.179 ns | 0.150 ns | 0.0102 |      64 B |
| ConcurrentDictionary_AFewEncodings | 94.89 ns | 0.495 ns | 0.413 ns | 0.0235 |     148 B |
| StaticUTF8_JustUTF8                | 23.92 ns | 0.353 ns | 0.313 ns | 0.0102 |      64 B |
| StaticUTF8_AFewEncodings           | 76.18 ns | 1.214 ns | 1.076 ns | 0.0293 |     184 B |
```

NightOwl888 (changes requested):
I did a more exhaustive review and I found some additional issues that we should address.

### Exception Handling

We need to fix the `ExceptionExtensions` class so that any existing catch blocks will account for gaps between encoding fallback exceptions between Java and .NET.

- In Java, [`CharacterCodingException`](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CharacterCodingException.html) subclasses `IOException`.
- In .NET, [`EncoderFallbackException`](https://learn.microsoft.com/en-us/dotnet/api/system.text.encoderfallbackexception?view=net-9.0) and [`DecoderFallbackException`](https://learn.microsoft.com/en-us/dotnet/api/system.text.decoderfallbackexception?view=net-9.0) subclass `ArgumentException`

I already checked where we are catching `ArgumentException` explicitly, and I think we are good there.

- [`IsIOException()`](https://github.com/apache/lucenenet/blob/c3b689680b58f5c29b79143061d2d043a593ecb0/src/Lucene.Net/Support/ExceptionHandling/ExceptionExtensions.cs#L211-L218) should be changed to include both `DecoderFallbackException` and `EncoderFallbackException`.
- [`IsIllegalArgumentException()`](https://github.com/apache/lucenenet/blob/c3b689680b58f5c29b79143061d2d043a593ecb0/src/Lucene.Net/Support/ExceptionHandling/ExceptionExtensions.cs#L363C28-L374) should be changed to exclude both `DecoderFallbackException` and `EncoderFallbackException`.

### Explicit Encoding Char Replacement

As I previously mentioned, there are a few places where we are using `?` as the replacement character instead of `\uFFFD`. I located them and they are all in tests.

- https://github.com/apache/lucenenet/blob/c3b689680b58f5c29b79143061d2d043a593ecb0/src/Lucene.Net.Tests.Facet/Taxonomy/WriterCache/TestCharBlockArray.cs#L43-L88
- https://github.com/apache/lucenenet/blob/c3b689680b58f5c29b79143061d2d043a593ecb0/src/Lucene.Net.Tests.Facet/Taxonomy/WriterCache/TestCompactLabelToOrdinal.cs#L46-L65

While technically, they are fine because they are in tests, during porting what often happens when trying to figure out how to do something is to search the codebase to see how it was done elsewhere. And leaving it like this could cause it to be copied incorrectly elsewhere.

So, let's change those tests to use `StandardCharsets.UTF_8` and add a comment to indicate that configuring `CodingErrorAction.REPLACE` explicitly is unnecessary because it is the default behavior of `StandardCharsets.UTF_8` (in .NET, anyway).

### Default Encoding

During review, I noticed that we have translated `Charset.defaultCharset()` to `Encoding.GetEncoding(0)`. It would be better to change this to `Encoding.Default`. We have comments to indicate `Encoding.GetEncoding(0)` represents the system default in most cases, but using the property would make it clearer.

ChatGPT recommends using `Encoding.Default` instead of `Encoding.GetEncoding(0)` because apparently the latter will throw an exception in some environments on .NET Core. I tried on my system using `net8.0`, but it didn't throw. However, the property definitely won't throw even if ChatGPT is wrong about this.

### Default Encoding with System.Console

Note that there is one place where the default encoding wasn't ported correctly.

- https://github.com/apache/lucenenet/blob/85c01412946ed1e2632cd2dfae4c672efd38caba/src/Lucene.Net.Benchmark/Quality/Trec/QueryDriver.cs#L83

This line can be changed to:

```c#
using TextWriter logger = new StreamWriter(Console.OpenStandardOutput(), Encoding.Default) { AutoFlush = true };
```

### ISO 8859-14 Encoding

This encoding (only the decoder) was ported from Hunspell in Lucene because neither .NET nor Java have an implementation.

#### Caching

Unlike the other encodings it is not cached in .NET. Let's turn `ISO8859_14Encoding` into a singleton by adding a static `Default` property.

#### Fallback Behavior

I ran some tests in Java to find that this neither throws nor replaces invalid input characters. Instead, it simply leaves them as-is. I checked the iso-8859-1 charset (which the Lucene implementation of iso-8859-14 extends) and it does the same thing by default in Java.

But since the user is supplying dictionaries (usually standardized ones that are downloaded from websites that gather them) it makes me wonder whether Lucene should be throwing on invalid input characters. Certainly, leaving them as-is is not a typical option in a .NET Encoding class. Maybe we should report this upstream to find out whether this was intentional or an oversight.

> **NOTE:** Hunspell was among the most complicated parts to port in the Lucene.Net.Analysis.Common package, so it might be worth the extra effort to at least make sure users aren't feeding it invalid input.

> Also, I downloaded all of the Hunspell dictionaries that Lucene uses in tests and put them in a repository here: https://github.com/NightOwl888/lucenenet-hunspell. These are not part of the main repository because it is a huge amount of binary data. The tests currently must be run manually after downloading the files. IIRC, there are still a few dictionaries that don't pass the tests in Lucene.NET, but I don't see that as a showstopper for the release because most of them work.

### GB2312 Encoding

`Encoding.GetEncoding("GB2312")` is being called inside of a tight loop in multiple places in Lucene.Net.Analysis.SmartCn. While the instance returned from `Encoding.GetEncoding("GB2312")` is cached by the encoding provider, it still needs to do a dictionary lookup every time it is called. Let's change it to statically cache this instance in `AbstractDictionary` so the dictionary lookup only occurs once.

NightOwl888:
It appears that there is a difference between the default when using:

- `StandardCharsets.UTF_8`
- `StandardCharsets.UTF_8.newDecoder()`

The default behavior of the latter is `CodingErrorAction.REPORT`, but the same is not true when using the former (which is how Lucene normally passes it to other APIs).

```java
  @Test
  public void testDefaultUTF8DecoderWithInvalidBytes() {
      // Define a mix of valid and invalid UTF-8 bytes
      byte[] testBytes = {
              (byte) 0xE2, (byte) 0x82, (byte) 0xAC, // Valid UTF-8 for €
              (byte) 0xC3, (byte) 0xA9,             // Valid UTF-8 for é
              (byte) 0xFF,                          // Invalid byte
              (byte) 0x41,                          // Valid ASCII for 'A'
              (byte) 0x80,                          // Invalid byte
              (byte) 0xF0, (byte) 0x9F, (byte) 0x92, (byte) 0x96 // Valid UTF-8 for 💖
      };

      // Decode the bytes using the default UTF-8 behavior
      String decodedString = new String(testBytes, StandardCharsets.UTF_8);

      // Print each character and its code point
      System.out.println("Decoded String Code Points: ");
      for (int i = 0; i < decodedString.length(); i++) {
          System.out.printf("Character at position %d: %d%n", i, (int) decodedString.charAt(i));
      }
      
      // Verify the decoded string (replacement character appears at invalid bytes)
      assertEquals("€é\uFFFDA\uFFFD💖", decodedString);
  }
```

paulirwin:
@NightOwl888 Good catch, thanks. I'll look out for any `newDecoder`/`newEncoder` usage that might use the defaults.

paulirwin:
@NightOwl888 Updated with feedback. There were no other cases where `newDecoder`/`newEncoder` were used that needed to be addressed.

src/Lucene.Net/Support/ExceptionHandling/ExceptionExtensions.cs
```diff
@@ -368,9 +372,11 @@ public static bool IsIllegalArgumentException(this Exception e)
             // LUCENENET: In production, there is a chance that we will upgrade to ArgumentNullExcpetion or ArgumentOutOfRangeException
```
NightOwl888, line 372: Apparently, this comment isn't prominent enough. I missed it on my first pass, also. There is a copy of this method in the test framework that also needs to be updated: https://github.com/apache/lucenenet/blob/Lucene.Net_4_8_0_beta00017/src/Lucene.Net.TestFramework/Support/ExceptionHandling/ExceptionExtensions.cs#L54-L62

I suppose we could cascade the call from `IsIllegalArgumentException()` the test framework to this one and then exclude `ArgumentNullException` and `ArgumentOutOfRangeException` from whatever is defined here so these two methods don't go out of sync.

NightOwl888 (changes requested):

src/Lucene.Net/Support/ExceptionHandling/ExceptionExtensions.cs
paulirwin: Updated to cascade the call.

NightOwl888 (approved):
Looks good.

PR #1092, opened 2025-01-08
Lucene.Net.TestFramework: Improved error handling and test reporting

NightOwl888:

Improved error handling and test reporting

## Description

This fixes several concerns regarding output during testing:

1. The details of the random test configuration are now output only if the test fails and will appear in the test message directly. This used to be where they appeared in older versions of NUnit and this restores that functionality.
2. Exceptions that were being called in `LuceneTestCase.OneTimeStartUp()` and `LuceneTestCase.OneTimeTearDown()` were not very visible because NUnit doesn't fail the test when they occur there. This changes it from re-throwing exceptions with an additional message to logging to the output. It goes to stderr in all cases except in `OneTimeTearDown()` where something is redirecting it to stdout. Either way, this changes the ADO pipeline to fail the build so we can avoid shipping the test framework with bugs in these methods.
3. Adds a system property named `tests:failontestfixtureonetimesetuperror` that can be used to cause the test framework to fail all of the tests when there is an error in `OneTimeSetUp()`. We may need to revisit this because there are asserts thrown here that users are supposed to see when dependency injection for codecs is not setup correctly.
4. Removes all messages from the `Assert.Pass()` method because this bloats the size of the test logs with no benefit.

### DIFF
--- a/.build/azure-templates/publish-test-results.yml
+++ b/.build/azure-templates/publish-test-results.yml
@@ -54,7 +54,6 @@ steps:
         $reader = [System.Xml.XmlReader]::Create($testResultsFileName)
         try {
             $countersFound = $false
-            $stdOutFound = $false
             $crashed = $false
             $inRunInfos = $false
             while ($reader.Read()) {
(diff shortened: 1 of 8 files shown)


paulirwin (approved):
This looks good to me. Happy to see the log noise reduced.

PR #1096, opened 2025-01-12
Added DisposeAfterSuite/DisposeAfterTest feature to test framework

### LINKED ISSUE #1094: Add DisposeAfterSuite/DisposeAfterTest functionaltiy to the test framework

### Task description

In Lucene, there were `closeAfterTest()` and `closeAfterSuite()` methods on `LuceneTestCase` that users can use to ensure resources are cleaned up.

The `closeAfterSuite()` method in Lucene registers a wrapper class for each `Directory` that is created, which does a check globally to ensure every `Directory` is closed. We should implement this to help ensure we aren't leaking any `Directory` instances in tests.

This also depends on the `TestRuleMarkFailure` feature, but that is a duplicate of the `RuleState` enum that already exists in NUnit so it is best to leverage that instead.

NightOwl888:

Added DisposeAfterSuite/DisposeAfterTest feature to test framework

Fixes #1094

## Description

This adds the `LuceneTestCase.DisposeAfterTest()` and `LuceneTestCase.DisposeAfterSuite()` methods to the test framework that exist in Lucene 4.8.0.

It also removes the `LuceneTestCase.SuiteFailureMarker` property that wasn't implemented. NUnit already tracks the result state of the tests in `NUnit.Framework.Internal.TestExectionContext.CurrentContext.TestResult.ResultState` and `NUnit.Framework.TestContext.CurrentContext.Result.Outcome` in a context-sensitive and thread safe manner, so there is no need to reinvent the wheel with this. For the same reason, we don't need the `TestFailureMarker` property or `TestRuleMarkFailure` class, which are all part of this feature.

The new feature also revealed that there were some directories that were not being disposed in `Lucene.Net.Tests.Grouping` which were also addressed.

For some reason `TestFsts.TestRealTerms()` does not reliably cleanup its line file docs temp file on Windows, so this adds the `[SuppressTempFileChecks]` attribute to suppress the issue when it happens (for now). At some point we should have a look at this, but it is not a blocker for the release.

### DIFF
--- a/src/Lucene.Net.TestFramework/Support/Util/LifecycleScope.cs
+++ b/src/Lucene.Net.TestFramework/Support/Util/LifecycleScope.cs
@@ -9,7 +9,7 @@ namespace Lucene.Net.Util
     /// <summary>
     /// Lifecycle stages for tracking resources.
     /// </summary>
-    public enum LifecycleScope // From randomizedtesing
+    internal enum LifecycleScope // From randomizedtesing
     {
         /// <summary>
         /// A single test case.
--- a/src/Lucene.Net.TestFramework/Support/Util/LifecycleScope.cs
+++ b/src/Lucene.Net.TestFramework/Support/Util/LifecycleScope.cs
@@ -6,6 +6,23 @@ using System.Threading.Tasks;
 
 namespace Lucene.Net.Util
 {
+    /*
+     * Licensed to the Apache Software Foundation (ASF) under one or more
+     * contributor license agreements.  See the NOTICE file distributed with
+     * this work for additional information regarding copyright ownership.
+     * The ASF licenses this file to You under the Apache License, Version 2.0
+     * (the "License"); you may not use this file except in compliance with
+     * the License.  You may obtain a copy of the License at
+     *
+     *     http://www.apache.org/licenses/LICENSE-2.0
+     *
+     * Unless required by applicable law or agreed to in writing, software
+     * distributed under the License is distributed on an "AS IS" BASIS,
+     * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     * See the License for the specific language governing permissions and
+     * limitations under the License.
+     */
+
     /// <summary>
     /// Lifecycle stages for tracking resources.
     /// </summary>
--- a/src/Lucene.Net.TestFramework/Support/Util/LifecycleScope.cs
+++ b/src/Lucene.Net.TestFramework/Support/Util/LifecycleScope.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Lucene.Net.Util
+{
+    /// <summary>
+    /// Lifecycle stages for tracking resources.
+    /// </summary>
+    public enum LifecycleScope // From randomizedtesing
+    {
+        /// <summary>
+        /// A single test case.
+        /// </summary>
+        TEST,
+
+        /// <summary>
+        /// A single suite (class).
+        /// </summary>
+        SUITE
+    }
+}
--- a/src/Lucene.Net.TestFramework/Support/Util/RandomizedContext.cs
+++ b/src/Lucene.Net.TestFramework/Support/Util/RandomizedContext.cs
@@ -1,7 +1,8 @@
-﻿using NUnit.Framework.Interfaces;
+﻿using Lucene.Net.Support.Threading;
+using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
 using System;
-using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Reflection;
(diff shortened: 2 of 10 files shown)


paulirwin (changes requested):

src/Lucene.Net.TestFramework/Support/Util/LifecycleScope.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Lucene.Net.Util
+{
```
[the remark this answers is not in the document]
NightOwl888: Good catch. The audit tool does a good job at catching these prior to a release, but it is always better to catch them earlier.

src/Lucene.Net.TestFramework/Support/Util/RandomizedContext.cs
```diff
@@ -37,6 +40,12 @@ internal class RandomizedContext
         private readonly long randomSeed;
         private readonly string randomSeedAsHex;
         private readonly long testSeed;
+        private List<IDisposable>? toDisposeAtEnd = null;
```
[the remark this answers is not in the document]
NightOwl888: I ended up using your suggestion of `Lazy<ConcurrentQueue<DisposableResourceInfo>>` after porting the `ClosableResourceInfo` class from randomizedtesting.

`DisposableResourceInfo` keeps track of the details of the caller of `LuceneTestCase.DisposeAfterSuite()` (including the stack trace) so it can be used to log the caller details when an exception is thrown from `Dispose()`. The cheapest way I could find to do this was to instantiate a `StackTrace` instance at the point where we want to record the details and wait until after an exception is thrown to call `ToString()` on it to allocate the string. Let me know if you can think of a more efficient way.

Also note that I didn't port the part of `randomizedtesting` that gives each thread a unique name (yet). So at present, `ThreadName` is always `null`.

src/Lucene.Net.TestFramework/Support/Util/RandomizedContext.cs
```diff
@@ -40,12 +40,7 @@ internal class RandomizedContext
         private readonly long randomSeed;
         private readonly string randomSeedAsHex;
         private readonly long testSeed;
-        private List<IDisposable>? toDisposeAtEnd = null;
-
-        /// <summary>
-        /// Coordination at context level.
-        /// </summary>
-        private readonly object contextLock = new object();
+        private Lazy<ConcurrentQueue<IDisposable>>? toDisposeAtEnd = null;
```
paulirwin: I think this is now "double-lazy" - and the coalesce below removes the thread-safety guarantees of `Lazy<T>`. I think this should be definitely initialized to a `new Lazy<T>(() => ...)`, either with the default thread-safety mode or an explicit one, rather than setting it to null. That way when `.Value` is accessed, it will run the initializer callback if it hasn't been created in a thread-safe way. 

This will also simplify the code below, because the property `ToDisposeAtEnd` can then just be `=> toDisposeAtEnd.Value;`

src/Lucene.Net.TestFramework/Support/Util/RandomizedContext.cs
```diff
@@ -189,19 +184,11 @@ internal void AddDisposableAtEnd(IDisposable resource)
 
         internal void DisposeResources()
         {
-            if (toDisposeAtEnd is not null)
+            Lazy<ConcurrentQueue<IDisposable>>? toDispose = Interlocked.Exchange(ref toDisposeAtEnd, null);
```
paulirwin: This can be simplified and still be thread-safe if the Lazy is non-nullable above. This line could be removed, and replace `toDispose` with `toDisposeAtEnd` below this.

paulirwin (changes requested):
I might have been somewhat misinterpreted about `Lazy<T>` here, so added clarifying comments. I think this code can be simplified further and be made more thread-safe at the same time, as the coalesce is not thread-safe as-is.

src/Lucene.Net.TestFramework/Support/Util/RandomizedContext.cs
NightOwl888: The primary concern is about object lifetime. It is more important to ensure that the list of disposable objects is allowed to go out of scope than it is to ensure it is lazily created. The lifetime of `RandomizedContext` is from the test discovery phase to the end of the test run and we need to ensure these disposable instances go out of scope after the test or class is finished executing. Calling `Clear()` after each test and/or class would be expensive, so it is better to let the list go out of scope without clearing it by setting the field to `null`.

I think that using a lock with a nullable `List<T>` field is simpler. These aren't the only features that use the lock in the upstream code. It would allow the other features of `RandomizedContext` to be added without having to go back and rework it later to use a lock again.

In addition, there won't be any lock contention caused by the test framework because NUnit is managing these events and they do not happen simultaneously. Lock contention could only happen if the user explicitly calls the `DisposeAfterTest()` and/or `DisposeAfterSuite()` methods in parallel. But for users, the `TearDown()` and `OneTimeTearDown()` methods are generally better options than these for calling `Dispose()` on disposable objects.

Thoughts?

NightOwl888: Just to clarify, there is 1 `RandomizedContext` instance per class and 1 `RandomizedContext` instance per test. So, when interacting with `RandomizedContext.CurrentContext`, the instance that is retrieved is different depending on which test or class is currently being executed. What I am trying to achieve is to let the disposable instances that are associated with that specific instance of `RandomizedContext` go out of scope as soon as the test is done with them. Neither the disposable objects or the list instance is used again after that point until the end of the test run.

paulirwin: It didn't seem to me from reading the code like it needed to be set back to null, you would just dispose of all objects in the queue. If you then needed to dispose of more objects, they could be added to the same queue if the instance is shared and then disposed when cleared out again. But if you think it needs to be set to null, I defer to your judgement on that.

PR #1099, opened 2025-01-12
Remove unnecessary NoInlining, #931

### LINKED ISSUE #931: Remove unnecessary`[MethodImpl(MethodImplOptions.NoInlining)]`
Note that we have added `[MethodImpl(MethodImplOptions.NoInlining)]` more liberally than what was actually required. For example, for this case we have added it to every method named `Flush` in the whole solution. There may be a slight performance improvement if we allowed inlining everywhere except for the specific methods that are actually being scanned for in the stack traces of all tests.

This is low priority, but perhaps we should add an issue to remove unnecessary `[MethodImpl(MethodImplOptions.NoInlining)]` attributes. An analysis of which specific methods are being scanned for in the stack trace would be required to work this out, though. Or they could be removed one at a time and rolled back if removing the attribute causes a failure in any one of multiple test runs.

_Originally posted by @NightOwl888 in https://github.com/apache/lucenenet/pull/926#discussion_r1520982896_

paulirwin:

Remove unnecessary use of `[MethodImpl(MethodImplOptions.NoInlining)]`

Partial #931

## Description

I need to finish reviewing the methods mentioned in calls to `StackTraceHelper.DoesStackTraceContainMethod(string methodName)` (the overload without the class name parameter) to make sure there aren't any false positives there, but I've at least validated that those existing usages of NoInlining _seem_ like they might plausibly match their use in the tests. 

This PR removes NoInlining from interface and abstract methods, where it has no effect because the attribute is not inherited. It also removes it from methods with empty bodies, because an empty body doesn't call anything, and thus those checks in the tests would never be true anyways. Finally, it removes it in a few other places that were not asserted in tests.

As a TODO, I left the SystemConsole attributes there, even though those are suspect. We had discussed possibly being able to remove that class anyways, but I'll defer that for now.

### DIFF
--- a/src/Lucene.Net.Facet/DrillDownQuery.cs
+++ b/src/Lucene.Net.Facet/DrillDownQuery.cs
@@ -158,7 +158,6 @@ namespace Lucene.Net.Facet
         /// Merges (ORs) a new path into an existing AND'd
         /// clause.
         /// </summary>
-        [MethodImpl(MethodImplOptions.NoInlining)]
         private void Merge(string dim, string[] path)
         {
             int index = 0;
--- a/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
+++ b/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
@@ -3087,7 +3087,6 @@ namespace Lucene.Net.Util
             }
         }
 
-        [MethodImpl(MethodImplOptions.NoInlining)]
         protected string GetFullMethodName([CallerMemberName] string memberName = "")
         {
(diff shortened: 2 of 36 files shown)


NightOwl888 (approved):
Thanks. Looks good to merge.

NightOwl888:
Oh, I didn't notice it is still a draft. Feel free to re-request a review when the PR is ready.

paulirwin:
I'll go ahead and split out the remaining investigation work into its own PR, and merge this one as it's a nice seam.

PR #1103, opened 2025-01-19
SWEEP: Make private/internal fields readonly where possible, #661

### LINKED ISSUE #661: Review for IDE0044: Add readonly modifier to private fields
https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0044
https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&id=nikcio_lucenenet&open=AYPAuPfOhbfJOGLOobGt

All private fields **in classes** that are not mutated by the class can be declared `readonly`. This diverges from Java Lucene, so we should add a comment `// LUCENENET: marked readonly`.

All private fields **in structs** should not be declared `readonly`, as this negatively affects performance. Instead, we should use an attribute on all structs to suppress this warning.

Any existing suppressions of IDE0044 in classes may be removed.

paulirwin:

Makes private/internal fields readonly where possible.

Fixes #661

## Description

This PR used ReSharper/Rider's analyzer to find fields that could be made readonly. This PR does not affect any static fields (that's #1079) or protected fields (which might be modified by subclasses). In a few cases, instance fields were made const or static readonly where possible.


NightOwl888 (approved):
Thanks for the PR. This looks good to merge.

PR #1104, opened 2025-01-19
Fix TestBufferedCharFilter.Test_Ready failing test, #1102

### LINKED ISSUE #1102: Failing test: TestBufferedCharFilter.Test_Ready

### Describe the bug

The `TestBufferedCharFilter.Test_Ready` test was missing the `[Test]` attribute, which was added in #1101. This test failed, however. It fails because it requires the provided TextReader to be a CharFilter where IsReady is true, and the test provides a StringReader that is not a CharFilter. We would need to determine if the upstream Harmony test was ported incorrectly, or if we need to find or create a CharFilter implementation that can be "ready" for this test to pass.

### Expected Behavior

The test passes.

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### Lucene.NET Version

_No response_

### .NET Version

_No response_

### Operating System

_No response_

### Anything else?

_No response_

paulirwin:

Fixes the `TestBufferedCharFilter.Test_Ready` failing test.

Fixes #1102

## Description

Background: CharFilter.IsReady returns true if any calls to Read are guaranteed not to block. Otherwise, it means that a call may or may not block.

This test was failing because StringReader is not a CharFilter and thus it did not have an IsReady property. BufferedCharFilter only is ready if the underlying TextReader is a CharFilter and it returns true for IsReady. So given that BufferedCharFilter already properly cascaded the property, this fixes the failing test by adding a new adapter class to adapt the StringReader to a CharFilter so that it is IsReady-aware. Since StringReaders do not block on calls to Read, this always returns true.

This is perhaps a little bit of a leaky abstraction by having to know these details, but I don't think it really harms much in the design. This property only signals whether Read will block or not, and that's not something .NET TextReader implementations care to surface, nor is it used in Lucene. The only other alternatives here would be to have BufferedCharFilter take a CharFilter instead of a TextReader in its constructor, and that would cascade a bunch of mess to callers further up (if it's even possible to do so) and deviate from upstream, or to implement a Reader abstraction to better match Java and have to adapt all TextReaders to that (and all the extra allocations that would go along with that). I don't think either of those alternatives are worth it just for this property. Also, out of all 8 non-test implementations of CharFilter, only BufferedCharFilter overrides IsReady, and it just delegates it to the inner CharFilter (if it is one) anyways. So while this test ensures that it works correctly for BufferedCharFilter for end user code, it doesn't seem to have much utility in Lucene/Lucene.NET, most likely because these calls would usually block (or at least not be guaranteed to block).

### DIFF
--- a/src/Lucene.Net.Tests.Analysis.Common/Analysis/Util/TestBufferedCharFilter.cs
+++ b/src/Lucene.Net.Tests.Analysis.Common/Analysis/Util/TestBufferedCharFilter.cs
@@ -714,19 +714,43 @@ namespace Lucene.Net.Analysis.Util
          * @tests java.io.BufferedReader#ready()
          */
         [Test, LuceneNetSpecific]
-        [AwaitsFix(BugUrl = "https://github.com/apache/lucenenet/issues/1102")] // LUCENENET TODO: fix test
         public void Test_Ready()
         {
             // Test for method boolean java.io.BufferedReader.ready()
             try
             {
-                br = new BufferedCharFilter(new StringReader(testString));
+                // LUCENENET specific: use TestStringReaderCharFilterAdapter to adapt StringReader to be IsReady-aware.
+                br = new BufferedCharFilter(new TestStringReaderCharFilterAdapter(new StringReader(testString)));
                 assertTrue("IsReady returned false", br.IsReady);
             }
             catch (Exception e) when (e.IsIOException())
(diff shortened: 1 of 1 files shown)


NightOwl888 (approved):
Yeah, one adapter in a test is a better solution than having several adapters in production code. Looks good.

Issue #920, opened 2024-02-25
Meta: Add Support unit tests


paulirwin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Task description

This is a meta-issue that should be broken out into separate issues.

From #914 via @NightOwl888:

I also noticed we don't have tests for the following types in the Support folder:

- [ ] [IndexableFieldExtensions](https://github.com/apache/lucenenet/blob/b1476aee4fe21131899c1f43b2e06e25971b3ebe/src/Lucene.Net/Support/Document/Extensions/IndexableFieldExtensions.cs)
- [ ] [ByteArrayOutputStream](https://github.com/apache/lucenenet/blob/b1476aee4fe21131899c1f43b2e06e25971b3ebe/src/Lucene.Net/Support/IO/ByteArrayOutputStream.cs) - Patch to provide ToString() output for Stream.
- [ ] [StreamExtensions](https://github.com/apache/lucenenet/blob/b1476aee4fe21131899c1f43b2e06e25971b3ebe/src/Lucene.Net/Support/IO/StreamExtensions.cs) - These methods were copied from Apache Harmony, and there are tests for them that didn't get ported over. Need to drill down in the code to find the location where these originated from.
- [ ] [ReferenceContext<T>](https://github.com/apache/lucenenet/blob/b1476aee4fe21131899c1f43b2e06e25971b3ebe/src/Lucene.Net/Support/Search/ReferenceContext.cs) - We added this to make it possible to use a using block instead of Acquire() with a try finally around Release(). We should convert this to a ref struct, so it doesn't cause a heap allocation to use. Ref structs cannot implement IDisposable, but the compiler allows a using block if Dispose() exists. Note that ReferenceManagerExtensions can be tested along with it.
- [ ] [LimitedConcurrencyLevelTaskScheduler](https://github.com/apache/lucenenet/blob/b1476aee4fe21131899c1f43b2e06e25971b3ebe/src/Lucene.Net/Support/Threading/LimitedConcurrencyLevelTaskScheduler.cs) - We originally grabbed this from MSDN. We might be able to use tests from Apache Harmony to test this.
- [ ] [ReentrantLock](https://github.com/apache/lucenenet/blob/b1476aee4fe21131899c1f43b2e06e25971b3ebe/src/Lucene.Net/Support/Threading/ReentrantLock.cs) - There are tests in Apache Harmony for this.
[...]

PR #1113, opened 2025-01-21
BREAKING: Remove SystemConsole class, #920, #936

paulirwin:

Removes the SystemConsole class from Lucene.NET.

Related: #920, #936

## Description

#920 had an item for adding unit tests for SystemConsole, but after discussion with @NightOwl888, we determined that this class is not needed in Lucene.NET, and only serves as an unnecessary abstraction. Users that wish to override the Console.Out/Error behavior can use Console.SetOut or Console.SetError to replace them with a different TextWriter. Most of the usage is in tests anyways, and the usage in non-test code is behind VERBOSE flags. 

#936 raised an issue with MAUI support, but this was only a problem with Console.In. Console.Out and Console.Error work fine in MAUI.

This is a breaking change because SystemConsole was previously public.


NightOwl888 (changes requested):
Thanks for the PR.

We can now remove `FEATURE_ARGITERATOR`, since it only existed in `SystemConsole`.

NightOwl888 (approved):

PR #1119, opened 2025-01-22
Add unit tests for LimitedConcurrencyLevelTaskScheduler, #1110

### LINKED ISSUE #1110: Add Support unit tests for LimitedConcurrencyLevelTaskScheduler

### Task description

From #920: "We originally grabbed this from MSDN. We might be able to use tests from Apache Harmony to test this."

paulirwin:

Adds unit tests for LimitedConcurrencyLevelTaskScheduler.

Fixes #1110

## Description

This adds some Apache Harmony tests for ThreadPoolExecutor and adapts them for testing LimitedConcurrencyLevelTaskScheduler (used in Lucene.NET in place of ThreadPoolExecutor).

This class, being a TaskScheduler, is only useful when used with i.e. a TaskFactory, so it does not have a very testable public API surface. So to exercise this class, some extension methods were added in the JSR166TestCase file to emulate a ThreadPoolExecutor's API surface. This keeps track of task state using a ConditionalWeakTable keyed on TaskScheduler, so that when the TaskScheduler is garbage collected, its state is cleaned up. The state is needed to track completed tasks, since LimitedConcurrencyLevelTaskScheduler removes/dequeues tasks when they are completed. This also implements the `joinPool` method in JSR166TestCase via waiting for all tracked tasks for the scheduler with a timeout.

This does not port all of the tests from Harmony, since most are irrelevant, but it ports enough to test executing tasks, concurrency level, the number of running/completed tasks, shutdown state, etc.

As noted in #1110, `FEATURE_THREADPOOL_UNSAFEQUEUEWORKITEM` was removed as all targets support this feature.

### DIFF
--- a/src/Lucene.Net/Support/Threading/LimitedConcurrencyLevelTaskScheduler.cs
+++ b/src/Lucene.Net/Support/Threading/LimitedConcurrencyLevelTaskScheduler.cs
@@ -1,50 +1,50 @@
 ﻿/*
 MICROSOFT LIMITED PUBLIC LICENSE version 1.1
-This license governs use of code marked as "sample" or "example" available on this web site 
-without a license agreement, as provided under the section above titled 
-"NOTICE SPECIFIC TO SOFTWARE AVAILABLE ON THIS WEB SITE." If you use such 
-code (the "software"), you accept this license. If you do not accept the 
+This license governs use of code marked as "sample" or "example" available on this web site
+without a license agreement, as provided under the section above titled
+"NOTICE SPECIFIC TO SOFTWARE AVAILABLE ON THIS WEB SITE." If you use such
+code (the "software"), you accept this license. If you do not accept the
 license, do not use the software.
 
 1. Definitions
-The terms "reproduce," "reproduction," "derivative works," and "distribution" have the 
+The terms "reproduce," "reproduction," "derivative works," and "distribution" have the
 same meaning here as under U.S. copyright law.
 A "contribution" is the original software, or any additions or changes to the software.
 A "contributor" is any person that distributes its contribution under this license.
 "Licensed patents" are a contributor’s patent claims that read directly on its contribution.
 
 2. Grant of Rights
-(A) Copyright Grant - Subject to the terms of this license, including the license conditions 
-and limitations in section 3, each contributor grants you a non-exclusive, worldwide, 
-royalty-free copyright license to reproduce its contribution, prepare derivative works 
+(A) Copyright Grant - Subject to the terms of this license, including the license conditions
+and limitations in section 3, each contributor grants you a non-exclusive, worldwide,
+royalty-free copyright license to reproduce its contribution, prepare derivative works
 of its contribution, and distribute its contribution or any derivative works that you create.
-(B) Patent Grant - Subject to the terms of this license, including the license conditions 
-and limitations in section 3, each contributor grants you a non-exclusive, worldwide, 
-royalty-free license under its licensed patents to make, have made, use, sell, 
-offer for sale, import, and/or otherwise dispose of its contribution in the 
+(B) Patent Grant - Subject to the terms of this license, including the license conditions
+and limitations in section 3, each contributor grants you a non-exclusive, worldwide,
+royalty-free license under its licensed patents to make, have made, use, sell,
+offer for sale, import, and/or otherwise dispose of its contribution in the
 software or derivative works of the contribution in the software.
 
 3. Conditions and Limitations
-(A) No Trademark License- This license does not grant you rights to use any contributors’ 
+(A) No Trademark License- This license does not grant you rights to use any contributors’
 name, logo, or trademarks.
-(B) If you bring a patent claim against any contributor over patents that you claim are 
-infringed by the software, your patent license from such contributor to the software 
+(B) If you bring a patent claim against any contributor over patents that you claim are
+infringed by the software, your patent license from such contributor to the software
 ends automatically.
-(C) If you distribute any portion of the software, you must retain all copyright, patent, 
+(C) If you distribute any portion of the software, you must retain all copyright, patent,
 trademark, and attribution notices that are present in the software.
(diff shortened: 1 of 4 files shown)


src/Lucene.Net/Support/Threading/LimitedConcurrencyLevelTaskScheduler.cs
```diff
@@ -208,10 +203,17 @@ protected sealed override IEnumerable<Task> GetScheduledTasks()
             }
         }
 
-        // Stops this TaskScheduler from queuing new tasks.
+        /// <summary>
+        /// Stops this TaskScheduler from queuing new tasks.
+        /// </summary>
         public void Shutdown()
```
NightOwl888, line 209: After taking a step back from this and considering that we are working on support for `CancellationToken`s (#1080), I wonder if we should consider removing `Shutdown()` and always use `CancellationToken` instead?

If we do keep the `Shutdown()` method, there are some things to consider:

1. Should we make `LimitedConcurrencyLevelTaskScheduler` public? The BCL doesn't have one and it is provided in the docs for `TaskScheduler`. But there is no `ShutDown()` on that implementation.
2. There are some tests where `Shutdown()` was called in Lucene that would likely perform better if we use it, since those lines were often commented out in Lucene.NET. However, using `CancellationToken` could be an alternative way to handle those cases.
    -  https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/core/src/test/org/apache/lucene/search/TestBooleanQuery.java#L177-L178
    - https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/core/src/test/org/apache/lucene/search/TestSearcherManager.java#L291-L292
    - https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/suggest/src/test/org/apache/lucene/search/spell/TestSpellChecker.java#L435-L437
    - https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/test-framework/src/java/org/apache/lucene/index/ThreadedIndexingAndSearchingTestCase.java#L645-L646
    - https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/core/src/test/org/apache/lucene/search/TestIndexSearcher.java#L119
    - https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/test-framework/src/java/org/apache/lucene/util/LuceneTestCase.java#L1570

The upside of using `Shutdown()` is that it aligns more closely to Lucene.

The upside of using `CancellationToken` is that it is more fine-grained and can potentially shut down a process quicker, since it will be able to intervene inside of the running task instead of prior to queuing the task.

Allowing a user-defined `TaskScheduler` still seems like a good idea, but since Microsoft didn't allow them to be cancelled, it seems like we should follow the `CancellationToken` approach instead of trying to shoehorn a way to cancel a `TaskScheduler`.

paulirwin: I think that could be worthy of discussion in #1080 and very relevant there holistically. This PR just brings over the Harmony tests, and I haven't changed any Shutdown behavior yet. I'd say let's split that out as part of #1080 feedback.

NightOwl888 (approved):

NightOwl888:
It looks like there is a problem with these tests on:

- `net9.0` x64, Windows
- `net9.0` x64 Linux
- `net6.0` x64 Windows

https://dev.azure.com/LuceneNET-Temp/Lucene.NET/_build/results?buildId=2410&view=ms.vss-test-web.build-test-results-tab

```
Expected: 1, Actual: 0
(Test: Lucene.Net.Support.Threading.TestLimitedConcurrencyLevelTaskScheduler.TestGetActiveCount)


Option 1:


[assembly: Lucene.Net.Util.RandomSeed("0x9054fc2057c057c9")]
[assembly: NUnit.Framework.SetCulture("ia")]


Option 2:


<RunSettings>
  <TestRunParameters>
    <Parameter name="tests:seed" value="0x9054fc2057c057c9" />
    <Parameter name="tests:culture" value="ia" />
  </TestRunParameters>
</RunSettings>
Option 3:


{
"tests": {
"seed": "0x9054fc2057c057c9",
"culture": "ia"
}
}


Fixture Test Values

Random Seed:           0x9054fc2057c057c9
Culture:               ia
Time Zone:             (UTC-04:00) Georgetown, La Paz, Manaus, San Juan
Default Codec:         Lucene45 (Lucene45RWCodec)
Default Similarity:    RandomSimilarityProvider(queryNorm=False,coord=crazy): {}


System Properties

Nightly:               False
Weekly:                False
Slow:                  True
Awaits Fix:            False
Directory:             random
Verbose:               False
Random Multiplier:     1
```
-------------------------

```
Expected: 1, Actual: 0
(Test: Lucene.Net.Support.Threading.TestLimitedConcurrencyLevelTaskScheduler.TestGetCompletedTaskCount)


Option 1:


[assembly: Lucene.Net.Util.RandomSeed("0x9054fc2057c057c9")]
[assembly: NUnit.Framework.SetCulture("ia")]


Option 2:


<RunSettings>
  <TestRunParameters>
    <Parameter name="tests:seed" value="0x9054fc2057c057c9" />
    <Parameter name="tests:culture" value="ia" />
  </TestRunParameters>
</RunSettings>
Option 3:


{
"tests": {
"seed": "0x9054fc2057c057c9",
"culture": "ia"
}
}


Fixture Test Values

Random Seed:           0x9054fc2057c057c9
Culture:               ia
Time Zone:             (UTC-04:00) Georgetown, La Paz, Manaus, San Juan
Default Codec:         Lucene45 (Lucene45RWCodec)
Default Similarity:    RandomSimilarityProvider(queryNorm=False,coord=crazy): {}


System Properties

Nightly:               False
Weekly:                False
Slow:                  True
Awaits Fix:            False
Directory:             random
Verbose:               False
Random Multiplier:     1

```

PR #1120, opened 2025-01-22
Add unit tests for ExceptionExtensions, #1112

### LINKED ISSUE #1112: Add Support unit tests for ExceptionExtensions

### Task description

From #920: [ExceptionExtensions](https://github.com/apache/lucenenet/blob/b1476aee4fe21131899c1f43b2e06e25971b3ebe/src/Lucene.Net/Support/Util/ExceptionExtensions.cs) - Apache Harmony should have tests for these.

paulirwin:

Add unit tests for ExceptionExtensions.

Fixes #1112 

## Description

Adds unit tests for Lucene.Net.Util.ExceptionExtensions around handling suppressed extensions.

### DIFF
--- a/src/Lucene.Net.Tests/Util/TestExceptionExtensions.cs
+++ b/src/Lucene.Net.Tests/Util/TestExceptionExtensions.cs
@@ -1,0 +1,76 @@
+using NUnit.Framework;
+using System;
+using System.Linq;
+
+#nullable enable
+
+namespace Lucene.Net.Util
+{
+    /*
+     * Licensed to the Apache Software Foundation (ASF) under one or more
+     * contributor license agreements.  See the NOTICE file distributed with
+     * this work for additional information regarding copyright ownership.
+     * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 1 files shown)


NightOwl888 (approved):
Thanks for the PR.

During the review I noticed that `addSuppressed()`, `getSuppressed()`, and `printStackTrace()` are all synchronized. This makes sense because calling `addSuppressed()` could change the result of the other two.

https://github.com/EricChows/JDK-1.8-sourcecode/blob/master/java/lang/Throwable.java#L1041-L1080

I can't think of a use case where we actually need thread safety here. We are dealing with exceptions in catch blocks where it is impossible for 2 threads to interact with  a single exception unless parallel execution is started inside of the catch block (or after if the exception is set to a variable).

But, if you wish to future-proof the implementation or know of some place where this will cause a problem, I have no objections to adding synchronization (and tests).

paulirwin:
I'll skip synchronization for now, it can always be added later as a fix without breaking the contract if there are any issues with it. I doubt there will be, though.

PR #1125, opened 2025-01-27
Added RepeatAttribute that is aware of RandomizedContext

### LINKED ISSUE #1106: Create [Repeat] attribute that is RandomizedContext aware

### Task description

The NUnit `[Repeat]` attribute is unaware of our `RandomizedContext` class. As a result, each iteration reuses the `RandomizedContext.RandomGenerator` instance without first resetting the seed. While this technically still makes the initial seed repeatable, it requires the `[Repeat]` attribute to repeat the result and it also requires the same number of iterations that the first result took to fail. This isn't very helpful when debugging, since the NUnit `[Repeat]` attribute doesn't report which iteration caused a failure and it is cumbersome to wrap a test in a loop so it can be repeated x number of times before stepping through in the iteration that is known to fail.

To fix this, we should create a custom `[Repeat]` attribute.

Much like our existing attributes, it should be named `RepeatAttribute` and nested within `LuceneTestCase`. This ensures that when participating in the test framework, adding `[Repeat(100)]` to a test will use our custom `RepeatAttribute` by default even if the `NUnit.Framework` namespace is already imported and there won't be a conflict. If test assembly has a class with tests that doesn't subclass `LuceneTestCase`, NUnit's `[Repeat(100)]` attribute will be the default.

The `[Repeat]` attribute should update the `RandomizedContext.TestSeed` on each iteration. `RandomizedContext.RandomSeedAsHex` should include both the `RandomizedContext.RandomSeed` and `RandomizedContext.TestSeed` separated by a colon. This ensures the correct test seed is used to repeat the exact random test run. When used without the `[Repeat]` attribute, this makes debugging much easier.

NightOwl888:

Fixes #1106

## Description

This creates a new `RepeatAttribute` that is nested within `LuceneTestCase`. In addition to repeating the test X number of times, this attribute ensures that each iteration resets the test seed on `RandomizedContext.RandomGenerator` so failed test conditions can be replicated without the need to apply the `RepeatAttribute`.

This sets a precedent for how to test other NUnit extensions (such as attributes) so we can also add tests to confirm they behave as expected. Some bits of the NUnit test approach were added so we can integration test the `RepeatAttribute` to confirm the correct behavior. The commit cdfd62bbc8e4f89382fd2624f6889d0750163357 includes additional asserts that verify:

- The same random seed is used for the duration of the test
- The same test seed is used for the duration of the test
- The test seed changes from one iteration to the next

Note that `RepeatAttribute` does not repeat the `OneTimeSetUp()` or `OneTimeTearDown()` during each iteration, so every repeated try will use the same culture, time zone, similarity, and codec. Perhaps we can do that at some future point, but it is more complicated to maintain the `TestExectutionContext`, `RandomizedContext` and seeds to make that happen.

The format of the random seed string has also changed to use 2 hexadecimal numbers separated by a `:`. However, existing seeds with one hexidecimal number can still be parsed to replicate test failures. This not only allows us to make features that update the seed, but also will make test replicability more stable across source code updates on tests.


src/dotnet/Lucene.Net.TestFramework.TestData.NUnit/TestUtilities/TestBuilder.cs
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        public static ITestResult ExecuteWorkItem(WorkItem work)
+        {
+            work.Execute();
+
+            // TODO: Replace with an event - but not while method is static
+            while (work.State != WorkItemState.Complete)
+            {
+                Thread.Sleep(1);
```
paulirwin, line 224: I was debating whether or not to comment this, because it could end up being pretty wasteful especially on multi-core laptops on battery power when developing this, but I'll post it for posterity anyways. I was thinking about whether SpinWait might be better here than sleeping for 1ms to avoid the thread blocking overhead. But perhaps not; many/most of our tests take longer than single-digit milliseconds, so that might be too wasteful in terms of CPU if it almost always exhausts its spinning state before yielding. (SpinWait has some optimizations that will prevent it from spinning too long or unnecessarily, and will always yield on single-core computers, i.e. containers/build agents that are single core.) If not SpinWait, it might also be worth considering an exponential backoff up to some maximum delay, maybe 256ms (2^8), so that it's not in a 1ms-tight loop. Or, another option, maybe we could Monitor.Wait/PulseAll on the WorkItem or some state within it. Either way, I just wonder if there's some better option than this tight sleep loop here.

paulirwin (approved):
This looks great, nicely done. The actual implementation is refreshingly simple! (Although I'm sure the work to get there was not.) I left one comment that could be looked into, but not a must-have change.

src/dotnet/Lucene.Net.TestFramework.TestData.NUnit/TestUtilities/TestBuilder.cs
paulirwin: Disregard this, I didn't realize this was only used in the unit tests for the RepeatAttribute itself 😄 Worth keeping in mind for any future tight-sleep-loops though.

NightOwl888: Good find. This was copied as-is from NUnit as it is part of their test framework for testing NUnit's attributes. It won't be shipped to end users, it is a separate library (separated out from the test library because it contains tests that we want to avoid having NUnit run and run manually).

Also, we don't currently test it in a way that it switches to a background thread, so this will only run once. Parallel execution would require that we copy over the `[assembly: Parallelizable(ParallelScope.Fixtures)]` attribute from the original test assembly to the test project. But we currently don't need it because Lucene's design has some static state that won't allow parallel testing. When we move this over to RandomizedTesting, we could potentially support parallel testing, though.

Issue #1117, opened 2025-01-21
Add Support unit tests for ConcurrentHashSet


paulirwin:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Task description

From #920: Apache Harmony should have tests for this. Note that not all `ISet<T>` members are implemented. If they were, we could potentially move this to J2N. Note that 1 test was added in [#938](https://github.com/apache/lucenenet/pull/938/files#r1597522965), but we should be more thorough than this.

Issue #1123, opened 2025-01-23
Occasional failing tests on LimitedConcurrencyLevelTaskScheduler


paulirwin:
It looks like there is a problem with these tests on:

- `net9.0` x64, Windows
- `net9.0` x64 Linux
- `net6.0` x64 Windows

https://dev.azure.com/LuceneNET-Temp/Lucene.NET/_build/results?buildId=2410&view=ms.vss-test-web.build-test-results-tab

```
Expected: 1, Actual: 0
(Test: Lucene.Net.Support.Threading.TestLimitedConcurrencyLevelTaskScheduler.TestGetActiveCount)


To reproduce this test result:


Option 1:


Apply the following assembly-level attributes:


[assembly: Lucene.Net.Util.RandomSeed("0x9054fc2057c057c9")]
[assembly: NUnit.Framework.SetCulture("ia")]


Option 2:


Use the following .runsettings file:


<RunSettings>
  <TestRunParameters>
    <Parameter name="tests:seed" value="0x9054fc2057c057c9" />
    <Parameter name="tests:culture" value="ia" />
  </TestRunParameters>
</RunSettings>
Option 3:


Create the following lucene.testsettings.json file somewhere between the test assembly and the root of your drive:


{
"tests": {
"seed": "0x9054fc2057c057c9",
"culture": "ia"
}
}


Fixture Test Values

Random Seed:           0x9054fc2057c057c9
Culture:               ia
Time Zone:             (UTC-04:00) Georgetown, La Paz, Manaus, San Juan
Default Codec:         Lucene45 (Lucene45RWCodec)
Default Similarity:    RandomSimilarityProvider(queryNorm=False,coord=crazy): {}


System Properties

Nightly:               False
Weekly:                False
Slow:                  True
Awaits Fix:            False
Directory:             random
Verbose:               False
Random Multiplier:     1
```
-------------------------

```
Expected: 1, Actual: 0
(Test: Lucene.Net.Support.Threading.TestLimitedConcurrencyLevelTaskScheduler.TestGetCompletedTaskCount)


To reproduce this test result:


Option 1:


Apply the following assembly-level attributes:


[assembly: Lucene.Net.Util.RandomSeed("0x9054fc2057c057c9")]
[assembly: NUnit.Framework.SetCulture("ia")]


Option 2:


Use the following .runsettings file:


<RunSettings>
  <TestRunParameters>
    <Parameter name="tests:seed" value="0x9054fc2057c057c9" />
[...]

PR #1129, opened 2025-02-16
Add Support unit tests for ConcurrentSet

### LINKED ISSUE #1118: Add Support unit tests for ConcurrentSet

### Task description

From #920: This is a concurrent wrapper for any `ISet<T>`. It should pass the same tests as `ConcurrentHashSet<T>`. It is mainly used for wrapping `LinkedHashSet<T>`. If we had a `ConcurrentLinkedHashSet<T>` in J2N, we could eliminate this class.

paulirwin:

Adds support unit tests for ConcurrentSet

Fixes #1118

## Description

As noted in #1118, this type should pass the same tests as ConcurrentSet (apart from the constructor differences). So this PR takes the tests created in #1117 and refactors them out into a base class that can be used for both ConcurrentHashSet and ConcurrentSet test fixtures. The two `protected abstract` methods are factory methods to create the specific implementation being tested. The two constructor tests that are specific to ConcurrentHashSet only exist on TestConcurrentHashSet.

Additionally, this PR removes a few more usages of SystemTypesHelpers in those shared tests.

### DIFF
--- a/src/Lucene.Net.Tests/Support/TestConcurrentSet.cs
+++ b/src/Lucene.Net.Tests/Support/TestConcurrentSet.cs
@@ -4,6 +4,7 @@
 
 using Lucene.Net.Support;
 using System.Collections.Generic;
+using JCG = J2N.Collections.Generic;
 
 namespace Lucene.Net
 {
@@ -34,9 +35,9 @@ namespace Lucene.Net
     public class TestConcurrentSet : BaseConcurrentSetTestCase
     {
         protected override ISet<T> NewSet<T>()
-            => new ConcurrentSet<T>(new HashSet<T>());
+            => new ConcurrentSet<T>(new JCG.LinkedHashSet<T>());
 
-        protected override ISet<T> NewSet<T>(IEnumerable<T> collection)
-            => new ConcurrentSet<T>(new HashSet<T>(collection));
+        protected override ISet<T> NewSet<T>(ISet<T> set)
+            => new ConcurrentSet<T>(set);
     }
 }
--- a/src/Lucene.Net.Tests/Support/TestConcurrentSet.cs
+++ b/src/Lucene.Net.Tests/Support/TestConcurrentSet.cs
@@ -3,8 +3,6 @@
 // https://github.com/apache/harmony/blob/02970cb7227a335edd2c8457ebdde0195a735733/classlib/modules/luni/src/test/api/common/org/apache/harmony/luni/tests/java/util/CollectionsTest.java
 
 using Lucene.Net.Support;
-using NUnit.Framework;
-using System;
 using System.Collections.Generic;
 
 namespace Lucene.Net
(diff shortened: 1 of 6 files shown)


src/Lucene.Net.Tests/Support/TestConcurrentSetBase.cs
```diff
@@ -0,0 +1,456 @@
...
+    /// <summary>
+    /// Base class for tests for <see cref="TestConcurrentHashSet"/> and <see cref="TestConcurrentSet"/>.
+    /// </summary>
+    /// <remarks>
+    /// Some tests (those not marked with <see cref="LuceneNetSpecificAttribute"/>)
+    /// are adapted from Apache Harmony's ConcurrentHashMapTest class. This class
+    /// tests ConcurrentHashMap, which is a dictionary, but the key behavior is
+    /// similar to <see cref="ConcurrentHashSet{T}"/> and <see cref="ConcurrentSet{T}"/>.
+    /// </remarks>
+    public abstract class TestConcurrentSetBase : JSR166TestCase
```
NightOwl888, line 45: For consistency, let's name this `BaseConcurrentSetTestCase`.

src/Lucene.Net.Tests/Support/TestConcurrentSet.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Tests for <see cref="ConcurrentSet{T}"/>.
+    /// </summary>
+    /// <remarks>
+    /// See the <see cref="TestConcurrentSetBase"/> class for most of the test cases.
+    /// This class specializes the tests for the <see cref="ConcurrentSet{T}"/> class.
+    /// </remarks>
+    public class TestConcurrentSet : TestConcurrentSetBase
+    {
+        protected override ISet<T> NewSet<T>()
+            => new ConcurrentSet<T>(new HashSet<T>());
```
NightOwl888, line 37: There are 2 main benefits for having `ConcurrentSet<T>`:

1. It enables concurrency on alternative set implementations, such as `LinkedHashSet<T>` or `OrderedHashSet<T>`.
2. It exposes the shared lock (through the `SyncRoot` property), so external operations (such as atomic enumeration) can be synchronized with the rest of the set.

> NOTE: Something smells funny about the implementation of the `ConcurentSet<T>.SyncRoot` property. It seems like the `System.Threading.Interlocked.CompareExchange<object?>(ref syncRoot, new object(), null);` line shouldn't be there or it should be setup to correctly update the value of `syncRoot`.

The `ConcurrentHashSet<T>` has the advantage of not locking the entire set so in theory, will perform better.

All of that said, it seems like it would be a more useful test if the [`OrderedHashSet<T>`](https://github.com/NightOwl888/J2N/issues/95) were tested here. But since it doesn't yet exist, perhaps we could settle for `LinkedHashSet<T>` for now. 

Better still, create another base class to provide the `ISet<T>` implementation to wrap with multiple subclasses to test each wrapped instance separately.

NightOwl888 (changes requested):

src/Lucene.Net.Tests/Support/TestConcurrentSet.cs
NightOwl888: It would also be good to have a concurrency test that uses the `SyncRoot` property to synchronize external work while other threads are attempting to use the set, just to ensure it is implemented correctly.

paulirwin: This is complete:

1. Removed the lazy-initialization logic of the `syncRoot` field by making it `private readonly object syncRoot = new object();`. This property was (correctly) accessed by every method in the type, so there's no need to lazy-initialize it, as it is expected that if you're creating the set, you're going to use it. The "cost" of initializing a new object once per set is lower than having to check if it has been initialized _twice_ in every set method. This should improve performance.
2. Fixed a thread-safety issue in GetEnumerator where the array was initialized based on the set count outside of the synchronization, so it is possible that the set could be modified between initializing the enumerator set and copying the array, possibly resulting in an error. 
3. Added unit test for concurrently synchronizing on the public SyncRoot property while simultaneously calling the ConcurrentSet Add method.
4. Added unit test for ensuring GetEnumerator returns an enumerator that does not fail while the set is being modified.
5. Changed the default inner set created by the parameterless NewSet method to be LinkedHashSet.
6. Created two versions of TestSynchronizedSet, one that uses HashSet as the inner set and one that uses LinkedHashSet.

I hope we do not need to go further for this PR.

NightOwl888 (commented):
I ran the tests on Azure DevOps. There are failures now on every run. Most of these are due to #1123. Could you have a look at that?

- https://dev.azure.com/LuceneNET-Temp/Lucene.NET/_build/results?buildId=2412&view=results
- https://dev.azure.com/lucene-net-temp2/Lucene.NET/_build/results?buildId=1092&view=results
- https://dev.azure.com/lucene-net-temp3/Lucene.NET/_build/results?buildId=896&view=ms.vss-test-web.build-test-results-tab
- https://dev.azure.com/lucene-net-temp4/Main/_build/results?buildId=318&view=ms.vss-test-web.build-test-results-tab


This one is new, but it might be a one-off:

```
  Passed Lucene.Net.Store.TestDirectory.TestLUCENENET521 [100 ms]
  Passed Lucene.Net.Store.TestDirectory.TestNotDirectory [3 ms]
  Passed Lucene.Net.Store.TestDirectory.TestRAMDirectoryFilter [< 1 ms]
[ERROR] OneTimeTearDown: An exception occurred during CleanupTemporaryFiles() in Lucene.Net.Store.TestDirectory:
System.IO.IOException: Could not remove the following files (in the order of attempts):
   C:\Users\VssAdministrator\AppData\Local\Temp\LuceneTemp\LUCENENET521-ujwvk5d1\_0.cfs
   C:\Users\VssAdministrator\AppData\Local\Temp\LuceneTemp\LUCENENET521-ujwvk5d1\_0.cfx
   C:\Users\VssAdministrator\AppData\Local\Temp\LuceneTemp\LUCENENET521-ujwvk5d1

   at Lucene.Net.Util.TestUtil.Rm(FileSystemInfo[] locations) in /_/src/Lucene.Net.TestFramework/Util/TestUtil.cs:line 66
   at Lucene.Net.Util.LuceneTestCase.CleanupTemporaryFiles() in /_/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs:line 3172
   at Lucene.Net.Util.LuceneTestCase.OneTimeTearDown() in /_/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs:line 1039
  Passed Lucene.Net.Store.TestDirectory.TestThreadSafety [7 s]
  Passed Lucene.Net.Store.TestFileSwitchDirectory.TestBasic [164 ms]
  Passed Lucene.Net.Store.TestFileSwitchDirectory.TestCompoundFileAppendTwice [34 ms]
```

- https://dev.azure.com/shad0962/Experiments/_build/results?buildId=2578&view=results

NightOwl888 (approved):
Given the test failures should now be resolved by #1131, this looks good to merge.

PR #1138, opened 2025-03-08
Fix random failing test with MMapDirectory capacity out of range, #1090

### LINKED ISSUE #1090: Random test failure: TestNeverDelete.TestIndexing

### Describe the bug

Got a random test failure on [this PR run](https://github.com/apache/lucenenet/actions/runs/12639130426/job/35216826838?pr=1089), ubuntu-latest, net9.0, x64. Re-running the test (with new random values) succeeded.

```
Error Message:
   System.ArgumentOutOfRangeException : The capacity may not be smaller than the file size. (Parameter 'capacity')
(Test: Lucene.Net.Index.TestNeverDelete.TestIndexing)

Option 1:

[assembly: Lucene.Net.Util.RandomSeed("0x778d6589462b9ad4")]
[assembly: NUnit.Framework.SetCulture("es-BO")]

Option 2:

<RunSettings>
  <TestRunParameters>
    <Parameter name="tests:seed" value="0x778d6589462b9ad4" />
    <Parameter name="tests:culture" value="es-BO" />
  </TestRunParameters>
</RunSettings>

Option 3:

{
  "tests": {
     "seed": "0x778d6589462b9ad4",
     "culture": "es-BO"
  }
}

  Stack Trace:
     at System.IO.MemoryMappedFiles.MemoryMappedFile.CreateCore(SafeFileHandle fileHandle, String mapName, HandleInheritability inheritability, MemoryMappedFileAccess access, MemoryMappedFileOptions options, Int64 capacity, Int64 fileSize)
   at Lucene.Net.Store.MMapDirectory.MMapIndexInput..ctor(MMapDirectory outerInstance, String resourceDescription, FileStream fc) in /home/runner/work/lucenenet/lucenenet/src/Lucene.Net/Store/MMapDirectory.cs:line 240
   at Lucene.Net.Store.MMapDirectory.OpenInput(String name, IOContext context) in /home/runner/work/lucenenet/lucenenet/src/Lucene.Net/Store/MMapDirectory.cs:line 186
   at Lucene.Net.Util.LuceneTestCase.SlowFileExists(Directory dir, String fileName) in /home/runner/work/lucenenet/lucenenet/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs:line 2938
   at Lucene.Net.Store.MockDirectoryWrapper.OpenInput(String name, IOContext context) in /home/runner/work/lucenenet/lucenenet/src/Lucene.Net.TestFramework/Store/MockDirectoryWrapper.cs:line 799
   at Lucene.Net.Index.SegmentInfos.Read(Directory directory, String segmentFileName) in /home/runner/work/lucenenet/lucenenet/src/Lucene.Net/Index/SegmentInfos.cs:line 513
   at Lucene.Net.Index.SegmentInfos.FindSegmentsFile.Run(IndexCommit commit) in /home/runner/work/lucenenet/lucenenet/src/Lucene.Net/Index/SegmentInfos.cs:line 1158
   at Lucene.Net.Index.SegmentInfos.Read(Directory directory) in /home/runner/work/lucenenet/lucenenet/src/Lucene.Net/Index/SegmentInfos.cs:line 625
   at Lucene.Net.Index.StandardDirectoryReader.DoOpenNoWriter(IndexCommit commit) in /home/runner/work/lucenenet/lucenenet/src/Lucene.Net/Index/StandardDirectoryReader.cs:line 380
   at Lucene.Net.Index.DirectoryReader.OpenIfChanged(DirectoryReader oldReader) in /home/runner/work/lucenenet/lucenenet/src/Lucene.Net/Index/DirectoryReader.cs:line 174
   at Lucene.Net.Index.TestNeverDelete.TestIndexing() in /home/runner/work/lucenenet/lucenenet/src/Lucene.Net.Tests/Index/TestNeverDelete.cs:line 88
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

### Expected Behavior

The test passes

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### Lucene.NET Version

_No response_

### .NET Version

9

### Operating System

ubuntu-latest

### Anything else?

x64

paulirwin:

Fixes a random failing test when the MMapDirectory's MemoryMappedFile capacity is out of range, plus some other related logic.

Fixes #1090

## Description

See discussion on #1090. The solution I came up with is to pass 0 for the capacity, per the docs on that method: "Specify 0 to set the capacity to the size of the file on disk." This should avoid the problem where the provided capacity was smaller than the file size, likely due to a concurrency bug.

This also fixes an "adjust" variable hack to allocate an empty buffer instead of subtracting 1 from the offset.

### DIFF
--- a/src/Lucene.Net/Store/MMapDirectory.cs
+++ b/src/Lucene.Net/Store/MMapDirectory.cs
@@ -314,7 +314,7 @@ namespace Lucene.Net.Store
                 input.memoryMappedFile = MemoryMappedFile.CreateFromFile(
                     fileStream: fc,
                     mapName: null,
-                    capacity: length,
+                    capacity: 0, // use the full length of the file
                     access: MemoryMappedFileAccess.Read,
 #if FEATURE_MEMORYMAPPEDFILESECURITY
                     memoryMappedFileSecurity: null,
--- a/src/Lucene.Net/Store/MMapDirectory.cs
+++ b/src/Lucene.Net/Store/MMapDirectory.cs
@@ -330,15 +330,15 @@ namespace Lucene.Net.Store
 
                 // LUCENENET: We get an UnauthorizedAccessException if we create a 0 byte file at the end of the range.
                 // See: https://stackoverflow.com/a/5501331
-                // We can fix this by moving back 1 byte on the offset if the bufSize is 0.
-                int adjust = 0;
-                if (bufSize == 0 && bufNr == (nrBuffers - 1) && (offset + bufferStart) > 0)
+                // We can fix this by using an empty ByteBuffer if the buffer size is 0.
+                if (bufSize == 0 && bufNr == (nrBuffers - 1))
                 {
-                    adjust = 1;
+                    buffers[bufNr] = ByteBuffer.Allocate(0).AsReadOnlyBuffer();
+                    break;
                 }
 
                 buffers[bufNr] = input.memoryMappedFile.CreateViewByteBuffer(
-                    offset: (offset + bufferStart) - adjust,
+                    offset: offset + bufferStart,
                     size: bufSize,
                     access: MemoryMappedFileAccess.Read);
                 bufferStart += bufSize;


src/Lucene.Net/Store/MMapDirectory.cs
```diff
@@ -330,15 +330,15 @@ internal virtual ByteBuffer[] Map(MMapIndexInput input, FileStream fc, long offs
...
+                if (bufSize == 0 && bufNr == (nrBuffers - 1))
                 {
-                    adjust = 1;
+                    buffers[bufNr] = ByteBuffer.Allocate(0).AsReadOnlyBuffer();
+                    break;
                 }
 
                 buffers[bufNr] = input.memoryMappedFile.CreateViewByteBuffer(
-                    offset: (offset + bufferStart) - adjust,
+                    offset: offset + bufferStart,
```
NightOwl888, line 341: Note that offset could overflow `long`. But I am not sure whether it is worth the cost of performance to do an extra check just to throw an `ArgumentOutOfRangeException` with a more sensible message than "less than zero". Microsoft sometimes throws "less than zero" exceptions for overflow cases to reduce the number of guard operations.

src/Lucene.Net/Store/MMapDirectory.cs
```diff
@@ -314,7 +314,7 @@ internal virtual ByteBuffer[] Map(MMapIndexInput input, FileStream fc, long offs
                 input.memoryMappedFile = MemoryMappedFile.CreateFromFile(
                     fileStream: fc,
                     mapName: null,
-                    capacity: length,
+                    capacity: 0, // use the full length of the file
```
NightOwl888, line 317: Given how unpredictable MemoryMappedFile behavior is across platforms, I am not sure we should only rely on the documentation for this. There were several problems with MemoryMappedFile tests on J2N when using Xamarin Android (Mono), but fortunately none of them affected Lucene.NET. The behavior seems to be broken on Mono: https://stackoverflow.com/q/9895384

Let's just use `length` here to be sure we have the full file length instead of falling back on the specific implementation, which may or may not behave as documented.

NightOwl888 (changes requested):

src/Lucene.Net/Store/MMapDirectory.cs
paulirwin: We can't use `length` here because that is precisely what was causing the exception that this PR aims to fix. Somehow the `fc.Length` did not equal `length`. However, what the underlying code is doing on .NET Framework and modern .NET is that it is setting the capacity to `fileStream.Length` if it's zero, so we can just pass that to be on the safe side. That way it will always be valid.

src/Lucene.Net/Store/MMapDirectory.cs
paulirwin: `CreateViewByteBuffer` will already throw if negative, so that should be sufficient here.

src/Lucene.Net/Store/MMapDirectory.cs
```diff
@@ -314,7 +314,7 @@ internal virtual ByteBuffer[] Map(MMapIndexInput input, FileStream fc, long offs
                 input.memoryMappedFile = MemoryMappedFile.CreateFromFile(
                     fileStream: fc,
                     mapName: null,
-                    capacity: length,
+                    capacity: fc.Length, // use the full length of the file
```
NightOwl888, line 317: Please revert this back to the passed in `length`. Again, there is no telling how broken the implementation is that we are dealing with. `FileStream.Length` may not be reliable.

NightOwl888 (changes requested):

src/Lucene.Net/Store/MMapDirectory.cs
paulirwin: If `FileStream.Length` is not reliable (I do not believe this to be the case) then we have bigger problems, because the only caller to this internal method passes in `fc.Length` as the `length` argument (see line 243). My hypothesis (since this bug is not reproducible) was that the file length on disk was growing between the call to this method and the call to create the memory mapped file as a race condition. I'm fine with reverting it and letting this fail in that case, it might have just been a one-off fluke. But note that if the reason for reverting it is that `FileStream.Length` is unreliable, then the calling code would be invalid as well (additionally, CreateFromFile would be bogus too, as no matter what we pass, it uses that same property to check the capacity argument). I do not think we should assume that it is unreliable and try to work around that unless you have a proposed solution. But I'm also fine with accepting this code as possibly rarely concurrency-unsafe.

NightOwl888 (changes requested):

src/Lucene.Net/Store/MMapDirectory.cs
paulirwin: Alright that has been reverted. This means this PR does not attempt to fix the bug in question, so I've updated the title and description to reflect the change to the adjust value alone. Given that a race condition is the only way I can explain the error, and that it has only happened once in tests, I guess we can just accept this for now. I think we should remove this issue from the 4.8.0 scope as well as a result. If someone runs into this in the wild, they'll probably find the issue and chime in with better reproduction steps. Or maybe it was the result of a cosmic ray bit flip and it'll never happen again in the wild...

NightOwl888: I reviewed #1090 and there are a few things that could be contributing to the failure. It is specifically testing the file locking functionality and in this case there was a failure on Linux.

1. Technically, Linux supports `FileStream.Lock()` which could be changed [here](https://github.com/apache/lucenenet/blob/54505e8964c112119b83adf8bf9163d687765cbe/src/Lucene.Net/Store/NativeFSLockFactory.cs#L100-L107), however after reading about it on dotnet/runtime and how it is possible to bypass the CLR to change it, we went with using the same approach we used on macOS, which is to use `SharingNativeFSLock`. We probably could have used `FileStream.Lock()` (with `NativeFSLock`) at the time, however, it seems that they have changed the native call in .NET 6 so it no longer gets an exclusive lock on Linux: https://github.com/dotnet/runtime/issues/64634.
2. On Windows, the HResult values of IOException are known. However, on all other platforms, these values depend on the underlying OS. We determine what the value is for the current OS by provoking the exception during static initialization and recording the value. There is a non-zero chance that this logic could fail to get a value and be set to `null`. When that happens, the implementation uses the `FallbackNativeFSLock`, which just does a best effort, but is known to have concurrency problems. Note there is some debug code to determine the HResult value that was found [here](https://github.com/apache/lucenenet/blob/54505e8964c112119b83adf8bf9163d687765cbe/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs#L3184-L3234).
3. We added a ["hole" in the locking implementation](https://github.com/apache/lucenenet/blob/54505e8964c112119b83adf8bf9163d687765cbe/src/Lucene.Net/Store/NativeFSLockFactory.cs#L496-L498) specifically so `RAMDirectory` can copy the lock file. This "hole" could cause the `SharingNativeFSLock` to fail. Ideally, the `share` setting would always be set to `Share.None` to ensure the lock functions properly in all cases. However, we would need to find some alternative for `RAMDirectory` to be able to copy the contents of the file while there is a lock held on it. This test failure could have occurred because the lock failed to throw its exception due to using `Share.Read` instead of `Share.None`. It might be worth a try to see if `RAMDirectory` can function without the lock file or with a fake lock file when copying the contents of the directory.

Detecting the HResult code by provoking the exception is not ideal. Also, due to the fact that MS changed the underlying native method for `FileStream.Lock()` in .NET 6, it doesn't sound like that is a viable alternative any longer on Linux. It may be that dropping to native code is the only way to properly solve it on Linux 100% of the time. Note that RavenDB changed the implementation to do all index file access using native code (including the memory mapped bit and file locking bit). I suspect that we could do the file locking in isolation without dropping to native code on everything, though.

NightOwl888: Reviewing this a bit more, the exception is almost certainly due to the `RAMDirectory` "hole". Allowing `Share.Read` access in `Obtain()` means the IOException that causes the lock to function will not occur. Without the exception, `IsLocked()` can never return `true`. Since the lock file is just a marker that indicates whether or not the directory is locked, using a fake one for `RAMDirectory` should be fine.

So, we would need to change

https://github.com/apache/lucenenet/blob/54505e8964c112119b83adf8bf9163d687765cbe/src/Lucene.Net/Store/RAMDirectory.cs#L104

so there is a special case to create a fake lock file instead of copy it. And remove the "hole" from 

https://github.com/apache/lucenenet/blob/54505e8964c112119b83adf8bf9163d687765cbe/src/Lucene.Net/Store/NativeFSLockFactory.cs#L496-L498

so it is always using `Share.None`.

------------------------

Alternatively, we could change the `SharingNativeFSLock` to attempt to write a byte in `IsLocked()`.

https://github.com/apache/lucenenet/blob/54505e8964c112119b83adf8bf9163d687765cbe/src/Lucene.Net/Store/NativeFSLockFactory.cs#L799-L802

This would cause an IOException when the file uses `Share.Read` since it would not be writable, but it would be a different exception than `NativeFSLockFactory.HRESULT_FILE_SHARE_VIOLATION`, so we would need to test for both cases (and depending on the exception, we may need to store another HResult value during initialization).

-----------------------

However, I still think that the best solution would be to change it to use native methods. This would eliminate exceptions thrown just to check true/false whether or not the directory is locked, since we could rely on native error codes instead. After all, it is supposed to be a `NativeFSLockFactory`...

paulirwin: Good finds! Can you point me to how RAMDirectory is related to this test, though? AFAICT it randomly gets one of the FS_DIRECTORIES instances which is SimpleFSDirectory, NIOFSDirectory, or MMapDirectory... not RAMDirectory. I'm not quite following how that is connected to this failure (although I understand how having Read access could mean the lock is never considered held).

NightOwl888 (approved):

PR #1139, opened 2025-03-09
Remove unused private members, #669

### LINKED ISSUE #669: Remove unused private/internal types and members
Mentioned in #648 

Specific issue: https://sonarcloud.io/project/issues?issues=AYPAuNd5hbfJOGLOoZ8u&open=AYPAuNd5hbfJOGLOoZ8u&id=nikcio_lucenenet (Solved)

This can be removed and replaced with a comment `// LUCENENET specific - removed FSTEntry because it is not in use`.

Similar issues found: https://sonarcloud.io/project/issues?resolved=false&rules=csharpsquid%3AS1144&id=apache_lucenenet

paulirwin:

Remove unused private members

Fixes #669

## Description

Removes many unused members found by SonarQube, except for those that have justifications.

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/El/GreekStemmer.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/El/GreekStemmer.cs
@@ -1003,8 +1003,8 @@ namespace Lucene.Net.Analysis.El
         /// <param name="len"> The length of the char[] array. </param>
         /// <param name="suffix"> A <see cref="string"/> object to check if the word given ends with these characters. </param>
         /// <returns> True if the word ends with the suffix given , false otherwise. </returns>
-        private static bool EndsWith(ReadOnlySpan<char> s, int len, string suffix) // LUCENENET: CA1822: Mark members as static
-            => s.Slice(0, len).EndsWith(suffix.AsSpan()); // LUCENENET specific - optimized for ReadOnlySpan<char>
+        private static bool EndsWith(char[] s, int len, string suffix) // LUCENENET: CA1822: Mark members as static
+            => s.AsSpan(0, len).EndsWith(suffix.AsSpan()); // LUCENENET specific - optimized for ReadOnlySpan<char>
 
         /// <summary>
         /// Checks if the word contained in the leading portion of <see cref="T:char[]"/> array ,
--- a/src/Lucene.Net.Analysis.Common/Analysis/El/GreekStemmer.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/El/GreekStemmer.cs
@@ -1,6 +1,7 @@
 // Lucene version compatibility level 4.8.1
 using Lucene.Net.Analysis.Util;
 using Lucene.Net.Util;
+using System;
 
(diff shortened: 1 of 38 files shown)


src/Lucene.Net.Analysis.Common/Analysis/El/GreekStemmer.cs
```diff
@@ -994,39 +994,40 @@ private static int Rule22(char[] s, int len) // LUCENENET: CA1822: Mark members
...
+        // LUCENENET: commented out unused private method
+        // /// <summary>
+        // /// Checks if the word contained in the leading portion of char[] array ,
+        // /// ends with the suffix given as parameter.
+        // /// </summary>
+        // /// <param name="s"> A char[] array that represents a word. </param>
+        // /// <param name="len"> The length of the char[] array. </param>
+        // /// <param name="suffix"> A <see cref="string"/> object to check if the word given ends with these characters. </param>
+        // /// <returns> True if the word ends with the suffix given , false otherwise. </returns>
+        // private static bool EndsWith(char[] s, int len, string suffix) // LUCENENET: CA1822: Mark members as static
```
NightOwl888, line 1006: In the upstream code, this method is being called instead of `StemmerUtil.EndsWith()`. Let's reenable this and have it cascade the call to `MemoryExtensions`, which is better optimized.

```c#
private static bool EndsWith(char[] s, int len, string suffix) // LUCENENET: CA1822: Mark members as static
    => MemoryExtensions.EndsWith(s.AsSpan(0, len), suffix.AsSpan());
```

We can probably also replace the `StemmerUtil` implementation with the same cascade. Although, it is probably more sensible for end users to call `MemoryExtensions` directly.

paulirwin: Changed this to have the method take `ReadOnlySpan<char>` and use `Slice` instead... let me know what you think.

NightOwl888: If we use `ReadOnlySpan<char>`, then the `int len` parameter is redundant, since the span should already be sliced to the length when passed in. The whole point is to eliminate any extra `start` and `length` parameters, since the one method will do it all just by slicing. But IMO, it makes more sense to keep passing the `char[]` here (for now) along with `int` and `string` to match Lucene.

> Note that we have a hybrid implementation of Snowball. We copied most of the bits from Lucene, but Snowball has a C# implementation that would be better to use. https://github.com/snowballstem/snowball/tree/master/csharp. The stemmer code is generated somehow, but I haven't looked at it in depth. We ended up using some of it because the Java code was using Reflection. But my point is there is a better optimized implementation that we can use in the future.

paulirwin: It's not redundant in this case, because it is not sliced when passed in (and there are many, many calls to this that would have to be updated to do that), but I'll change it to `char[]` even though I disagree with the reasoning.

NightOwl888 (approved):

PR #1147, opened 2025-03-18
BREAKING: De-nest AttributeSource.AttributeFactory, #1145

### LINKED ISSUE #1145: De-nest AttributeFactory from AttributeSource

### Task description

The `AttributeFactory` class should not be nested within `AttributeSource`, since it is designed for public use. Instead, it should be moved up into the `Lucene.Net.Util` namespace.

Once it is moved the `AttributeSource.GetAttributeFactory()` method can be made into an `AttributeFactory` proprty.

paulirwin:

De-nest AttributeSource.AttributeFactory to be a top-level type.

Fixes #1145

## Description

This PR de-nests AttributeFactory so it is a top-level type in Lucene.Net.Util. Because it is public, this is a breaking change. 

This also changes `AttributeSource.GetAttributeFactory()` into the `AttributeSource.AttributeFactory` property, now that there is no longer a conflict with the nested type name.


NightOwl888 (approved):
Thanks for the PR. Looks good.

PR #1148, opened 2025-03-18
Make Lucene.Net.Store.Lock reusable, #1073

### LINKED ISSUE #1073: Make Lucene.Net.Store.Lock reusable

### Task description

From investigating #265:

> I have also been made aware that there is at least one class (Lucene.Net.Store.Lock, if I recall correctly) that is designed to be re-opened after it is closed.

In investigating the Lock class, I found that in Lucene 5.3.0, [the Lock API was simplified](https://github.com/apache/lucene/commit/fe6c3dc9397b0397aebf84fedb57875beb163d1d#diff-cd03a8db5d3607ff62811dcb2f8a8b6cae90402664beda95a702132d99992c18) to remove the ability to re-obtain a lock using the same instance, and instead it only provides a method to close (dispose) the lock once you're done with it. In that version of Lucene, it seems like IDisposable is the correct choice. I wish we could mark all these other methods as Obsolete, but unfortunately Obtain et al. are how it is intended to be used today, so we should not be throwing warnings upon use, even though the methods will be removed in a future version of Lucene.NET.

In 4.8, it certainly does seem like this type is designed to be reusable, as you could theoretically obtain the lock when you need it, `close()` it to release it, and re-obtain it to do something else. This indicates that we should update Lock to use the new `ICloseable` interface once #271 is merged. To encourage best practices and compatibility with future Lucene.NET, we should leave this as IDisposable where Dispose calls Close, but provide this additional interface to allow for reuse. To avoid this being a breaking change, we should not throw if Dispose is called more than once, nor make it a no-op in that case. It can simply delegate the call to Close. 

If possible, we should create an analyzer that looks for calls to any Lock methods after Dispose is called, and warn against such use.

paulirwin:

Make Lucene.Net.Store.Lock reusable via ICloseable

Fixes #1073

## Description

See rationale in #1073. It is theoretically possible to reuse a Lock instance in Lucene 4.8.1 by calling `obtain()` after a call to `close()`, and some Lucene.NET code does this implicitly (such as in `FallbackNativeFSLock.IsLocked()`, which attempts to obtain and then release the lock). Therefore, the IDisposable pattern doesn't really always work, as you're not supposed to reuse an object after disposing of it.

This PR adds the ICloseable interface to Lock, and makes its Close method the primary extension point for Lock subclasses to implement their release logic, better matching Lucene. Dispose has been changed into a virtual method to allow for custom Lock subclasses to dispose of resources in a non-reusable way if needed.

This will be changing in a future version of Lucene.NET post-5.3.0, where the locks are no longer reusable, so at that time we will remove the ICloseable interface and just use IDisposable.

Because of the change to the Lock base class, this is a breaking change for any custom Lock implementations.

### DIFF
--- a/src/Lucene.Net.TestFramework/Store/MockLockFactoryWrapper.cs
+++ b/src/Lucene.Net.TestFramework/Store/MockLockFactoryWrapper.cs
@@ -81,13 +81,10 @@ namespace Lucene.Net.Store
                 }
             }
 
-            protected override void Dispose(bool disposing)
+            public override void Close()
             {
-                if (disposing)
-                {
-                    delegateLock.Dispose();
-                    outerInstance.dir.openLocks.Remove(name);
-                }
+                delegateLock.Close();
+                outerInstance.dir.openLocks.Remove(name);
             }
 
             public override bool IsLocked()
--- a/src/Lucene.Net.Tests/Store/TestLock.cs
+++ b/src/Lucene.Net.Tests/Store/TestLock.cs
@@ -53,7 +53,7 @@ namespace Lucene.Net.Store
                 return false;
             }
 
-            protected override void Dispose(bool disposing)
+            public override void Close()
             {
                 // do nothing
             }
--- a/src/Lucene.Net.Tests/Store/TestLockFactory.cs
+++ b/src/Lucene.Net.Tests/Store/TestLockFactory.cs
@@ -463,7 +463,7 @@ namespace Lucene.Net.Store
                     return true;
                 }
 
-                protected override void Dispose(bool disposing)
+                public override void Close()
                 {
                     // do nothing
                 }
--- a/src/Lucene.Net/Store/Lock.cs
+++ b/src/Lucene.Net/Store/Lock.cs
@@ -1,4 +1,4 @@
-﻿using Lucene.Net.Support.Threading;
+﻿using Lucene.Net.Util;
 using System;
 using System.IO;
 using System.Threading;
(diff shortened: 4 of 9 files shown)


paulirwin:
@NightOwl888 After thinking about this more, I'm having doubts about whether we want to do this at all. While Dispose is _semantically_ not the right way to release a lock in a way that is reusable, there's nothing in our code that prevents that today without this PR. The only place in our code that seems to reuse a lock is what I mentioned above in FallbackNativeFSLock, and that has been fine for many releases. Most typical uses of locks I've seen dispose/close them after obtaining them once and aren't reused. And, that capability is still there today with Dispose and then call Obtain again if you need it. The fact that possible reuse is going away in 5.3.0+ seems to indicate further that it might not be a good idea to introduce this change, and instead encourage people to treat locks as single-use only (apart from our internal reuse in FallbackNativeFSLock).

The only possible downside I could see of _not_ doing this PR is if someone has a custom Lock implementation (which I think is probably fairly limited to begin with) _and_ that implementation has unmanaged resources that must be cleaned up in Dispose _and_ they can't make that implementation reusable on a call to Obtain after Dispose _and_ their users of their Lock subclass expect it to be reusable _and_ despite all of that, they have a plan for Lucene 5.3.0+ compat where it's not reusable and are okay with that future breaking change... but it's probably a pretty remote chance that all of those things are true. And we haven't heard from anyone (that I know of) that has needed this in the prior betas.

While an interesting exercise, my vote would be to abandon this PR and close the issue.

NightOwl888:
Thanks for looking into this. I knew there was something in Lucene.Net.Store that was not using `Dispose()` correctly. But, I agree with your points about not changing this. Sometimes it makes sense to use (abuse?) `Dispose()` just so the compiler will give us a free way to put it into a `using` block, even if the implementation is meant to be reused.

Also, since you mentioned that this eventually *will* be non-reusable in the future, it would probably be less intrusive for users to have to change to `Close()` only to change again to `Dispose()` later.

In addition, I agree that creating a custom reusable `Lock` with disposable components is probably an extreme edge case.

`FallbackNativeFSLock` also only loads when the HResult for the lock sharing exception cannot be determined on the current OS, and AFAIK, that never happens in practice. There is a non-zero chance that it *could* happen, and `FallbackNativeFSLock` is there to fill the gap. However, I suspect we can completely eliminate that implementation and approach if `NativeFSLockFactory` were converted to actually use native method calls instead of relying on exceptions (which are raw `IOExceptions`, making them ambiguous except for the HResult value).

PR #1152, opened 2025-03-25
Fix MMapDirectory performance issue, #1151

### LINKED ISSUE #1151: Search performance issue with MMapDirectory under load

### Describe the bug

I have discovered that MMapDirectory has serious performance problems under parallel load when searching (with no background writes), across all platforms. The demo repo to reproduce this issue is at https://github.com/paulirwin/LuceneAdventureWorks/

This demo builds a simple index from the AdventureWorks2022 database data, and does a load test against it with 100k requests, using Parallel.For to search in parallel as fast as possible.

Of note: MMapDirectory is the default Directory implementation returned by `FSDirectory.Open` on Windows and Linux 64-bit platforms.

| OS | Architecture | Runtime | SimpleFS | NIOFS | MMap | Compared to SimpleFS |
| --- | --- | --- | --- | --- | --- | --- |
| macOS 15 | arm64 | .NET 9.0 | 4.07s | 5.46s | 3m25s | 51x slower |
| Ubuntu 24.04 | x64 | .NET 9.0 | 2.21s | 2.43s | 35.57s | 16x slower |
| Windows 11 | x64 | .NET 9.0 | 4.13s | 11.66s | 2m33s | 37x slower |
| Windows 11 | x64 | .NET Framework 4.8.1 | 7.83s | 2m21s^ | 2m21s | 18x slower |

^ This was an outlier for NIOFS, and yes it was identical to MMap. Also note that hardware between these machines is not exactly comparable but all are 6-8 performance core machines with at either 32GB or 64GB RAM.

Given that MMapDirectory is the default Directory implementation returned by FSDirectory on Windows and Linux 64-bit (which is the majority of Windows and Linux use nowadays), we definitely need to resolve this, as MMapDirectory is supposed to be a fast implementation.

### Expected Behavior

MMapDirectory performs well on all platforms.

### Steps To Reproduce

See demo repo.

### Exceptions (if any)

_No response_

### Lucene.NET Version

4.8.0-beta00017

### .NET Version

_No response_

### Operating System

_No response_

### Anything else?

_No response_

paulirwin:

Fix MMapDirectory performance issue by using BufferedIndexInput instead of ByteBufferIndexInput.

Fixes #1151

## Description

This is a draft PR to run tests on all platforms and gather feedback on this approach. 

As noted in #1151, MMapDirectory has a severe performance problem that causes it to be as much as 51x slower in my testing than SimpleFSDirectory. I was able to narrow down the root cause: reading a single byte from a MemoryMappedViewAccessor is very slow in .NET. I believe this is because it has to do logic like acquire and release a pointer, in addition to the range checking. Reading a single byte at a time is very common in our codebase, such as in LZ4 decompression which was a hotspot in the profile. The performance difference of memory-mapped files between .NET and Java disappears when you read multiple bytes at once via ReadArray.

While it might be possible to create a ByteBuffer implementation for memory-mapped files that also has an internal buffer to read 1kB at a time, I decided to take a stab at making MMapDirectory use a similar approach to SimpleFSDirectory, and have its IndexInput implementation inherit from BufferedIndexInput, which maintains a 1kB buffer around the input. Then, when it needs to refill its buffer, it reads from the memory-mapped file view accessor. By buffering 1kB at a time, ReadByte reads from the already-filled buffer which is a fast array operation, resulting in a massive speed-up.

Initial performance results (compare to results in #1151), macOS arm64, .NET 9:
- SimpleFS: 3.96s
- NIOFS: 6.01s
- MMap: 2.47s (0.62x SimpleFS)

### DIFF
--- a/src/Lucene.Net/Store/MMapDirectory.cs
+++ b/src/Lucene.Net/Store/MMapDirectory.cs
@@ -244,7 +244,7 @@ namespace Lucene.Net.Store
 
         public sealed class MMapIndexInput : BufferedIndexInput
         {
-            internal MemoryMappedFile memoryMappedFile; // .NET port: this is equivalent to FileChannel.map
+            internal readonly MemoryMappedFile memoryMappedFile; // .NET port: this is equivalent to FileChannel.map
             private readonly MemoryMappedViewAccessor accessor;
             internal readonly FileStream fc;
             private int disposed = 0; // LUCENENET specific - allow double-dispose
@@ -315,10 +315,14 @@ namespace Lucene.Net.Store
                     {
                         try
                         {
+                            if (this.accessor != null)
+                            {
+                                this.accessor.Dispose();
+                            }
+
                             if (this.memoryMappedFile != null)
                             {
                                 this.memoryMappedFile.Dispose();
-                                this.memoryMappedFile = null;
                             }
                         }
                         finally
--- a/src/Lucene.Net/Store/MMapDirectory.cs
+++ b/src/Lucene.Net/Store/MMapDirectory.cs
@@ -185,38 +185,50 @@ namespace Lucene.Net.Store
             EnsureOpen();
             var file = new FileInfo(Path.Combine(Directory.FullName, name));
             var fc = new FileStream(file.FullName, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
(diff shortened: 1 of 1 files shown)


paulirwin:
I am abandoning this PR in favor of #1267.

PR #1170, opened 2025-08-26
Add async PerformAsync and FlushAsync methods with CancellationToken support

### LINKED ISSUE #1169: Add asynchronous PerformAsync and FlushAsync methods to IReplicationService and IReplicationResponse

Currently, IReplicationService.Perform and IReplicationResponse.Flush are synchronous. In modern .NET applications, blocking synchronous calls can reduce scalability and responsiveness, especially for I/O-bound replication operations.

I am trying to implement a background replication server using Kestrel and ASP.NET Core. Without asynchronous support, calls to Perform block the thread, limiting concurrency and scalability. Moreover, synchronous usage in ASP.NET Core requires AllowSynchronousIO = true on Kestrel, which is discouraged and may be disabled in future framework versions.

Introduce asynchronous methods:

->Task IReplicationService.PerformAsync(IReplicationRequest request, IReplicationResponse response, CancellationToken cancellationToken = default)

->Task IReplicationResponse.FlushAsync(CancellationToken cancellationToken = default)

These methods should use the existing I/O streams asynchronously (WriteAsync, CopyToAsync, etc.) and accept an optional CancellationToken.

### Additional context

This change allows Lucene.NET replication to be used efficiently in modern .NET applications (ASP.NET Core, background services) without blocking threads and without needing AllowSynchronousIO = true. Existing synchronous methods would remain for backward compatibility.

NehanPathan:
---
* @paulirwin for review.

**Linked Issue:**
Fixes #1169

**Description:**
This PR implements asynchronous support for the replication framework in Lucene.NET, based on suggestions from @paulirwin. The changes include:

1. **`IReplicationService` interface**

   * Added `Task PerformAsync(IReplicationRequest request, IReplicationResponse response, CancellationToken cancellationToken = default)`
   * Supports cancellation and async execution of replication tasks.

2. **`IReplicationResponse` interface**

   * Added `Task FlushAsync(CancellationToken cancellationToken = default)`
   * Provides asynchronous flush support to the underlying response stream.

3. **`ReplicationService` class**

   * Added `PerformAsync` method mirroring `Perform`, using `await` for I/O operations (`Stream.CopyToAsync`, `WriteAsync`, `FlushAsync`).
   * Accepts optional `CancellationToken` for cooperative cancellation.

4. **`SessionToken` class**

   * Added `Task SerializeAsync(Stream output, CancellationToken cancellationToken = default)` to allow asynchronous serialization.

5. **`AspNetCoreReplicationResponse` class**

   * Added `FlushAsync` method that calls `HttpResponse.Body.FlushAsync` with optional `CancellationToken`.

**Motivation:**

* Modern .NET applications require async I/O for scalability and responsiveness.
* Existing synchronous methods require `AllowSynchronousIO=true` in Kestrel, which is not recommended for production.
* This PR ensures the replication API follows .NET async best practices without breaking existing synchronous behavior.

**Additional Context:**

* The synchronous APIs (`Perform` and `Flush`) remain unchanged for backward compatibility.
* This PR improves compatibility for ASP.NET Core hosting scenarios where synchronous I/O is restricted.

---

### DIFF
--- a/src/Lucene.Net.Replicator/Http/ReplicationService.cs
+++ b/src/Lucene.Net.Replicator/Http/ReplicationService.cs
@@ -139,6 +139,7 @@ namespace Lucene.Net.Replicator.Http
             if (!replicators.TryGetValue(pathElements[SHARD_IDX], out IReplicator replicator))
                 throw ServletException.Create("unrecognized shard ID " + pathElements[SHARD_IDX]);
 
+            // SOLR-8933 Don't close this stream.
             try
             {
                 switch (action)
@@ -175,7 +176,7 @@ namespace Lucene.Net.Replicator.Http
             }
             catch (Exception)
             {
-                response.StatusCode = (int)HttpStatusCode.InternalServerError;
+                response.StatusCode = (int)HttpStatusCode.InternalServerError; // propagate the failure
             }
             finally
             {
@@ -208,8 +209,8 @@ namespace Lucene.Net.Replicator.Http
                     }
                     return Task.CompletedTask;
                 },
-                () => { response.Flush(); return Task.CompletedTask; }
-            ).GetAwaiter().GetResult(); // // keep sync behavior
+                () => { response.Body.Flush(); return Task.CompletedTask; }
+            ).ConfigureAwait(false).GetAwaiter().GetResult(); // keep sync behavior
         }
 
 
@@ -241,7 +242,7 @@ namespace Lucene.Net.Replicator.Http
                         await token.SerializeAsync(response.Body, cancellationToken);
                     }
                 },
-                () => response.FlushAsync(cancellationToken)
+                () => response.Body.FlushAsync(cancellationToken)
             );
         }
     }
--- a/src/Lucene.Net.Replicator/Support/Http/Abstractions/IReplicationResponse.cs
+++ b/src/Lucene.Net.Replicator/Support/Http/Abstractions/IReplicationResponse.cs
@@ -1,6 +1,4 @@
 using System.IO;
-using System.Threading;
-using System.Threading.Tasks;
 
 namespace Lucene.Net.Replicator.Http.Abstractions
 {
@@ -39,17 +37,5 @@ namespace Lucene.Net.Replicator.Http.Abstractions
         /// The response content.
         /// </summary>
         Stream Body { get; }
-
-        /// <summary>
-        /// Flushes the reponse to the underlying response stream.
-        /// </summary>
-        void Flush();
-
-        /// <summary>
-        /// Flushes the response to the underlying response stream asynchronously.
-        /// </summary>
-        /// <param name="cancellationToken">Optional cancellation token.</param>
-        /// <returns>A task representing the asynchronous operation.</returns>
-        Task FlushAsync(CancellationToken cancellationToken = default);
     }
 }
--- a/src/Lucene.Net.Tests.Replicator/Http/ReplicationServlet.cs
+++ b/src/Lucene.Net.Tests.Replicator/Http/ReplicationServlet.cs
@@ -2,9 +2,9 @@
 using Lucene.Net.Replicator.AspNetCore;
 using Lucene.Net.Replicator.Http.Abstractions;
 using Microsoft.AspNetCore.Builder;
-using Microsoft.AspNetCore.Routing;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
 using System;
 using System.Threading.Tasks;
 
--- a/src/dotnet/Lucene.Net.Replicator.AspNetCore/AspNetCoreReplicationResponse.cs
+++ b/src/dotnet/Lucene.Net.Replicator.AspNetCore/AspNetCoreReplicationResponse.cs
@@ -1,8 +1,6 @@
 using Lucene.Net.Replicator.Http.Abstractions;
 using Microsoft.AspNetCore.Http;
 using System.IO;
-using System.Threading;
-using System.Threading.Tasks;
 
 namespace Lucene.Net.Replicator.AspNetCore
 {
@@ -59,29 +57,5 @@ namespace Lucene.Net.Replicator.AspNetCore
         /// This simply returns the <see cref="HttpResponse.Body"/>.
         /// </remarks>
         public Stream Body => response.Body;
-
-        /// <summary>
-        /// Flushes the reponse to the underlying response stream.
-        /// </summary>
-        /// <remarks>
-        /// This simply calls <see cref="Stream.Flush"/> on the <see cref="HttpResponse.Body"/>.
-        /// </remarks>
-        public void Flush()
-        {
-            response.Body.Flush();
-        }
-
-        /// <summary>
-        /// Flushes the response to the underlying response stream asynchronously.
-        /// </summary>
-        /// <param name="cancellationToken">Optional cancellation token.</param>
-        /// <returns>A task representing the asynchronous operation.</returns>
-        /// <remarks>
-        /// This simply calls <see cref="Stream.FlushAsync(CancellationToken)"/> on the <see cref="HttpResponse.Body"/>.
-        /// </remarks>
-        public async Task FlushAsync(CancellationToken cancellationToken = default)
-        {
-            await response.Body.FlushAsync(cancellationToken);
-        }
     }
 }
--- a/src/Lucene.Net.Tests.Replicator/Support/Net/HttpListenerReplicationResponse.cs
+++ b/src/Lucene.Net.Tests.Replicator/Support/Net/HttpListenerReplicationResponse.cs
@@ -40,10 +40,5 @@ namespace Lucene.Net.Replicator.Net
         }
 
         public Stream Body => _response.OutputStream;
-
-        public void Flush() => _response.OutputStream.Flush();
-
-        public Task FlushAsync(CancellationToken cancellationToken = default) =>
-            _response.OutputStream.FlushAsync(cancellationToken);
     }
 }
--- a/src/Lucene.Net.Tests.Replicator/Support/Net/TestServer.cs
+++ b/src/Lucene.Net.Tests.Replicator/Support/Net/TestServer.cs
@@ -107,7 +107,7 @@ namespace Lucene.Net.Replicator.Net
                         }
                         finally
                         {
-                            await response.FlushAsync(token);
+                            await response.Body.FlushAsync(token);
                             context.Response.Close();
                         }
                     }, token);


src/dotnet/Lucene.Net.Replicator.AspNetCore/AspNetCoreReplicationResponse.cs
```diff
@@ -68,5 +70,21 @@ public void Flush()
...
+        /// Flushes the response to the underlying response stream asynchronously.
+        /// </summary>
+        /// <param name="cancellationToken">Optional cancellation token.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        /// <remarks>
+        /// This simply calls <see cref="Stream.FlushAsync(CancellationToken)"/> on the <see cref="HttpResponse.Body"/>.
+        /// </remarks>
+        public async Task FlushAsync(CancellationToken cancellationToken = default)
+        {
+            if (response.Body.CanWrite)
```
paulirwin, line 84: This check does not exist in Flush above, I wonder if this is needed?

src/Lucene.Net.Replicator/SessionToken.cs
```diff
@@ -112,6 +114,33 @@ public void Serialize(DataOutputStream writer)
...
+        /// <summary>
+        /// Asynchronously serialize the token data for communication between server and client.
+        /// </summary>
+        /// <param name="output">The <see cref="Stream"/> to write the token data to.</param>
+        /// <param name="cancellationToken">A cancellation token to observe while waiting for the flush to complete.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        public async Task SerializeAsync(Stream output, CancellationToken cancellationToken = default)
+        {
+            using var writer = new DataOutputStream(output);
+            writer.WriteUTF(Id);
```
paulirwin, line 126: I think we might need async overloads of these `WriteUTF` and `WriteInt32` methods too (and any similar others on that type, might as well). Otherwise, these calls might throw if synchronous IO is not enabled, right?

src/Lucene.Net.Replicator/Http/ReplicationService.cs
```diff
@@ -188,5 +191,70 @@ public virtual void Perform(IReplicationRequest request, IReplicationResponse re
...
         }
+
+        /// <summary>
+        /// Executes the replication task asynchronously.
+        /// </summary>
+        /// <param name="request">The replication request containing action and parameters.</param>
+        /// <param name="response">The replication response used to send data back to the client.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> to observe while performing the replication.</param>
+        /// <exception cref="InvalidOperationException">Thrown when required parameters are missing or invalid.</exception>
+        public virtual async Task PerformAsync(IReplicationRequest request, IReplicationResponse response, CancellationToken cancellationToken = default)
```
paulirwin, line 202: This is largely duplicated code from above. I wonder if there's a way to reuse some code between the methods?

paulirwin (changes requested):
@NehanPathan Thanks for the PR! This mostly looks good. A few comments addedd.

src/Lucene.Net.Replicator/SessionToken.cs
NightOwl888: FYI - `WriteUTF` is a proprietary UTF format from Java. It exists on the `DataOutputStream` only, with the complementary operation on `DataInputStream`. Both of those classes are already synchronized with a lock, so it would require more than simply slapping another method onto the class and making it async. ChatGPT recommends replacing the lock with `SemaphoreSlim` as the preferred synchronization option.

https://chatgpt.com/share/68adb578-c44c-8005-93f0-c1dc123e909c

paulirwin: @NightOwl888 Thanks for the insight. That further reinforces my (second thought) suggestion about leaving it as-is for now if possible. @NehanPathan Let us know if that works as-is without making those methods async.

paulirwin: I forgot that DataOutputStream is part of J2N. That's unfortunate, as we'd need to cascade yet another PR and release to fix this. I'd like to get this fixed sooner than a J2N release so we can wrap up this GSoC work without having this async problem hanging over us. Since we only need WriteUTF and WriteInt32, I wonder if we can temporarily copy the logic into Lucene.NET from J2N as private methods in this type with a comment that they should be replaced with J2N methods later, and remove the use of DataOutputStream entirely. The locking provided by DataOutputStream is not needed since the type is only used within this method as a local variable.

paulirwin: A prerelease would be fine but I don't want this to get mired down in dealing with nuances or risk of the locking change which is unnecessary for this feature.

paulirwin: To help prime this if we decide to copy the methods, I had Claude Code generate a copy of those methods as private static methods that operate on a stream parameter instead of a field, and make them async with cancellation token support. I have not tested this code, it needs to be validated. 

```c#
private static async Task WriteUTFAsync(Stream output, string value, CancellationToken cancellationToken = default)
  {
      if (value is null)
          throw new ArgumentNullException(nameof(value));

      long utfCount = CountUTFBytes(value);
      if (utfCount > 65535)
      {
          throw new FormatException(SR.Format_InvalidUTFTooLong);
      }
      byte[] buffer = new byte[(int)utfCount + 2];
      int offset = 0;
      offset = WriteInt16ToBuffer((int)utfCount, buffer, offset);
      offset = WriteUTFBytesToBuffer(value, (int)utfCount, buffer, offset);
      await output.WriteAsync(buffer, 0, offset, cancellationToken).ConfigureAwait(false);
  }

  private static async Task WriteInt32Async(Stream output, int value, CancellationToken cancellationToken = default)
  {
      byte[] buff = new byte[4];
      buff[0] = (byte)(value >> 24);
      buff[1] = (byte)(value >> 16);
      buff[2] = (byte)(value >> 8);
      buff[3] = (byte)value;
      await output.WriteAsync(buff, 0, 4, cancellationToken).ConfigureAwait(false);
  }

  private static async Task WriteInt64Async(Stream output, long value, CancellationToken cancellationToken = default)
  {
      byte[] buff = new byte[8];
      buff[0] = (byte)(value >> 56);
      buff[1] = (byte)(value >> 48);
      buff[2] = (byte)(value >> 40);
      buff[3] = (byte)(value >> 32);
      buff[4] = (byte)(value >> 24);
      buff[5] = (byte)(value >> 16);
      buff[6] = (byte)(value >> 8);
      buff[7] = (byte)value;
      await output.WriteAsync(buff, 0, 8, cancellationToken).ConfigureAwait(false);
  }

  // Helper methods remain synchronous as they don't perform I/O
  private static long CountUTFBytes(string value)
  {
      int utfCount = 0, length = value.Length;
      for (int i = 0; i < length; i++)
      {
          int charValue = value[i];
          if (charValue > 0 && charValue <= 127)
          {
              utfCount++;
          }
          else if (charValue <= 2047)
          {
              utfCount += 2;
          }
          else
          {
              utfCount += 3;
          }
      }
      return utfCount;
  }

  private static int WriteInt16ToBuffer(int value, byte[] buffer, int offset)
  {
      buffer[offset++] = (byte)(value >> 8);
      buffer[offset++] = (byte)value;
      return offset;
  }

  private static int WriteUTFBytesToBuffer(string value, long count, byte[] buffer, int offset)
  {
      int length = value.Length;
      for (int i = 0; i < length; i++)
      {
          int charValue = value[i];
          if (charValue > 0 && charValue <= 127)
          {
              buffer[offset++] = (byte)charValue;
          }
          else if (charValue <= 2047)
          {
              buffer[offset++] = (byte)(0xc0 | (0x1f & (charValue >> 6)));
              buffer[offset++] = (byte)(0x80 | (0x3f & charValue));
          }
          else
          {
              buffer[offset++] = (byte)(0xe0 | (0x0f & (charValue >> 12)));
              buffer[offset++] = (byte)(0x80 | (0x3f & (charValue >> 6)));
              buffer[offset++] = (byte)(0x80 | (0x3f & charValue));
          }
      }
      return offset;
  }
  ```

Comment (NightOwl888) [reply src/Lucene.Net.Replicator/SessionToken.cs:126]: Gotcha.

Maybe it would make more sense to make `ReadUTF` and `WriteUTF` into extension methods of `BinaryReader`/`BinaryWriter` or even `Stream`. I think there are one or two other methods that didn't exist in the BCL that we would need. Other than that, the error handling might be slightly different than in the BCL, but that may not matter very much.

Once you have the extension methods battle tested, we can move them into J2N and integrate them into `DataInputStream` and `DataOutputStream`. So, it would be sensible to make low-level methods that accept `ReadOnlySpan<T>` and `Span<T>` to do the actual UTF conversions that can then be utilized to write to the various types of streams. The exceptions should also be changed from `FormatException` to `DecoderFallbackException` and `EncoderFallbackException` to match the BCL, which also means that catch blocks will need to be re-evaluated.

Comment (paulirwin) [reply src/Lucene.Net.Replicator/SessionToken.cs:126]: @NightOwl888 Sounds good!

@NehanPathan Please proceed with what Shad suggested above, apart from the J2N bits which will come later.

Comment (NightOwl888) [reply src/Lucene.Net.Replicator/SessionToken.cs:126]: > ## @NightOwl888
> I wanted to update you regarding the low-level UTF and integer stream methods. I initially tried using `Span<byte>` / `ReadOnlySpan<byte>` to reduce allocations and align with BCL-style patterns.
> 
> However, this causes compiler issues in the current project setup. The async `ReadAsync`/`WriteAsync` overloads with `Span<byte>` / `ReadOnlySpan<byte>` cannot be used, producing errors like:
> 
> * `No overload for method 'WriteAsync' takes 2 arguments`
> * `Cannot convert from 'Span<byte>' to 'Memory<byte>'`
> * `Feature 'ref and unsafe in async and iterator methods' is not available in C# 11.0`
> 
> Because of this, I propose we **use `byte[]` for all Read/Write methods**, including the UTF methods. This approach is consistent with the existing synchronous Read/Write helpers. The **byte[] approach works perfectly**, all tests have been run, and everything functions as expected. We can still include the updated UTF exception handling (`EncoderFallbackException` / `DecoderFallbackException`) as per BCL conventions.
> 
> This ensures compatibility with the current codebase without causing any compiler issues.
> 
> Any suggestion to solve issue or thoughts on this byte approach??, and I can finalize the changes accordingly.

Understood. This overload of `WriteAsync` is not available prior to .NET Core. The actual method isn't very complicated:

https://github.com/dotnet/runtime/blob/1d1bf92fcf43aa6981804dc53c5174445069c9e4/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs#L740-L762

However, it is clear after looking at the implementation why it accepts `ReadOnlyMemory<byte>` instead of `ReadOnlySpan<byte>` - because the latter has no way to convert to an array even if it is on the heap.

So, don't worry about using spans for the time being. It seems that Microsoft hasn't fully implemented support for them in `Stream`. Go ahead and use one of the `byte[]` overloads.

**Comment by paulirwin** [comment]
Alternatively, we could make the serialize methods internal and go ahead and accept a breaking change that they're not supposed to be part of our public API? It seems to me like this type is intended to be somewhat opaque anyways, and arguably it's a SRP violation to have serialization inside of it. Just a thought.

File: src/Lucene.Net/Support/IO/StreamExtensions.cs [hunk 4]
```diff
@@ -210,5 +213,154 @@ public static long ReadInt64(this Stream stream)
...
+            byte[] buff = new byte[4];
+            buff[0] = (byte)(value >> 24);
+            buff[1] = (byte)(value >> 16);
+            buff[2] = (byte)(value >> 8);
+            buff[3] = (byte)value;
+
+            await output.WriteAsync(buff, 0, buff.Length, cancellationToken).ConfigureAwait(false);
+        }
+
+        public static async Task WriteInt64Async(this Stream output, long value, CancellationToken cancellationToken = default)
```
NightOwl888, line 232: Please change the name to `WriteInt64BigEndianAsync()` to distinguish it against standard little-endian writes.

src/Lucene.Net/Support/IO/StreamExtensions.cs
```diff
@@ -210,5 +213,154 @@ public static long ReadInt64(this Stream stream)
...
+
+            byte[] buffer = new byte[(int)utfCount + 2];
+            int offset = 0;
+            offset = WriteInt16ToBuffer((int)utfCount, buffer, offset);
+            offset = WriteUTFBytesToBuffer(value, (int)utfCount, buffer, offset);
+
+            await output.WriteAsync(buffer, 0, offset, cancellationToken).ConfigureAwait(false);
+        }
+
+        public static async Task<int> ReadInt32Async(this Stream input, CancellationToken cancellationToken = default)
```
NightOwl888, line 269: Please change the name to `ReadInt32BigEndianAsync()` to distinguish it against standard little-endian reads.

src/Lucene.Net/Support/IO/StreamExtensions.cs
```diff
@@ -210,5 +213,154 @@ public static long ReadInt64(this Stream stream)
...
+                throw new ArgumentNullException(nameof(input));
+
+            byte[] buff = new byte[4];
+            int read = await input.ReadAsync(buff, 0, buff.Length, cancellationToken).ConfigureAwait(false);
+            if (read < buff.Length) throw new EndOfStreamException();
+
+            return (buff[0] << 24) | (buff[1] << 16) | (buff[2] << 8) | buff[3];
+        }
+
+        public static async Task<long> ReadInt64Async(this Stream input, CancellationToken cancellationToken = default)
```
NightOwl888, line 281: Please change the name to `ReadInt64BigEndianAsync()` to distinguish it against standard little-endian reads.

src/Lucene.Net/Support/IO/StreamExtensions.cs
```diff
@@ -210,5 +213,154 @@ public static long ReadInt64(this Stream stream)
...
+                    utfCount++;
+                else if (ch <= 2047)
+                    utfCount += 2;
+                else
+                    utfCount += 3;
+            }
+            return utfCount;
+        }
+
+        private static int WriteInt16ToBuffer(int value, byte[] buffer, int offset)
```
NightOwl888, line 336: Please change the name to `WriteInt16BigEndianToBuffer()` to distinguish it against standard little-endian writes.

src/Lucene.Net/Support/IO/StreamExtensions.cs
```diff
@@ -210,5 +213,154 @@ public static long ReadInt64(this Stream stream)
...
+        {
+            if (input is null)
+                throw new ArgumentNullException(nameof(input));
+
+            byte[] lenBuff = new byte[2];
+            int readLen = await input.ReadAsync(lenBuff, 0, lenBuff.Length, cancellationToken).ConfigureAwait(false);
+            if (readLen < lenBuff.Length) throw new EndOfStreamException();
+
+            int length = (lenBuff[0] << 8) | lenBuff[1];
+            byte[] buffer = new byte[length];
```
NightOwl888, line 310: We can use the `ArrayPool` here, since this array may be any size up to 64kb.

```c#
            byte[] buffer = ArrayPool<byte>.Shared.Rent(length);
            try
            {
                int read = await input.ReadAsync(buffer, 0, length, cancellationToken).ConfigureAwait(false);
                if (read < length)
                    throw new EndOfStreamException("Unexpected end of stream while reading UTF string.");

                return Encoding.UTF8.GetString(buffer, 0, length);
            }
            finally
            {
                // Always return, even on exception
                ArrayPool<byte>.Shared.Return(buffer);
            }
```

NightOwl888 (changes requested):

NightOwl888:
> Alternatively, we could make the serialize methods internal and go ahead and accept a breaking change that they're not supposed to be part of our public API? It seems to me like this type is intended to be somewhat opaque anyways, and arguably it's a SRP violation to have serialization inside of it. Just a thought.

Good point. Since `SerializeAsync()` is a brand new method, we can mark it `internal` until we have it finalized. `PerformAsync()` is the public method that gets called by the hosted network service, so it isn't likely something that `Lucene.Net.Extensions` will need. And if so, there is always `InternalsVisibleTo`.

src/Lucene.Net.Replicator/Http/ReplicationService.cs
```diff
@@ -118,75 +121,128 @@ private static string ExtractRequestParam(IReplicationRequest request, string pa
...
+        /// <exception cref="InvalidOperationException">Thrown when required parameters are missing or invalid.</exception>
+        public virtual Task PerformAsync(
+            IReplicationRequest request,
+            IReplicationResponse response,
+            CancellationToken cancellationToken = default)
+        {
+            return ExecuteReplicationAsync(
+                request,
+                response,
+                stream => stream.CopyToAsync(response.Body, 81920, cancellationToken),
```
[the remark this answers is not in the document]
NightOwl888: Gotcha. Since you had a perfectly good reason to choose this value, then leave it as-is. It just seemed odd that it was not a power of two but is exactly 1 digit more than a power of two, so I figured it was a typo.

src/Lucene.Net/Support/IO/StreamExtensions.cs
```diff
@@ -210,5 +213,154 @@ public static long ReadInt64(this Stream stream)
...
+        }
+
+        public static async Task<string> ReadUTFAsync(this Stream input, CancellationToken cancellationToken = default)
+        {
+            if (input is null)
+                throw new ArgumentNullException(nameof(input));
+
+            byte[] lenBuff = new byte[2];
+            int readLen = await input.ReadAsync(lenBuff, 0, lenBuff.Length, cancellationToken).ConfigureAwait(false);
+            if (readLen < lenBuff.Length) throw new EndOfStreamException();
```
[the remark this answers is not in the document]
NightOwl888: Since @paulirwin decided against `System.IO.Pipelines`, then yes, we need to change the logic to also check for `== 0`, since that is also a possibility that means end of stream. In each place where the check is done, that is.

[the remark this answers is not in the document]
NehanPathan: @NightOwl888 
Thanks for clarifying — from my reading, our current < length check and other similar checks already covers the == 0 case i guess. So it seems more like a style/documentation improvement rather than a functional gap. Do you  still want me to make the check more explicit for clarity.....

src/Lucene.Net/Support/IO/StreamExtensions.cs
NehanPathan: @NightOwl888 @paulirwin 
All suggested changes have been applied — method renames, ArrayPool usage for buffers, and cleaned up comments.
Please let me know if anything looks off or feels odd. Happy to adjust further if needed.

NehanPathan: @NightOwl888 

Not sure, but I think the issue might be here:

```csharp
var syncIoFeature = context.Features.Get<IHttpBodyControlFeature>();
if (syncIoFeature != null)
{
    syncIoFeature.AllowSynchronousIO = true;
}

service.Perform(context.Request, context.Response);
```

If the synchronous `Perform()` method was removed and only `PerformAsync()` exists in the extensions, then calling `Perform()` while allowing synchronous I/O could cause the CheckIndex failures.

In my current setup, using `AllowSynchronousIO = false` and calling `PerformAsync()` works consistently. 

In mine I have updated both places using `AllowSynchronousIO = false` and calling `PerformAsync()
ie lline 93 as you suggest
and also ine 58 
i guess may be u missed out line 58

NightOwl888: The bottom line is that there are currently no tests in this repository that are testing these changes you are making. So, we need some way to verify they work.

NightOwl888:
> @NightOwl888
> 
> Thanks — that makes sense.
> 
> Your approach sounds clean and avoids too much duplication. I’m happy to try setting that up now, unless you’d prefer to handle it yourself since you already have the structure in mind.
> 
> Either way, I’ll follow your lead — just let me know what you prefer.

I have been working on this, but came to a bit of a fork in the road when it comes to setting it up because of how the different configurations are setup between legacy apps and new. I have a setup that simply does what I said, but it would be much cleaner if there were only 1 `NewHttpServer()` overload without so many branching paths, so I am attempting to unify the configuration a bit more between older and newer versions of ASP.NET Core.

Anyway, let me work on it a bit more before handing it off to you. I think I can come up with a compromise without digging too deep into the custom configuration territory (where `Lucene.Net.Extensions` is meant to shine).

> Also, one quick thing — after my last push yesterday, I noticed in the checks that 2 tests failed (I also got the email around 5/6 ie github try to re run but it fail). But now everything shows green and all tests are passing. Do you know if there was any specific reason for that? Did you make some changes, or does GitHub sometimes re-run and fix it automatically? I wasn’t fully sure why that happened.

Yeah, I spotted the failures and saw that they were due to server connectivity problems, not actual software failures. So, I restarted the tests and they succeeded the second time.

paulirwin:
> so I am attempting to unify the configuration a bit more between older and newer versions of ASP.NET Core.

Versions older than ASP.NET Core 8 are [out of support](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core), so if you're referring to <= 7 in use by the existing AspNetCore sample code, I think we should drop support for them. I do not think we need the burden of supporting older, unsupported versions of ASP.NET Core. (Part of the reason why we're looking at moving the extensions into a separate repo.)

NightOwl888:
> > so I am attempting to unify the configuration a bit more between older and newer versions of ASP.NET Core.
> 
> Versions older than ASP.NET Core 8 are [out of support](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core), so if you're referring to <= 7 in use by the existing AspNetCore sample code, I think we should drop support for them. I do not think we need the burden of supporting older, unsupported versions of ASP.NET Core. (Part of the reason why we're looking at moving the extensions into a separate repo.)

Understood. But replicator is both a client and sever-side technology. For simplicity, we are testing both client and server-side on the same target framework using `Microsoft.AspNetCore.TestHost`. I suppose we could aim to find a way to host the server-side on ASP.NET Core while we test the client side on .NET Framework if we are going to discuss dropping server-side support on all but .NET Core.

Do note there are other server-side HTTP stacks that `ReplicationServer` could be hosted on that are in support.

- ASP.NET (System.Web)
- WCF (server-side)
- HttpListener
- OWIN / Katana

I don't feel strongly about supporting these stacks, but just thought I would point out there are other options for hosting a replication server than on ASP.NET Core. But it doesn't make a lot of sense to test them explicitly (which is why we stuck with `Microsoft.AspNetCore.TestHost`). We specfically added interfaces to be able to support these as well as future HTTP stacks.

However, it seems like it would be extremely limiting if we didn't allow .NET Framework users to use the client-side replicator components.

NehanPathan:
> Minor feedback added about some comments in the code. @NehanPathan I'm open to approving this without the J2N-ported unit tests if you want to do that in a follow-up PR, or let me know if you are already working on porting those and will have that done soon. Thanks!

@paulirwin 
 I have already added the J2N-ported async tests in the last commit, as per Shad’s suggestion. 
 pls correct me if i misunderstood....
And
Regarding the commented-out code: that is our old synchronous code. We left it commented for future reference while adding the new async code. Shad is currently working on that part only ie separating the testing code so that both synchronous and asynchronous code can be tested independently.

src/Lucene.Net.Tests.Replicator/Http/ReplicationServlet.cs
```diff
@@ -52,14 +52,16 @@ public void Configure(IApplicationBuilder app, IReplicationService service, Repl
             {
                 // LUCENENET: This is to allow synchronous IO to happen for these requests.
                 // LUCENENET TODO: Allow async operations from Replicator.
```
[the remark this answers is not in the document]
NightOwl888: Done

src/Lucene.Net.Tests.Replicator/Http/ReplicationServlet.cs
```diff
@@ -84,14 +86,15 @@ public async Task InvokeAsync(HttpContext context)
         {
             // LUCENENET: This is to allow synchronous IO to happen for these requests.
             // LUCENENET TODO: Allow async operations from Replicator.
```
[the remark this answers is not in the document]
NightOwl888: Done

src/Lucene.Net.Replicator/Support/Http/Abstractions/IReplicationResponse.cs
```diff
@@ -42,5 +44,12 @@ public interface IReplicationResponse
         /// Flushes the reponse to the underlying response stream.
         /// </summary>
         void Flush();
+
+        /// <summary>
+        /// Flushes the response to the underlying response stream asynchronously.
+        /// </summary>
+        /// <param name="cancellationToken">Optional cancellation token.</param>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        Task FlushAsync(CancellationToken cancellationToken = default);
```
NightOwl888, line 53: This is the only thing I was unsure about so I withheld comment until I could do some research.

While async HTTP is supported on all server stacks as far back as we care to support, the direction things are headed will likely end up with no synchronous HTTP support on stacks in the future. So, it is not reasonable for there to be a `Flush()` implementation on every implementation of `IReplicationResponse`. A technology lacking synchronous support would have no good options for implementing it.

### Throwing an exception

```c#
void Flush() => throw new NotSupportedException();
```

This seems like a hack and underscores the fundamental issue with the interface design.

### Faking it

```c#
void Flush() => FlushAsync().GetAwaiter().GetResult();
```

Also not a good option, since it can lead to deadlocks.

## Options

### Option A: Keep both, but make Flush() optional

- Document that `Flush()` may throw `NotSupportedException` if the underlying server disallows sync I/O.
- This makes the abstraction maximally portable, but puts burden on consumers to handle failure.

### Option B: Move Flush() to a separate interface

- Keeps async-first in the main abstraction.
- Allows servers that can’t support sync to just implement the async interface.
- Clear separation: consumers who truly need sync must check for it explicitly.

There are a couple of different ways to do this:

#### Inheritance

```c#
public interface ISynchronousReplicationResponse : IReplicationResponse
{
    void Flush();
}
```

This means that if someone only wanted a synchronous `Flush()` they would also be required to implement `FlushAsync()`.

#### Separate interfaces

```c#
public interface IReplicationResponse
{
    int StatusCode { get; set; }

    Stream Body { get; }

    void Flush();
}

public interface IAsyncReplicationResponse
{
    int StatusCode { get; set; }

    Stream Body { get; }

    Task FlushAsync(CancellationToken cancellationToken = default);
}
```

This one has my vote. It cleanly separates the sync and async workflows so a server could implement one, the other, or both.

The naming convention could be `ISynchronousReplicationResponse` and `IReplicationResponse`, which might be better if we end up dropping the former in the future. But in the BCL, there are interfaces such as `IAsyncDisposable` and `IAsyncEnumerable<T>`, but none specifying synchronous I/O, so the above convention is more in line with the BCL.

I looked through the usages, and there are no issues at all in the production code. In tests, we can just use `IAsyncReplicationService` and switch if we need to:

```c#
if (service is IReplicationService synchronousService)
    synchronousService.Perform(...);
else
    throw new NotSupportedException("The service doesn't support synchronous I/O");
```
---------------------

Whatever option we choose, we should give the same treatment to `IReplicationService.Perform()` and `IReplicationService.PerformAsync()` for consistency.

NightOwl888 (commented):
I have pushed a few commits to update the tests for both synchronous and asynchronous I/O.

Please review my changes, but everything before my commits looks pretty good (aside from one issue, mentioned inline).

Also, based on your feedback, I went ahead and had ChatGPT create a `TestServer` based on `System.Net.HttpListener` so we can continue testing server functionality on target frameworks older than `net8.0` without having to reference old versions of `Microsoft.AspNetCore.TestHost`. I also dropped all targets from `Lucene.Net.Replicator.AspNetCore` except for `net8.0`. But I am sure that whole assembly will be going away soon, anyway.

Since configuring an HTTP server to use `ReplicationService` is a bit of a black art, the tests were designed not only as tests but also as a demos on a couple of different options for integrating it into ASP.NET Core. However, once the replicator extensions in `lucenenet-extensions` have matured, we can move all demos over there and do only the simplified `HttpListener` tests on all target frameworks so we can drop the dependency on `Microsoft.AspNetCore.TestHost`.

I also spotted a potential issue with adding `FlushAsync()` to the same interface as `Flush()` and `PerformAsync()` on the same interface as `Perform()`. See my inline comment for details.

src/Lucene.Net.Replicator/Support/Http/Abstractions/IReplicationResponse.cs
paulirwin: I agree with separate interfaces as well. Let's get it done. +1 for `IAsyncReplicationResponse` and `IAsyncReplicationService`.

NightOwl888 (changes requested):
Well, crap. I totally missed the fact that both `Flush()` and `FlushAsync()` are simply delegating the call to the public property `Body` on the same interface. So, rather than splitting `IReplicationService` into 2 interfaces, it is more sensible to remove `Flush()` and `FlushAsync()` from the interface and change all callers to do the operation directly on the `IReplicationService. Body`.

The simplest way back is to revert 349621f595a0fe2d396dc537ae1b6c018e301c13 and then remove `Flush()` and `FlushAsync()` from `IReplicationResponse`, then fix all of the compile errors.

Since `IReplicationService` is something we own and isn't likely something a user will need to implement, we can also just revert it to include both overloads, also.

src/Lucene.Net.Replicator/Support/Http/Abstractions/IReplicationService.cs
```diff
@@ -1,4 +1,7 @@
 using System;
+using System.Threading;
```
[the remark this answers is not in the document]
NehanPathan: We need System.Threading and System.Threading.Tasks for PerformAsync; System is optional and can be removed

src/Lucene.Net.Replicator/Http/ReplicationService.cs
```diff
@@ -118,75 +121,151 @@ private static string ExtractRequestParam(IReplicationRequest request, string pa
...
+                    }
+                    else
+                    {
+                        response.Body.Write(new byte[] { 1 }, 0, 1);
+                        token.Serialize(new DataOutputStream(response.Body));
+                    }
+                    return Task.CompletedTask;
+                },
+                () => { response.Flush(); return Task.CompletedTask; }
+            ).GetAwaiter().GetResult(); // // keep sync behavior
```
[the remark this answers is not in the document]
NehanPathan: done

PR #1172, opened 2025-09-05
Cache NuGet Packages (GitHub Actions)

### LINKED ISSUE #857: Add NuGet Caching to GitHub Actions

paulirwin:

Adds caching of NuGet packages to GitHub Actions.

Fixes #857

## Description

This is a re-creation of PR #866, and will experiment with using the updated caching options available in GitHub Actions.


paulirwin:
> I think a way around that may be to simply add the $(TargetFramework) and OS to the cache key.

That's not really the problem (or solution), sorry, I should have been clearer. If the lock files have all of our target frameworks in source control, it would have no problem restoring the packages on macOS or Linux build agents, where the cache would be used, so there's no need to differentiate that by OS AFAICT (at least not for this specific problem). The problem I was referring to above is that for _contributors building on macOS or Linux_, the lock files are regenerated on our dev machines each time we build since it doesn't have that target. That makes it at best highly annoying to avoid having to commit those files. I suppose most of the "at worst" scenarios are protected by CI builds... so perhaps this isn't a "blocker" but more of an "extremely annoying impediment to contribution."

Because of that, I'll reopen this and leave it as draft, and see if anyone has any other ideas. We can also revisit this in the future if we are able to upgrade IKVM without using Mono, accept Mono as an annoying contributor requirement (and go through the work of adding it to our CI pipelines), or (fingers crossed) drop .NET Framework support (at least for Analysis.OpenNLP) in the future.

NightOwl888:
FYI - There is an INFRA-approved cache option called Stash that allows more than 10GB and doesn't require any special permissions to set up. It is meant to be a drop-in replacement for actions/cache.

https://github.com/apache/infrastructure-actions/blob/main/stash/README.md

paulirwin:
> I propose that we change this PR to use **restore caching** only, because that will certainly have a big impact. From that point, we can decide whether it will be worth the effort to enable incremental build caching (separate PR). My best guess is that it will because we have so many parallel builds that share dependencies.
> 
> I would be happy to fix the PR if you agree.

Agreed, and please proceed!

NightOwl888:
The config is too complex to go through everything, but here is a representative of the difference between a cache hit and cache miss comparing apples to apples (same project, same os, same platform, same target framework). In this case, we shaved nearly 2 minutes off of the runtime.

### Cache Miss

image

### Cache Hit

image

As you can see above, the run time for build is ~35s, so we may be able to shave another 15-20s if we enable incremental build caching, although I am not sure whether that will offset the extra amount of time downloading the cache will take. Not sure it is worth the complexity. I went to find the worst case I could think of (OpenNLP) and it still only takes about 1 minute to build. The below is also a cache miss, so the runtime is still fairly short (due to the simple tests).

We gain a lot by building each project on a separate job here on GitHub Actions. However, on Azure DevOps it will probably be worth it to do both types of caching, since the build takes a significant amount of time when the solution is built all at once.

image

paulirwin:
GitHub won't let me approve my own PR, even though it includes your changes... so, consider this message approval 😆

PR #1173, opened 2025-09-05
Fix integer overflow in IndexInputStream.Read(), #1158

### LINKED ISSUE #1158: negative readCount in IndexInputStream.Read

### Describe the bug

The readCount might be negative in [ IndexInputStream.Read ](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Replicator/IndexInputInputStream.cs#L66C1-L72C10) and the exception is thrown

- Message :Non-negative number required. Parameter name: length
- StackTrace: 
   at J2N.IO.MemoryMappedFiles.MemoryMappedViewByteBuffer.Get(Byte[] destination, Int32 offset, Int32 length), 
   at Lucene.Net.Replicator.IndexInputStream.Read(Byte[] buffer, Int32 offset, Int32 count), 

```
public override int Read(byte[] buffer, int offset, int count)
{
	int remaining = (int) (input.Length - input.Position); // remaining might be a negative number due to (int) longNum
	int readCount = Math.Min(remaining, count); //  readCount might be a negative number
	input.ReadBytes(buffer, offset, readCount);
	return readCount;
}
```

a possible fix could be 
```
public override int Read(byte[] buffer, int offset, int count)
{
	long remaining = input.Length - input.Position; 
	int readCount = (int)Math.Min(remaining, count);
	input.ReadBytes(buffer, offset, readCount);
	return readCount;
}
```

### Expected Behavior

_No response_

### Steps To Reproduce

```
[Test]
public void IndexInputStream_ReadIssueWhenInputLengthGreaterThanIntMax()
{
	var mockIndexInput = new Mock<IndexInput>(MockBehavior.Strict, "abc");
	mockIndexInput.Setup(x => x.Length).Returns(long.MaxValue).Verifiable();
	mockIndexInput.Setup(x => x.Position).Returns(0).Verifiable();
	mockIndexInput.Setup(x => x.ReadBytes(It.IsAny<byte[]>(), It.IsAny<int>(), It.Is<int>(v => v < 0)))
		.Throws(new ArgumentException("length cannot be negative"))
		.Verifiable();

	var indexInputStream = new IndexInputStream(mockIndexInput.Object);

	Assert.That(() => indexInputStream.Read(new byte[1], 0, 1), Throws.InstanceOf<ArgumentException>());
	mockIndexInput.VerifyAll();
}
```

### Exceptions (if any)

_No response_

### Lucene.NET Version

_No response_

### .NET Version

_No response_

### Operating System

_No response_

### Anything else?

_No response_

paulirwin:

Fix integer overflow in IndexInputStream.Read()

Fixes #1158

## Description

When input.Length - input.Position exceeds int.MaxValue, casting to int causes overflow resulting in negative values. This fix uses long for the remaining calculation before safely casting to int via Math.Min.

Thanks to @ChaoMaWisetech for providing the proposed solution.

### DIFF
--- a/src/Lucene.Net.Replicator/IndexInputInputStream.cs
+++ b/src/Lucene.Net.Replicator/IndexInputInputStream.cs
@@ -65,8 +65,8 @@ namespace Lucene.Net.Replicator
 
         public override int Read(byte[] buffer, int offset, int count)
         {
-            int remaining = (int)(input.Length - input.Position); // LUCENENET specific: Renamed from getFilePointer() to match FileStream
-            int readCount = Math.Min(remaining, count);
+            long remaining = input.Length - input.Position; // LUCENENET specific: Renamed from getFilePointer() to match FileStream
+            int readCount = (int)Math.Min(remaining, count);
             input.ReadBytes(buffer, offset, readCount);
             return readCount;
         }
--- a/src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
+++ b/src/Lucene.Net.Tests.Replicator/IndexInputStreamTest.cs
@@ -1,6 +1,7 @@
 using Lucene.Net.Attributes;
 using Lucene.Net.Index;
 using Lucene.Net.Replicator;
+using Lucene.Net.Store;
 using Lucene.Net.Util;
 using NUnit.Framework;
 using System;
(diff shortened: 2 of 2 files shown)


NightOwl888 (changes requested):
Thanks for fixing this. I double-checked that the test fails with the old code.

> I noticed that the `IndexInputStream` subclasses `System.IO.Stream` and implements only part of it, so it is probably worth checking whether all of the public APIs are sensible to call (in a separate issue, of course). If the methods implemented were all from Java, there may be issues because .NET uses different cascading behavior in the base class than Java. So, we may need to implement more of the virtual methods to make it fully functional or alternatively throw `NotSupportedException`. Also, as pointed out by the TODOs, there are some weird decisions on error types that are thrown from `IndexInputStream` to review.

PR #1178, opened 2025-09-06
pull_request_template.md: fix link to the `Contributor Guide`

jbampton:
Used a relative link

Fixes #{issue number} (in this specific format)

## Description

{Detail}

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,7 +5,7 @@
 <!-- If this is your first PR in the Lucene.NET repo, please run through the checklist
 below to ensure a smooth review and merge process for your PR. -->
 
-- [ ] You've read the [Contributor Guide](../CONTRIBUTING.md) and [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html).
+- [ ] You've read the [Contributor Guide](https://github.com/apache/lucenenet/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html).
 - [ ] You've included unit or integration tests for your change, where applicable.
 - [ ] You've included inline docs for your change, where applicable.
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -5,7 +5,7 @@
 <!-- If this is your first PR in the Lucene.NET repo, please run through the checklist
 below to ensure a smooth review and merge process for your PR. -->
 
-- [ ] You've read the [Contributor Guide](https://github.com/apache/lucenenet/blob/main/CONTRIBUTING.md) and [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html).
+- [ ] You've read the [Contributor Guide](../CONTRIBUTING.md) and [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html).
```
NightOwl888, line 8: When submitting a PR, the default view is plain text. So from that perspective, the relative URL cannot be copied and pasted into a browser.

`../CONTRIBUTING.md`

However, if the user realizes they can click on the Preview tab to get it to resolve to a complete URL, it shouldn't be a problem. Unfortunately, there may be some users who won't come to that realization.

The correct absolute URL is:

`https://github.com/apache/lucenenet/blob/master/CONTRIBUTING.md`

IMO, we should make it absolute unless there is some other trick to make a PR template into a better UX. This template originated from the ASP.NET Core repo and since the cleanup of the URLs were manual, the main > master conversion was overlooked.

NightOwl888 (changes requested):

.github/pull_request_template.md
jbampton, line 8:
```suggestion
- [ ] You've read the [Contributor Guide](https://github.com/apache/lucenenet/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html).
```

PR #1193, opened 2025-09-29
Drop .NET Standard 2.1 (netstandard2.1) target

### LINKED ISSUE #1040: Proposal: Drop .NET Standard 2.1 target

### Task description

As part of implementing #1039, it dawned on me. I'd like to propose that we drop the .NET Standard 2.1 (netstandard2.1) target. I don't think it makes much sense going into 2025 with the limited amount of currently-supported implementations that support it that aren't already covered by our netstandard2.0 or net8.0 targets.

Docs: https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1

As a reminder, no version of .NET Framework or UWP supports netstandard2.1, so those can either use our net462 target or netstandard2.0. Xamarin is now out of support as well, along with .NET (Core) 3.0-7.0.

The only _currently supported_ .NET implementations that support netstandard2.1 in that list are:

- .NET 8 (we already have a net8.0 target for this)
- .NET 9 (would use our net8.0 target)
- Mono 6.4+
- Unity 2021.2, maybe?

The first two are covered by our net8.0 target and thus don't apply. (If we added a net9.0 target, it also wouldn't use the netstandard2.1 build.) The latter two presumably have no issues running netstandard2.0, and Mono 6.4 is listed on the netstandard2.0 tab as well. As far as I can tell, Unity releases get 2 years of LTS support, so 2021.2 should be out of support by now anyways, if I understand correctly. But again, even then, it should be able to use netstandard2.0.

So, basically, we'd be keeping this target around for some slight possible performance optimizations for... people running apps on Mono. (And again, I'm referring to Mono specifically here, not Xamarin running on Mono, which is out of support.)

This target was added in https://github.com/apache/lucenenet/commit/0482957f8c265db9c8a892bff76942efa68b03f5 in 2019, and I don't see any issues about justifications for why it was added. I assume it was added as an attempt to have just one target to support both Xamarin and .NET Core 3.x-5 primarily, since the next major target that was added was net6.0. 

For posterity, the Lucene.NET compilation features that .NET Standard 2.1 supports that 2.0 does not are:
- FEATURE_ARRAY_FILL
- FEATURE_CONDITIONALWEAKTABLE_ENUMERATOR
- FEATURE_CONDITIONALWEAKTABLE_ADDORUPDATE
- FEATURE_NUMBER_PARSE_READONLYSPAN
- FEATURE_STREAM_READ_SPAN
- FEATURE_STRINGBUILDER_APPEND_READONLYSPAN
- FEATURE_QUEUE_TRYDEQUEUE_TRYPEEK

As we get closer to a 4.8.0 final release, I think we have to determine if this is worth supporting for however many years into the future we have to support it, just for already-older versions of Mono and Unity. Microsoft themselves recommend skipping netstandard2.1 if you can.

Dropping .NET Standard 2.1 would have several benefits:

- One less target to build for and faster local builds (at least until we add another target)
- We would not need to run tests against .NET 6 (as of #1039; .NET 5 before this) to test netstandard2.1, reducing test time and resources
- We therefore would also not need to install the unsupported .NET 6 SDK in the builds
- We could simplify some of the logic for the feature constants

Given the tradeoffs, is it still worth supporting netstandard2.1 as a target today?

paulirwin:

Drops the .NET Standard 2.1 (netstandard2.1) target and .NET 6 tests for it.

Fixes #1040

## Description

In draft until agreement is reached that we should drop this target, and to ensure tests/workflows pass. See rationale in discussion of #1040. Only Unity and Mono are currently supported runtimes that can use it, and both can use the .NET Standard 2.0 target. (Mono should even be able to use the .NET Framework target.)

Note that the GitHub workflow files previously had a UTF-8 BOM. When running the script to generate the workflows, it removed the BOM.

### DIFF
--- a/src/Lucene.Net.Tests.Analysis.OpenNLP/Lucene.Net.Tests.Analysis.OpenNLP.csproj
+++ b/src/Lucene.Net.Tests.Analysis.OpenNLP/Lucene.Net.Tests.Analysis.OpenNLP.csproj
@@ -39,7 +39,7 @@
     <IsPublishable>false</IsPublishable>
     <IsTestProject>false</IsTestProject>
     <!-- For the IDE, the above doesn't work. Redirect to a supported test framework instead. -->
-    <TargetFramework Condition=" '$(TargetFramework)' == 'net472' ">net48</TargetFramework>
+    <TargetFramework>net48</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>


src/Lucene.Net.Tests.Analysis.OpenNLP/Lucene.Net.Tests.Analysis.OpenNLP.csproj
```diff
@@ -34,12 +34,11 @@
...
 
   <!-- NOTE: Technically, we can test on net472, but we don't want to publish net472 assemblies when we are testing netstandard2.0. -->
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net472' Or $(TargetFramework.StartsWith('net6.')) ">
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'net472'">
     <!-- For CI, we don't publish the projects and exclude them from the test run by setting IsPublishable=false and IsTestProject=false -->
     <IsPublishable>false</IsPublishable>
     <IsTestProject>false</IsTestProject>
     <!-- For the IDE, the above doesn't work. Redirect to a supported test framework instead. -->
-    <TargetFramework Condition=" $(TargetFramework.StartsWith('net6.')) ">net8.0</TargetFramework>
     <TargetFramework Condition=" '$(TargetFramework)' == 'net472' ">net48</TargetFramework>
```
NightOwl888: We can now remove this condition.

NightOwl888 (approved):
Alright, looks like this is (finally) good to go.

------------------------------------

The issue with GitHub Actions was that MS released a couple of bad SDKs for .NET 8 and .NET 9.

https://github.com/dotnet/msbuild/issues/12751

I regenerated the test workflows with explicit SDK versions. We will need to keep doing this until Microsoft has a patch for the broken ones (8.0.416 and 9.0.307). Here is the Powershell command I used:

```powershell
./.github/workflows/Generate-TestWorkflows.ps1 -DotNet9SDKVersion 9.0.306 -DotNet8SDKVersion 8.0.415
```
---------------------------------------

There is also an unrelated issue with the actions/setup-dotnet task where it installs *both* the requested version and the .NET 10 SDK. But this doesn't affect our build in any way.

https://github.com/actions/setup-dotnet/issues/679

Issue #1185, opened 2025-09-19
Self-host all external website dependencies


paulirwin:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Task description

Per [the Committer's FAQ](https://privacy.apache.org/faq/committers.html), we must self-host all external resources like Font Awesome icons, fonts, and book images. 

See https://whimsy.apache.org/site/project/lucenenet for details under "Resources"

mmafrar:
@paulirwin I’m not participating in Hacktoberfest, but I’d like to work on this task. Can I proceed?

paulirwin:
@mmafrar Absolutely! I'll assign the issue to you. And thanks for your contribution!

PR #1197, opened 2025-10-02
fix: Self-host all external website dependencies

mmafrar:

Fix CSP violations by self-hosting external resources

See #1185 (this only does the fonts, we still have the book images to do, but require permission to host them).

## Description

External stylesheets and images from domains like cdnjs.cloudflare.com, fonts.googleapis.com, and images-na.ssl-images-amazon.com were blocked due to restrictive Content Security Policy (CSP). All dependencies are now self-hosted to comply with the site's CSP configuration.


NightOwl888:
@mmafrar 

Thanks for the PR.

While we do want to host these files on our website (which we do when we deploy the built site to https://github.com/apache/lucenenet-site, which backs our hosting), we ideally wouldn't be committing the binary files to this repository. Rather, we should have the build scripts download them prior to deploying.

Ideally, we would set up some sort of package manager such as npm, bower, etc. to download the versioned packages such as font-awesome to make this process simpler.

The build scripts are named `site.ps1` and `docs.ps1` as referred to in the [website & docs documentation](https://lucenenet.apache.org/contributing/documentation.html), which also describes how we deploy the website and docs site.

Let us know if you need any assistance with any of this (for example if you are unfamiliar with package managers or powershell), and we will do our best to assist.

mmafrar:
@NightOwl888 I initially assumed that binary files should be committed to the repository, as the issue description didn’t clearly specify otherwise. I’ll investigate the correct approach and work on implementing it. If I encounter any blockers or am unable to proceed, I’ll reach out for guidance.

mmafrar:
@NightOwl888 Is the https://github.com/apache/lucenenet-site repository intentionally unlicensed?

NightOwl888:
> @NightOwl888 Is the https://github.com/apache/lucenenet-site repository intentionally unlicensed?

Good point. You are right, it should have a `LICENSE.txt` file in the `main` branch.

If you are so inclined, you may submit a PR. Use the license file from here: https://github.com/apache/lucenenet-codeanalysis-dev/blob/main/LICENSE.txt, but change the beginning of the date range to 2020.

mmafrar:
> > @NightOwl888 Is the https://github.com/apache/lucenenet-site repository intentionally unlicensed?
> 
> Good point. You are right, it should have a `LICENSE.txt` file in the `master` branch.
> 
> If you are so inclined, you may submit a PR. Use the license file from here: https://github.com/apache/lucenenet-codeanalysis-dev/blob/main/LICENSE.txt, but change the beginning of the date range to 2020.

@NightOwl888 Here is the link: https://github.com/apache/lucenenet-site/pull/57

PR #1202, opened 2025-10-06
Upgrade dev analyzers to 1.0.0-alpha.18; make WindowsFsyncSupport internal, #1100

### LINKED ISSUE #1100: Dev Analyzer Rule to ensure there are no public types in the Lucene.Net.Support namespace

### Task description

While we probably will have a `Support` folder in the repository for some time going forward, it is important to enforce the fact that we won't have a public `Lucene.Net.Support` namespace. This is easier to enforce using a Roslyn code analyzer than trying to keep track of during reviews because we can easily miss it.

The developer should get a build failure if they add a public type to the `Lucene.Net.Support` namespace or any namespace below that namespace. They should be instructed:

- If the type is not intended for users, make it internal
- If the type is intended for users:
   - If it is an adaption of something available in the JDK, consider submitting it to J2N
   - If it is more specialized, move it to an appropriate namespace below Lucene.Net, such as Lucene.Net.Util

The latter two options can be discussed either on the dev mailing list or on Slack to help make the right choice.

> Side Note: I noticed that the `Lucene.Net.Support.IO.WindowsFsyncSupport` type is public and made it into the 4.8.0-beta00017 release. This would make a good test case for the analyzer.

paulirwin:

Fixes #1100

## Description

Upgrades dev analyzers which adds the LuceneDev1005 analyzer for public types in the Lucene.Net.Support namespace. Fixes the only violation: WindowsFsyncSupport.

Note that upgrading this analyzer reference adds a lot of build warnings due to the new LuceneDev1006 analyzer.

### DIFF
--- a/.build/dependencies.props
+++ b/.build/dependencies.props
@@ -40,7 +40,7 @@
     <!-- J2N will break binary compatibility in 3.0.0 to fix the APIs of collection types -->
     <J2NPackageVersion>[2.2.0-alpha-0021, 3.0.0)</J2NPackageVersion>
     <LiquidTestReportsMarkdownPackageVersion>1.0.9</LiquidTestReportsMarkdownPackageVersion>
-    <LuceneNetCodeAnalysisDevPackageVersion>1.0.0-alpha.6</LuceneNetCodeAnalysisDevPackageVersion>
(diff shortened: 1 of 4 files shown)


NightOwl888 (approved):
This looks good. However, we should also disable `LuceneDev1006`, since we will need to update J2N with a string interpolation handler and the code fix before we can fix it.

PR #1212, opened 2025-10-27
Change HttpClientBase.ExecutePost to match upstream better and drop Newtonsoft.Json dependency

### LINKED ISSUE #1176: Change HttpClientBase.ExecutePost to match upstream better and drop Newtonsoft.Json dependency

### Task description

The `HttpClientBase` implementation is currently:

https://github.com/apache/lucenenet/blob/53f020be99b7fa5196666a471a38fe9f9a8dbdce/src/Lucene.Net.Replicator/Http/HttpClientBase.cs#L173-L188

There is only one serialization option, and it is JSON. This does not match the upstream code:

https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/replicator/src/java/org/apache/lucene/replicator/http/HttpClientBase.java#L156-L167

```java
  /**
   * <b>internal:</b> execute a request and return its result
   * The <code>params</code> argument is treated as: name1,value1,name2,value2,...
   */
  protected HttpResponse executePOST(String request, HttpEntity entity, String... params) throws IOException {
    ensureOpen();
    HttpPost m = new HttpPost(queryString(request, params));
    m.setEntity(entity);
    HttpResponse response = httpc.execute(m);
    verifyStatus(response);
    return response;
  }
```

Per ChatGPT, a better translation of this method is:

```c#
    protected virtual HttpResponseMessage ExecutePost(string request, HttpContent content, params string[] parameters)
    {
        EnsureOpen();
        var req = new HttpRequestMessage(HttpMethod.Post, QueryString(request, parameters))
        {
            Content = content
        };
        var resp = client.Send(req);
        VerifyStatus(resp);
        return resp;
    }
```

Which can then be used like:

```c#
    [Test]
    public void TestExecutePostRaw()
    {
        var client = new TestClient();
        var raw = "hello world";

        var content = new StringContent(raw);
        var resp = client.ExecutePost("http://localhost:5000/echo", content);
        var body = resp.Content.ReadAsStringAsync().Result;

        Assert.AreEqual(raw, body);
    }
```

Or, for JSON using System.Text.Json:

```c#
// JSON
var jsonContent = new StringContent(JsonSerializer.Serialize(entity), Encoding.UTF8, "application/json");
ExecutePost("/myendpoint", jsonContent);
```

This allows the client to decide what type of content goes into the body instead of always assuming it will be JSON. But it still supports JSON. It also means that we can drop the dependency on Newtonsoft.Json, since this is the only place in our solution that depends on it directly.

Unfortunately, there are no tests covering this method (no callers at all), so we should create some. Ideally, they would be backported to Java to verify the port. I asked ChatGPT to frame some up:

https://chatgpt.com/share/68bc2717-e4f4-8005-8a53-f4e4bf7417e0

But, I don't agree with ChatGPT that we should keep a JSON overload since it requires NewtonSoft.Json or another JSON serializer (this overload could be provided as an extension method in `lucenenet-extensions`, though).

Note it would also be good to check whether this fixes the reported bad use of HTTP as specified in the comment, but this may fix those issues.

```c#
//.NET Note: No headers? No ContentType?... Bad use of Http? 
```

AnastasiaJacson:

Changed the HttpClientBase.ExecutePost to accept any kind of `HttpContent` and removed Newtonsoft Nuget reference.

Fixes #1176 (in this specific format)

## Description
Changed the HttpClientBase.ExecutePost to accept any kind of `HttpContent` and removed Newtonsoft Nuget reference. Now it is not dependent on Newtonsoft.Json and better matches the upstream ([see](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/replicator/src/java/org/apache/lucene/replicator/http/HttpClientBase.java#L156-L167)).

### DIFF
--- a/.build/dependencies.props
+++ b/.build/dependencies.props
@@ -65,7 +65,6 @@
     <MorfologikPolishPackageVersion>$(MorfologikFsaPackageVersion)</MorfologikPolishPackageVersion>
     <MorfologikStemmingPackageVersion>$(MorfologikFsaPackageVersion)</MorfologikStemmingPackageVersion>
     <NETStandardLibrary20PackageVersion>2.0.3</NETStandardLibrary20PackageVersion>
-    <NewtonsoftJsonPackageVersion>13.0.1</NewtonsoftJsonPackageVersion>
     <NUnit3TestAdapterPackageVersion>4.6.0</NUnit3TestAdapterPackageVersion>
     <NUnitPackageVersion>3.14.0</NUnitPackageVersion>
     <RandomizedTestingGeneratorsPackageVersion>2.7.8</RandomizedTestingGeneratorsPackageVersion>
--- a/src/Lucene.Net.Replicator/Http/HttpClientBase.cs
+++ b/src/Lucene.Net.Replicator/Http/HttpClientBase.cs
@@ -7,6 +7,7 @@ using System.Net;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+#nullable enable
 
 namespace Lucene.Net.Replicator.Http
 {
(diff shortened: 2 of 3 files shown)


paulirwin (approved):
Thanks for this PR. Since you enabled edits by maintainers, I went ahead and did some minor cleanup of HttpClientBase and HttpReplicator that inherits from it. Some comments were no longer relevant with this PR, and the classes in general just needed some cleanup.

I also removed the Newtonsoft.Json dependency version from `.build/dependencies.props` since that is no longer needed.

PR #1198, opened 2025-10-02
Added ASF-required links using drop-down menu and unified navigation

zka26:
- [x] You've read the [Contributor Guide](https://github.com/apache/lucenenet/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html).
- [ ] You've included unit or integration tests for your change, where applicable.
- [ ] You've included inline docs for your change, where applicable.
- [x] There's an open issue for the PR that you are making. If you'd like to propose a change, please [open an issue](https://github.com/apache/lucenenet/issues/new/choose) to discuss the change or find an existing issue.

Added ASF-required links via dropdown and unified navigation.

Fixes #1186 

## Description

This change attempts to standardize the navbar with the newly introduced ASF menus across the Site and Apidocs and, to fix any sort of layout glitches and to improve zoom and resize and to address and solve problem that could have arisen from what was changed. DevTools were used to ensure that everything works accordingly.

### General Design:

- Mobile < 768
- ~769 < Compact < 1199~
- **(New)** 768 =< Wrap < 1200
- 1200 <= Desktop

**Desktop**: Classic horizontal navigation bar extended with the ASF drop-down menu upon hover. The header is fixed when scrolling.

**Mobile:** Center-aligned column menu with a tap/click-to-open ASF accordion.

~**Compact**: Transition in-between Mobile and Desktop for legacy resolutions (or maybe some tablets). It uses the hamburger menu already and the header is no longer fixed when scrolling focusing on the visibility of the content while the ‘Table of Content’ still appears as the left sidebar, and to a certain point the ‘In this Article’ appears as the right sidebar.~

**(New) Wrap:** Reintroduced that the navbar items ( + searchbar) wraps under the logo.

### Improvements:

- Unified navigation: Same feel and look across the Site and Apidocs.
- ~Compact mode (with new breakpoints): When the single classic horizontal navigation bar is no longer possible, we turn to the hamburger menu to prevent the header covering up too much of the content.~
**(New)** Wrap mode instead of the compact mode. The buttons' heights were decreased slightly.
- Ribbon safety: Keeps the hamburger menu clear of the “Fork me on GitHub” ribbon, so it is always accessible, never covered by it.
- [Apidocs-Only] Search overlay: In-between 1400px and 1200px instead of the base search function, we temporarily introduce a search button to open an overlay, therefore delaying the need to switch to hamburger (or multi-line) too early, and Apidocs can switch to it at the same time as the Site.

	
#### Site:
[...]

PR #1232, opened 2026-01-11
Fix/apidocs breadcrumb toc asf

### LINKED ISSUE #1215: Website caching issue
The Lucene.NET website has a server-side caching issue where, after rolling out an update, the returned `etag` value was the same, which caused browsers that have the previously-cached contents to not get the update. More details below.

_Originally posted by @paulirwin in https://github.com/apache/lucenenet/issues/1198#issuecomment-3481361813_

I found the problem (but not the cause, yet) of the caching issue: the `etag` returned has not changed between the old and new sites.

I fortunately had another browser instance still open that had the old cached version. Doing a normal load of the page was sending `If-None-Match` with etag value `"24f2-6428c36d87de9-gzip"`. (If you're unfamiliar, `If-None-Match` sends a previously returned `etag` value, to check if it matches. If it doesn't match, the server should return 200 OK with the updated contents along with a new `etag` header to update its cached value, and if it matches, that means the cached version is up to date so the server returns 304 Not Modified with an empty body.)

The server returned with a 304 Not Modified and also included the `etag` response header with the same value, which is interpreted by the browser as "the server confirmed that my cached version is correct". So this is not a client-side caching issue, this is actually the _server_ saying that the cached version is correct.

I then looked at what was happening in a browser that had the updated HTML/etc, and it was sending and receiving the exact same header value. So something is wrong on the server to where it's returning the same etag value even if the code has changed. 

Note that the server does not send either an `Expires` header nor `Cache-Control` header with `max-age`, so effectively this cache may never expire since, according to the server, the `etag` value is exactly the same as before.

I am not sure why the server is returning the same `etag` value as before, as that is usually produced as effectively a hash of the contents of that page/resource, which should have changed with the updated HTML. If we are producing that etag and hardcoding it somewhere, we should fix that. Or perhaps this is an issue with the server infrastructure?

zka26:

Render apidocs navbar links via DocFX metadata; fix breadcrumb; fix event link

Fixes #1215 (Continuation of #1198)

## Description

- **Fixed breadcrumb rendering:** the breadcrumb is no longer split across multiple lines and is displayed as a single line with / separators.
- **Updated the ASF “Events” link:** adjusted it so it matches the regex.
- **Improved navbar rendering for site + apidocs:** ASF menu is emitted as static HTML during DocFX build on both the main site and apidocs.
- **Now the ASF dropdown menu appears in the** `api / {subsites}`. 

Képernyőkép 2026-01-11 221609


zka26:
Hey @nightowl888 , @paulirwin! 

First of all: Sorry this took so long. Some personal stuff came up and I should have communicated earlier.

I did keep working on it on and I managed to invent a DocFX issue that did not need to exsist in the first place, and I spent too much time trying to fix it up until I found the actual cause and the current approach works now.

The ASF links (and also the navigation links) are now in the docfx metadata. As for the apidocs, they are in the `docfx.global.subsites.json` and I added it `docfx.site.json`'s `globalMetadataFiles` so it appears everywhere, both the main page and the /api pages and it is updated only at one place. (Added auto-update to the `docs.ps1`)

If you want a different structure where these links values are, just let me know and I will adjust it quickly! 
Also, I left the exsisting JS-based runtime injection, I did not notice any issue, but let me know if you want anything about that.

I just realized that the site was 'left behind' as I was troubleshooting the apidocs, and I will quickly make them more uniform by modify the site to match how the links are stored and called at the navbar partial. Then I will revert it from draft.

zka26:
Now the site and apidocs  are more uniform in how the navbar links are stored and rendered.

Apologies about the delay again and thanks for your patience!

paulirwin (approved):
Sorry for the delay on this, looks good! I pushed up some small changes to whitespace, a minor accessibility fix, adding in the `a` tag titles to match the other file in the site project, and shortened the header link text so that it fits without wrapping. 

Before:
Image

After:

Image

PR #1235, opened 2026-01-19
Bugfix: Lucene.Net.Replicator.SessionToken ToString() overload doesn't output values correctly

### LINKED ISSUE #1226: Lucene.Net.Replicator.SessionToken ToString() overload doesn't output values correctly

### Describe the bug

ReplicationClient logs Lucene.Net.Replicator.SessionToken as a string and its ToString() overload results in a logging output that can't be used for debugging. For example, on this line:

https://github.com/apache/lucenenet/blob/36f587de8481a39a2695305c06728df2a3dedccc/src/Lucene.Net.Replicator/ReplicationClient.cs#L176

```cs
WriteToInfoStream(string.Format("DoUpdate(): handlerVersion={0} session={1}", version, session));
```

It will end up logging like:

```
ReplicationThread - DoUpdate(): handlerVersion= session=id=1 version=2:1 files=System.Collections.Generic.Dictionary`2[System.String,System.Collections.Generic.IList`1[Lucene.Net.Replicator.RevisionFile]]
```

Which is using the ToString() override in SessionToken, but this is just outputting the generic ToString() of a generic list instead of outputting what is in that list.

### Expected Behavior

Ideally it would output all of this information:

Image

However, this will incur some overhead, so then all places where a SessionToken is output as a string within the ReplicationClient would need to be put behind a check like 

```cs
if (infoStream.IsEnabled(INFO_STREAM_COMPONENT))
```

Before the ToString() overload would be called. In fact, this should be done anyways since it is concatenating and allocating a ton of strings even if the above check was in place because the string allocations happen before the call to WriteToInfoStream which does this check.

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### Lucene.NET Version

4.8.latest

### .NET Version

8

### Operating System

windows

### Anything else?

_No response_

Shazwazza:
Fixes #1226 

`SessionToken.ToString()` was not writing out the key/values and instead was just the ToString() output of Dictionary which is the type name. This PR replaces it with the standard `Collections.ToString()` helper used throughout the codebase.

**Changes:**
- Updates `SessionToken.ToString()` to use `Collections.ToString(SourceFiles)`
- Added `using Lucene.Net.Support` import for Collections helper
- Output format unchanged: `id={id} version={ver} files={source1=[file1, file2], source2=[file3]}`
- Removes duplicate call to WriteToInfoStream

This aligns with the standard pattern used in 20+ other files in the codebase and handles nested collections recursively.

### DIFF
--- a/src/Lucene.Net.Replicator/ReplicationClient.cs
+++ b/src/Lucene.Net.Replicator/ReplicationClient.cs
@@ -179,7 +179,6 @@ namespace Lucene.Net.Replicator
                     return;
 
                 IDictionary<string, IList<RevisionFile>> requiredFiles = RequiredFiles(session.SourceFiles);
-                WriteToInfoStream(string.Format("DoUpdate(): handlerVersion={0} session={1}", version, session));
 
                 foreach (KeyValuePair<string, IList<RevisionFile>> pair in requiredFiles)
                 {
--- a/src/Lucene.Net.Replicator/SessionToken.cs
+++ b/src/Lucene.Net.Replicator/SessionToken.cs
@@ -1,4 +1,5 @@
 using J2N.IO;
+using Lucene.Net.Support;
 using System.Collections.Generic;
 using System;
 using System.IO;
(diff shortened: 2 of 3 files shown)


paulirwin (approved):
Looks good, thanks for this PR!

PR #1239, opened 2026-02-01
Upgrade actions/upload-artifact to v6

jbampton:
Upgraded from both v4 and v5 to v6

https://github.com/actions/upload-artifact/releases/tag/v6.0.0


paulirwin (approved):
Looks good. Note that (due to the delay in reviewing this PR; sorry about that) there is now v7 out, but it primarily just adds the ability to upload single artifacts without zipping. Getting the action onto Node 24 with v6 is higher priority. Thanks for the contribution!

PR #1241, opened 2026-02-05
fix: move unused limited license file to tests

alinvelea27:

The file contains code derived from MSDN with Microsoft Limited Public License, which has a 
  platform limitation clause (section 3F). Moving it to TestFramework ensures the main        
  Lucene.Net library is purely Apache 2.0 licensed, as this class is only used in test code.


paulirwin (approved):
Since this file is internal, this seems like a fine change to make. Thanks for the contribution!

PR #1244, opened 2026-04-10
[CI] Add pre-commit hook to auto remove trailing whitespace

jbampton:
https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#trailing-whitespace

Official pre-commit hook that auto removes trailing whitespace

Applied to both Markdown and XML files

refs #1236


paulirwin (approved):
Looks good, validated that the only file with non-whitespace changes is `.pre-commit-config.yaml`. Also successfully tested the pre-commit hooks locally. Thanks for the contribution!

PR #1249, opened 2026-04-11
Add Dependabot and configure pre-commit ecosystem

jbampton:
Monthly updates, limit open pull requests and group updates to reduce noise.

Add 7 day cooldown

https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference

https://docs.github.com/en/code-security/concepts/supply-chain-security/about-dependabot-version-updates

https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,30 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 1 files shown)


paulirwin:
This seems reasonable to me for just pre-commit, since @NightOwl888 has an open PR #1187 to use Renovate for general dependency updates instead of Dependabot. But if it's just scoped to this, I think it's reasonable. I'd say let's let the pre-commit stuff bake a little longer and get any feedback from the community before we adopt this, though.

paulirwin (approved):
Pre-commit has been stable, and no one has chimed in with any objections since my comment a month ago. Approved.

PR #1251, opened 2026-04-11
Add pre-commit hook oxipng for lossless image compression

jbampton:
"Oxipng is a multithreaded lossless PNG/APNG compression optimizer. It can be used via a command-line interface or as a library in other Rust programs. It is fast and highly effective."

https://github.com/oxipng/oxipng

https://en.wikipedia.org/wiki/Lossless_compression

Lossless compression is a class of [data compression](https://en.wikipedia.org/wiki/Data_compression) that allows the original data to be perfectly reconstructed from the compressed data with no loss of information

So we get smaller file sizes with the same quality saves bandwidth etc.

https://github.com/oxipng/oxipng/blob/0cf8b782ffb7af65191e8c702cb503116177a0ec/MANUAL.txt#L167


paulirwin (approved):
Looks good, thanks!

PR #1255, opened 2026-04-15
pre-commit trailing whitespace check more file types

jbampton:
Only `.config` files were auto cleaned up


paulirwin (approved):
Thanks! I'll approve this primarily for the pre-commit-config change to scan more files. The `lib` folder stuff can likely be removed (that NUnit version is no longer referenced, neither is StyleCop. We need to update how we're pulling in apache-rat to match what we're doing in the lucenenet-codeanalysis-dev project, but I'll do that in a separate PR.

PR #1258, opened 2026-04-15
Remove lib folder; add Apache RAT script (#1256)

### LINKED ISSUE #1256: Remove legacy `lib` folder

### Task description

The `lib/` folder in the repo contains three things that are either no longer being used or can be replaced:

- `apache-rat/` - should be replaced with a script to download RAT like we do in https://github.com/apache/lucenenet-codeanalysis-dev/blob/main/rat.ps1
- `NUnit/` - contains Unit ConsoleRunner 3.5.0, no longer referenced. We now bring in NUnit via NuGet.
- `StyleCop.4.5/` - no longer referenced

We should remove this folder from the repo, and reduce the amount of binaries in source control.

paulirwin:

Removes the legacy `lib/` folder from the repo, and adds an Apache RAT script for release auditing.

Fixes #1256

## Description

See #1256 for details/justification. This removes the `lib/` folder that only contained unreferenced binaries and a now-outdated version of the Apache RAT tool. To replace the RAT jar, the rat.ps1 script was copied from the lucenenet-codeanalysis-dev repo, and updated to use version 0.18 (including addressing deprecated command-line syntax; hence the small difference in args from the other repo that is on 0.16.1). 

The AuditRelease MSBuild task was removed since it's now simpler to just run `./rat.ps1`. make-release.md docs were updated accordingly.

The newer versions of Apache RAT require a newer version of Java, which [the changelog docs recommend](https://creadur.apache.org/rat/changes.html) using at least version 21 now. I've updated the link to point to the Microsoft OpenJDK build which has good cross-platform support and is open source, instead of the commercial Oracle JRE 8.


websites/site/contributing/make-release.md
```diff
@@ -26,7 +26,7 @@ For Package Version, NuGet version and branch naming guidelines see the [Version
 
 The following steps need only to be performed once.
 
-- Install the [Java Runtime Environment](https://www.oracle.com/java/technologies/javase-jre8-downloads.html). This is a dependency of the Release Audit Tool.
+- Install the [Java Runtime Environment](https://learn.microsoft.com/en-us/java/openjdk/install?tabs=exe%2Chomebrew%2Cubuntu) version 21 or later. This is a dependency of the Release Audit Tool.
```
NightOwl888, line 31: According to the docs, RAT requires Java 17 or later and only requires Java 21 to build it. Please change it to 17.

https://creadur.apache.org/rat/

> NOTE: The RAT version we were using before seemed to run fine on Java 8.

.build/release.targets
```diff
@@ -20,25 +20,6 @@
 -->
 <Project>
 
-  <PropertyGroup Label="Release Audit Tool Settings">
```
NightOwl888, line 23: This file was my attempt to start to harmonize the build and remove the dependency on PowerShell and depend only on the .NET SDK. We ended up stuck with a scripting language external to MSBuild because at the time we started this, .NET Core was very unstable and didn't play well with .NET Framework. In fact, they were using the DNX JSON format at the time which wasn't compatible at all with MSBuild or .NET Framework, so many options required advanced scripting to get to (if at all). If we hadn't had that limitation, I probably would have written our entire build in MSBuild at the time.

Side note: We ended up renaming the Lucene.Net.Core folder to Lucene.Net because DNX had no way to build an assembly with a different name than its parent folder. That caused a ripple throughout the Lucene.NET community because they thought we were dropping support for .NET Core when in fact we were doing what we had to in order to support it.

Maybe we could have converted the RAT .jar to .NET using IKVM and to eliminate the Java prerequisite, also.
Anyway, downloading the .jar is still an improvement.  I am lukewarm on the idea of using Powershell, but since it is xplat and even new projects I set up ended up using it for CI, it should be okay. Powershell also has an official way to test scripts (Pester) and MSBuild does not.

But I noticed there is nothing in the release procedure about PowerShell being a dependency, although when you get to the part where you run the build locally, PowerShell is required to do that step. We should add that dependency to the setup list.

https://lucenenet.apache.org/contributing/make-release.html#steps-for-the-release-manager

NightOwl888 (changes requested):

websites/site/contributing/make-release.md
paulirwin: I noted in the PR description that the [changelog](https://creadur.apache.org/rat/changes.html) recommends Java 21:

> In addition, the Java language level required to build RAT has been raised to 17. However, we recommend using at least JDK 21 due to a Javadoc issue affecting certain JDK versions (tracked under RAT-497).

Java 21 has been out for longer than .NET 8 has (and we even support newer versions of that) so this seems reasonable.

paulirwin: Also noted re: building versus running. But I figure if we're going to go newer than Java 8 (which does not have great support on all platforms, like macOS arm64), might as well go for a newer LTS version of Java (although I think 25 is a bit _too_ new).

.build/release.targets
paulirwin: PowerShell 7 runs great cross-platform as you mention, unlike other shell scripting methods (Bash on Windows usually requires the not-quite-native Cygwin-based Bash or WSL; Batch scripts are difficult to impossible to run on other platforms than Windows). Also given that it runs on .NET, it is pretty familiar for most .NET developers already. I run it as my primary shell on macOS, too.

Updated the make-release.md docs to list PowerShell as a dependency.

websites/site/contributing/make-release.md
paulirwin: One more minor note on this: OpenJDK vendors like Microsoft are [going to end support for 17 next year](https://learn.microsoft.com/en-us/java/openjdk/support).

PR #1287, opened 2026-05-01
SWEEP: Replace BCL Dictionary<TKey,TValue> with JCG.Dictionary<TKey,TValue> (#1273)

### LINKED ISSUE #1273: Replace `Dictionary<TKey, TValue>` with `JCG.Dictionary<TKey, TValue>` in Java-derived Sources

### Task description

[J2N 2.1.0](https://github.com/NightOwl888/J2N/releases/tag/v2.1.0) has switched to using a low-level implementation of `Dictionary<TKey, TValue>` rather than simply wrapping the one in the BCL to add nullable key support. We have added several improvements since that release to bring it up to about the same level as the .NET 10 implementation in terms of performance. However, the J2N implementation also includes extra features:

1. `null` can be used as a dictionary key
2. Elements can be removed during enumeration (including in .NET Framework)
3. Alternate lookup support for `ReadOnlySpan<char>` (including in .NET Framework)
4. Structural equatable comparisons in `Equals()` and `GetHashCode()`
5. Structural formatting in `ToString()`

The task is to replace the `System.Collections.Generic.Dictionary<TKey, TValue>` declarations with `J2N.Collections.Generic.Dictionary<TKey, TValue>` in all code derived from Java with a few exceptions (as pointed out below). Please consistently use the alias `JCG.Dictionary<TKey, TValue>` throughout the project. If isn't already aliased, simply add this to the top of the file after all other `using` directives (except other aliases, where it should be ordered alphabetically):

```c#
using JCG = J2N.Collections.Generic;
```

## Exceptions

Please do NOT make these replacements in the following cases:

1. Places where LINQ `ToDictionary()` is used, which returns a BCL dictionary (although please note if you find any of these so we can review them)
2. `Lucene.Net.CodeAnalysis.*` projects and directories
3. `Lucene.Net.Replicator.AspNetCore`
4. `lucene-cli`
5. `Lucene.Net.Tests.Cli`
6. `Lucene.Net.Tests.TestFramwork.DependencyInjection`
7. `Lucene.Net.Tests.TestFramework.NUnitExtensions`
8. All types in any `Support` folder or the `Lucene.Net.Support` namespace. Except for the following which we SHOULD replace:
  a. `TagSoup` files
  b. `Sax` files

## Justification

The JDK `HashMap<K, V>` type allows a `null` key while in .NET attempting to do so will throw an `ArgumentNullException`. Since we started out with a replacement that wasn't very well optimized, we used a "squeaky wheel" approach to decide whether to use `System.Collections.Generic.Dictionary<TKey, TValue>` or `Collections.Generic.Dictionary<TKey, TValue>`. But Lucene's test coverage isn't very good in extensions and there may be cases where a `null` key is part of the design that we have missed. So, at this point there is more to gain by using J2N consistently than we lose. Using `System.Collections.Generic.Dictionary<TKey, TValue>` is potentially hiding bugs in untested code.

In J2N 3.0, we are also planning to fix the enumerators on the public API to return the structs rather than virtualizing them to `IEnumerator<KeyValuePair<TKey, TValue>>`, etc, they will be as performant as the BCL by the time we release. But no futher updates to Lucene.Net will be required when this happens.

vwilson:
- [x] There's an open issue for the PR that you are making.

Replace BCL `Dictionary<TKey, TValue>` with `JCG.Dictionary<TKey, TValue>` in Java-derived sources

Fixes #1273

## Description

Mechanical sweep that replaces every `System.Collections.Generic.Dictionary<TKey, TValue>` use in Java-derived sources with `J2N.Collections.Generic.Dictionary<TKey, TValue>`, consistently using the `JCG.Dictionary<TKey, TValue>` alias. ~54 files were already migrated; this sweeps the remaining ~272.

J2N's ``Dictionary<,>`` is feature-comparable to .NET 10's BCL implementation in performance, plus it adds JDK-`HashMap`-equivalent capabilities the BCL type lacks: `null` keys, removal-during-enumeration, `ReadOnlySpan<char>` alternate lookup, structural `Equals`/`GetHashCode`/`ToString`. Java code ported to .NET routinely passes `null` keys; using BCL `Dictionary<,>` was potentially masking bugs in extensions with low test coverage.

### What changed

- 552 `new Dictionary<...>` instantiations → `new JCG.Dictionary<...>`
- 133 files had `using JCG = J2N.Collections.Generic;` added (placed after non-alias `using` directives, alphabetical with other aliases)
- ~25 bare-type LHS declarations updated where the field/local type had to match the new RHS (compiler-driven via CS0029)
- 3 nested-generic args updated (e.g., `JCG.List<Dictionary<...>>` → `JCG.List<JCG.Dictionary<...>>`)
- 1 class inheritance updated: `HashMapAnonymousClass : Dictionary<string, string>` → `: JCG.Dictionary<string, string>` ([TestAnalysisSPILoader.cs](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Tests.Analysis.Common/Analysis/Util/TestAnalysisSPILoader.cs))
- 1 method return type updated: `TestIndexWriter.GetLiveCommitData`

### Exclusions honored (per the issue)

- `Lucene.Net.CodeAnalysis.*` projects
- `Lucene.Net.Replicator.AspNetCore`
- `lucene-cli` and `Lucene.Net.Tests.Cli`
- `Lucene.Net.Tests.TestFramework.{DependencyInjection,NUnitExtensions}`
- All `Support/` folders and the `Lucene.Net.Support` namespace — **except** `TagSoup` and `Sax` under `Lucene.Net.Benchmark/Support/`, which are explicitly in scope per the issue
- LINQ `.ToDictionary()` calls (would return BCL `Dictionary<,>` by design — only one such call exists, and it's in an excluded project)

### Compatibility

Public API surface continues to expose `IDictionary<TKey, TValue>` (interface), not the concrete BCL type. **This is not a binary-breaking change.**

### Verification

- Full solution builds clean (0 errors, 44 pre-existing CA1859 perf warnings, none introduced)
- Core lib builds clean across all target frameworks: `net462`, `netstandard2.0`, `net8.0`, `net10.0`
- Full test suite run: **170 test runs across 36 test projects × multiple TFMs, 52,518 tests passed, 0 failures**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Lucene.Net.TestFramework/Store/BaseDirectoryTestCase.cs
+++ b/src/Lucene.Net.TestFramework/Store/BaseDirectoryTestCase.cs
@@ -10,6 +10,7 @@ using System;
 using System.IO;
 using System.Threading;
 using Assert = Lucene.Net.TestFramework.Assert;
+using JCG = J2N.Collections.Generic;
 using Test = NUnit.Framework.TestAttribute;
 
 namespace Lucene.Net.Store
@@ -376,7 +377,7 @@ namespace Lucene.Net.Store
         //[Test]
         //public virtual void TestMapOfStrings()
         //{
-        //    IDictionary<string, string> m = new Dictionary<string, string>()
+        //    IDictionary<string, string> m = new JCG.Dictionary<string, string>()
         //    {
         //        ["test1"] = "value1",
         //        ["test2"] = "value2"
--- a/src/Lucene.Net.TestFramework/Store/BaseDirectoryTestCase.cs
+++ b/src/Lucene.Net.TestFramework/Store/BaseDirectoryTestCase.cs
@@ -10,7 +10,7 @@ using System;
 using System.IO;
 using System.Threading;
 using Assert = Lucene.Net.TestFramework.Assert;
-using JCG = J2N.Collections.Generic;
+//using JCG = J2N.Collections.Generic;
 using Test = NUnit.Framework.TestAttribute;
 
 namespace Lucene.Net.Store
--- a/src/Lucene.Net.Tests.Analysis.SmartCn/Hhmm/TestBuildDictionary.cs
+++ b/src/Lucene.Net.Tests.Analysis.SmartCn/Hhmm/TestBuildDictionary.cs
@@ -7,7 +7,7 @@ using NUnit.Framework;
 using System;
 using System.IO;
 using Assert = Lucene.Net.TestFramework.Assert;
-using JCG = J2N.Collections.Generic;
+//using JCG = J2N.Collections.Generic;
 
 namespace Lucene.Net.Analysis.Cn.Smart.Hhmm
 {
--- a/src/Lucene.Net.Tests.Analysis.SmartCn/Hhmm/TestBuildDictionary.cs
+++ b/src/Lucene.Net.Tests.Analysis.SmartCn/Hhmm/TestBuildDictionary.cs
@@ -7,6 +7,7 @@ using NUnit.Framework;
 using System;
 using System.IO;
 using Assert = Lucene.Net.TestFramework.Assert;
+using JCG = J2N.Collections.Generic;
 
 namespace Lucene.Net.Analysis.Cn.Smart.Hhmm
 {
@@ -99,7 +100,7 @@ namespace Lucene.Net.Analysis.Cn.Smart.Hhmm
 
     //            string bigramInput = @"real/path/analysis-data/bigramdict.dct";
     //            string bigramOutput = @"real/path/analysis-data/custom-dictionary-input/bigramdict.dct";
-    //            var bigramReplacementFrequencies = new Dictionary<string, int>
+    //            var bigramReplacementFrequencies = new JCG.Dictionary<string, int>
     //            {
     //                ["锲而不舍@、"] = 10000,
     //                ["聆听@了"] = 15000,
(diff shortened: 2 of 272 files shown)


vwilson:
@paulirwin @NightOwl888 hello there. it's been a while since I've done a fork PR and tbh thought I would be able to review this on my own fork before it was submitted to the repo here. opus 4.7 usually does good work for me but i want to look through the changes before moving back out of draft. i have some extra bandwidth over the summer and am looking forward to working within the project. cheers and have a good weekend.

paulirwin:
No problem, sounds good! When you get a chance, if you can give us an ETA on when you think it'll be ready to review. Looking forward to it, and thanks for the contribution!

vwilson:
I have, in fact, laid eyes upon every change, and I think we're good here. I suppose there is the possibility of a downstream `ToDictionary()` that neither the agents nor I caught.

src/Lucene.Net.TestFramework/Store/BaseDirectoryTestCase.cs
```diff
@@ -10,6 +10,7 @@
 using System.IO;
 using System.Threading;
 using Assert = Lucene.Net.TestFramework.Assert;
+using JCG = J2N.Collections.Generic;
```
paulirwin: Minor: if the only usage is in the comment below, maybe comment out this using as well if it's unused.

src/Lucene.Net.Tests.Analysis.SmartCn/Hhmm/TestBuildDictionary.cs
```diff
@@ -7,6 +7,7 @@
 using System;
 using System.IO;
 using Assert = Lucene.Net.TestFramework.Assert;
+using JCG = J2N.Collections.Generic;
```
paulirwin: Minor: same as earlier, if only usage is in comments, comment this out as well if unused.

NightOwl888 (approved):
Looks good.

I ran the tests a couple of times in Azure DevOps and there were no regressions. The tests have slowed down quite a bit in recent months, but any difference between `master` and this branch were within noise. I didn't run any benchmarks, though.

paulirwin (changes requested):
Overall looks good, thanks! Two minor things found if you could fix before merging.

paulirwin (approved):
Looks good, thanks!

PR #1289, opened 2026-05-02
[CI] Add ASF Allow list workflow

jbampton:
A composite GitHub Action that verifies all uses: refs in a project's workflow files are on the ASF Infrastructure approved allowlist:

https://github.com/apache/infrastructure-actions/blob/main/allowlist-check/README.md

Examples here:

https://github.com/apache/airflow/blob/main/.github/workflows/asf-allowlist-check.yml

https://github.com/apache/airflow-steward/blob/main/.github/workflows/asf-allowlist-check.yml


paulirwin (approved):
Very cool, thanks!

PR #1290, opened 2026-05-02
Update site copyright year to 2026 in docfx.json

jbampton:
There was 2024 and 2025 in two different places so made both 2026

### DIFF
--- a/websites/apidocs/docfx.json
+++ b/websites/apidocs/docfx.json
@@ -582,7 +582,7 @@
       "_appFaviconPath": "logo/favicon.ico",
       "_enableSearch": true,
       "_appLogoPath": "logo/lucene-net-color.png",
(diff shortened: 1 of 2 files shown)


NightOwl888:
Thanks. There is an issue with the copyright date, but the issue is that updating it shouldn't be a manual task, but an automated one. We already have automation to do that in `docs.ps1` at the time we build the docs site:

https://github.com/apache/lucenenet/blob/fa6b6cfe92fa5be74cf7cc8f17ba225cbadb06c3/websites/apidocs/docs.ps1#L110

However, the automation is missing from `site.ps1` (the main website):

https://github.com/apache/lucenenet/blob/fa6b6cfe92fa5be74cf7cc8f17ba225cbadb06c3/websites/site/site.ps1

So, in the former, all of that text is replaced and this date never gets deployed. Maybe it would make more sense to change the string to "auto-inserted" so it won't look like a problem that it is out of date to people who maintain the repo.

We have documentation on [how to build the website](https://lucenenet.apache.org/contributing/documentation.html) if you want to take a stab at updating the script to insert the footer (and current year) automatically so this is automated in both places.

paulirwin:
Converting this to draft since we need to address the changes @NightOwl888 mentioned above.

jbampton:
Just following up on this issue.  My Linux machine is no longer working so now I am using a Mac.

I used Homebrew and installed both the `dotnet` and `powershell` formula.

Got the website now running on localhost see the two images of my desktop.

Perhaps going forwards I can work on improving the website and docs.

Screenshot 2026-06-20 at 6 51 39 am
Screenshot 2026-06-20 at 6 51 45 am

PR #1292, opened 2026-05-03
Fix FSDirectory fsync race on output close

### LINKED ISSUE #894: Investigate Failing Test: Lucene.Net.Index.TestIndexWriterOnJRECrash::TestNRTThreads_Mem()

### Task description

This test was ported and added in #786 (to close #768).

Unfortunately, it fails intermittently. A user shared in [Lucene.net corrupted index (segments.gen)](https://stackoverflow.com/questions/77525791/lucene-net-corrupted-index-segments-gen) on StackOverflow that this appears to be due to a real problem that happens in production.

The failure may or may not be related to the deprecation of thread interrupts (#555) in Lucene.NET which were supported in the Java version. We make a best effort to support them using `UninterrruptableMonitor` instead of `lock` statements, but since a lock may be taken in any library we depend on that may throw `ThreadInterruptedException` from the action of taking a lock, we cannot 100% guarantee that we can catch every one of these exceptions in order to rollback an in-process commit. In Java, taking a lock does not throw an exception in any case.

vwilson:
Fix FSDirectory stale-file race and re-enable crash regression test

Fixes #894  

## Description

- Flush `FSIndexOutput`’s managed `FileStream` buffer before marking the file stale for `Sync()`.
- Add a regression test that observes the file length at stale-mark time to cover the race.
- Re-enable `TestIndexWriterOnJRECrash.TestNRTThreads_Mem` and make `CheckIndexes()` dispose the wrapper safely without leaking the directory.

## Testing
- Ran `Lucene.Net.Store.TestDirectory` in `src/Lucene.Net.Tests._J-S` successfully.
- Ran `TestIndexWriterOnJRECrash.TestNRTThreads_Mem` in `src/Lucene.Net.Tests._I-J` successfully, including the forked crash path.

This code generated by GPT 5.5 Pro

### DIFF
--- a/src/Lucene.Net/Store/FSDirectory.cs
+++ b/src/Lucene.Net/Store/FSDirectory.cs
@@ -510,6 +510,7 @@ namespace Lucene.Net.Store
             private readonly FileStream file;
 #pragma warning restore CA2213 // Disposable fields should be disposed
             private volatile bool isOpen; // remember if the file is open, so that we don't try to close it more than once
+            private int disposed; // LUCENENET specific: 0 = open, 1 = disposed. Used with Interlocked to make Dispose run-once.
             private readonly CRC32 crc = new CRC32();
 
             public FSIndexOutput(FSDirectory parent, string name)
@@ -574,45 +575,43 @@ namespace Lucene.Net.Store
             /// <inheritdoc/>
             protected override void Dispose(bool disposing)
             {
-                if (disposing && isOpen)
+                // LUCENENET specific: Ensure dispose only runs once, even if two threads race here.
+                // A plain isOpen check is not enough: the read is atomic, but two threads could both
+                // observe true and both run the dispose logic. Atomically transition disposed 0 -> 1
+                // so only the first thread proceeds.
+                if (disposing && 0 == Interlocked.CompareExchange(ref disposed, 1, 0))
                 {
                     Exception priorE = null; // LUCENENET: No need to cast to IOException
 
+                    UninterruptableMonitor.Enter(parent.m_syncLock);
                     try
                     {
-                        // LUCENENET: FileStream has a managed buffer. If we add this file
-                        // to m_staleFiles before flushing that buffer, another thread can
-                        // fsync the file through a separate handle, remove the stale marker,
-                        // and then this FileStream can flush later. That leaves bytes written
-                        // after the fsync and can make a committed index non-durable after a
-                        // process or machine crash.
+                        // LUCENENET: FileStream has a managed buffer, so a write is not guaranteed to
+                        // have reached the OS until the buffer is flushed. (Java's RandomAccessFile.write()
+                        // is unbuffered, so upstream has no such buffer and no flush race here.) If we add
+                        // this file to m_staleFiles (via OnIndexOutputClosed) before flushing that buffer,
+                        // another thread can fsync the file through a separate handle, remove the stale
+                        // marker, and only then does this FileStream flush its buffer to the OS. That leaves
+                        // bytes written after the fsync, which can make a committed index non-durable after
+                        // a process or machine crash.
                         //
-                        // Keep the FileStream flush and stale-file bookkeeping atomic with
-                        // FSDirectory.Sync() so a file is never considered synced until all
-                        // managed FileStream buffers have at least reached the OS.
-                        UninterruptableMonitor.Enter(parent.m_syncLock);
+                        // Keep the FileStream flush and stale-file bookkeeping atomic with FSDirectory.Sync()
+                        // (which also takes m_syncLock) so a file is never considered synced until all managed
+                        // FileStream buffers have at least reached the OS.
                         try
                         {
-                            try
-                            {
-                                file.Flush(flushToDisk: false);
-                            }
-                            catch (Exception ioe) when (ioe.IsIOException())
-                            {
-                                priorE = ioe;
-                            }
-                            finally
-                            {
-                                parent.OnIndexOutputClosed(this);
-                            }
+                            file.Flush(flushToDisk: false);
                         }
-                        finally
+                        catch (Exception ioe) when (ioe.IsIOException())
                         {
-                            UninterruptableMonitor.Exit(parent.m_syncLock);
+                            priorE = ioe;
                         }
+
+                        parent.OnIndexOutputClosed(this);
                     }
                     finally
                     {
+                        UninterruptableMonitor.Exit(parent.m_syncLock);
                         isOpen = false;
                         IOUtils.DisposeWhileHandlingException(priorE, file);
                     }
--- a/src/Lucene.Net/Store/FSDirectory.cs
+++ b/src/Lucene.Net/Store/FSDirectory.cs
@@ -574,26 +574,47 @@ namespace Lucene.Net.Store
             /// <inheritdoc/>
             protected override void Dispose(bool disposing)
             {
-                if (disposing)
+                if (disposing && isOpen)
                 {
-                    parent.OnIndexOutputClosed(this);
-                    // only close the file if it has not been closed yet
-                    if (isOpen)
+                    Exception priorE = null; // LUCENENET: No need to cast to IOException
+
+                    try
                     {
-                        Exception priorE = null; // LUCENENET: No need to cast to IOException
+                        // LUCENENET: FileStream has a managed buffer. If we add this file
+                        // to m_staleFiles before flushing that buffer, another thread can
+                        // fsync the file through a separate handle, remove the stale marker,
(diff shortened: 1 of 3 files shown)


src/Lucene.Net/Store/FSDirectory.cs
```diff
@@ -574,27 +574,48 @@ public override void Flush()
...
             {
-                if (disposing)
+                if (disposing && isOpen)
                 {
-                    parent.OnIndexOutputClosed(this);
-                    // only close the file if it has not been closed yet
-                    if (isOpen)
+                    Exception priorE = null; // LUCENENET: No need to cast to IOException
+
+                    try
```
paulirwin, line 587: The three levels of try here are likely overkill, we should be able to collapse them to two (and add back your helpful comments):

```csharp
if (disposing && isOpen)
{
    Exception priorE = null;
    UninterruptableMonitor.Enter(parent.m_syncLock);
    try
    {
        try 
        {
            file.Flush(flushToDisk: false); 
        }
        catch (Exception ioe) when (ioe.IsIOException())
        {
            priorE = ioe;
        }

        parent.OnIndexOutputClosed(this);
    }
    finally
    {
        UninterruptableMonitor.Exit(parent.m_syncLock);
        isOpen = false;
        IOUtils.DisposeWhileHandlingException(priorE, file);
    }
}
```

Aside: it's arguable whether or not setting isOpen to false should be inside the lock or not. On the one hand, it is volatile and not one of the fields that is considered for the lock elsewhere. But, if we moved setting it inside the lock (up one line in my code above) and did a double-check inside the first try, we could possibly avoid some unnecessary extra work. Not sure if it's worth fixing, but just noting for consideration.

src/Lucene.Net/Store/FSDirectory.cs
```diff
@@ -574,27 +574,48 @@ public override void Flush()
...
                     {
-                        Exception priorE = null; // LUCENENET: No need to cast to IOException
+                        // LUCENENET: FileStream has a managed buffer. If we add this file
```
paulirwin: We should call out in this comment _why_ Java doesn't have this problem. As far as I can tell, it seems to be because `RandomAccessFile.write()` does not buffer like FileStream does, so there is no flush race.

paulirwin (changes requested):
Thanks for this PR! It's awesome to see this issue fixed. Some minor things noted below.

src/Lucene.Net/Store/FSDirectory.cs
```diff
@@ -574,27 +574,48 @@ public override void Flush()
             /// <inheritdoc/>
             protected override void Dispose(bool disposing)
             {
-                if (disposing)
+                if (disposing && isOpen)
```
NightOwl888: While the read of a boolean is atomic, there is still a chance that 2 threads could make in here. Was that intentional?

Note that in the `FSDirectory` subclasses, we do this in the dispose block to ensure dispose only happens by the first thread and all others are ignored.

```c#
if (0 != Interlocked.CompareExchange(ref this.disposed, 1, 0)) return;
```

Then for the `IsOpen` checks, we do the check atomically.

```c#
        // LUCENENET specific - ensure we can't be abused after dispose
        private bool IsOpen => Interlocked.CompareExchange(ref this.disposed, 0, 0) == 0 ? true : false;

        // LUCENENET specific - ensure we can't be abused after dispose
        private void EnsureOpen()
        {
            if (!IsOpen)
            {
                throw AlreadyClosedException.Create(this.GetType().FullName, "this IndexOutput is disposed.");
            }
        }
```

Not sure this is contributing to the problem at hand, but it is usually better not to make assumptions that running the `Dispose()` logic more than once is supposed to be safe to do.

paulirwin:
@vwilson I pushed an update to your PR. The now-running TestNRTThreads_Mem test was failing on arm64, which was never noticed before since it was disabled with AwaitsFix. The architecture detection was hard-coded to only support x64 and x86, so this adds ARM architecture support. The tests pass locally for me on macOS arm64 now.

NightOwl888:
I am running the complete barrage of tests on Azure DevOps (30 complete runs). That will shake out any issues if there are any. It will take a few hours to complete all of those runs, though.

So far, there is one test failure on this test:

https://dev.azure.com/lucene-net-temp4/Main/_build/results?buildId=353&view=ms.vss-test-web.build-test-results-tab

But, it happened on `net10.0`, and we are always getting random failures on `net10.0`. So, I cannot say for sure this broke this particular test in any way that `net10.0` isn't already broken.

I also added `[Repeat(50)]` attribute locally and ran it twice (10 minutes each time), and it was green both times. This is definitely an improvement.

paulirwin:
@NightOwl888 Oof, a check index failed error. That's potentially bad. I haven't seen that on net10 yet.

Are you able to get any more details on that one? The message leaves a bit to be desired.

NightOwl888:
After all of these runs are done, I will open an issue and include some more details. `net10.0` is like a shooting gallery. There are failures that randomly occur in multiple tests.

But this specific test that is being fixed in this issue is specifically running CheckIndex to check the result and intentionally trying to stress the scenario to force it to write a broken index if something is not right.

So far, most of the failures are on x86, on `net9.0` and `net10.0`. But there was one on `net48` on `x64` that I am downloading the logs for - the runner indicates a crash in `Lucene.Net.Tests._I-J`.

NightOwl888:
Well, the results are in. Unfortunately, this didn't seem to be the entire cause. However, it may indeed have been one of the contributing factors.

## Results

You can see the runs with the title `Fix FSDirectory fsync race on output close` here. I did 6 runs on 5 different projects.

https://dev.azure.com/shad0962/Experiments/_build?definitionId=2&_a=summary
https://dev.azure.com/LuceneNET-Temp/Lucene.NET/_build?definitionId=1&_a=summary
https://dev.azure.com/lucene-net-temp2/Lucene.NET/_build?definitionId=1&_a=summary
https://dev.azure.com/lucene-net-temp3/Lucene.NET/_build?definitionId=1&_a=summary
https://dev.azure.com/lucene-net-temp4/Main/_build?definitionId=1&_a=summary

There are some irrelevant failures of `TestRandomStrings` (fixed in #122) and anything that failed on `net10.0` is suspect because we have been having concurrency problems since the target was added (see #1295). That doesn't leave so many failures, but unfortunately this didn't eliminate them all.

----------------------------------------

There seems to be some rot with this test that didn't age too well, also. I am seeing freezes when trying to run the tests on Windows x64 and selecting x86 as the target platform. I stepped through and the correct value gets built into the `dotnet test` command, but the problem is with the SDK itself.

Since .NET 8, we can no longer assume that when calling `dotnet` on 64-bit Windows that an x86 call will work if only the 64-bit SDK is installed. Now, the 32-bit SDK must be installed separately for the test to function when running the parent test as `x86`. While I think @paulirwin made the right call by choosing the current runtime platform to pass on to the forked test, this detail means it is not guaranteed to run on that platform.

My thought is to attempt to detect whether the SDK is installed and either make the test inconclusive or do a hard failure to indicate that the SDK must be installed. But maybe there is another way to deal with this. Good test design usually means tests should be ready to run just by cloning the repo. However, now that this SDK behavior has changed, we have inadvertently inherited a dependency on an SDK that wasn't required before.

One other thing we should address is to do a hard failure if the forked process has a non-zero exit code and report the STDERR message in the test. That would prevent a "forever wait" state that we are seeing when the platform is specified wrong as well as give us better info to investigate the failure.

This test has a lot of moving parts, and it is hard to get it to a point where it is consistently stable to run.

paulirwin:
Agreed about needing to hard-fail on non-zero exit code. I didn't put those pieces together but the test was hanging before I added the arm64 fix.

Also, can we leave off the specified architecture if it's not x86? I think that's the only case we have to worry about for the test runs.

(Note that technically .NET supports 32-bit ARM as well, but we don't currently run tests on it. Not sure how practical that is. It basically only exists for Raspberry-Pi-like device support.)

It might be worth seeing if there's some inspiration we could gain from how BenchmarkDotNet does this, too.

paulirwin:
I'm going to convert this to draft until we're sure we're stable here.

src/Lucene.Net/Store/FSDirectory.cs
paulirwin: @vwilson I pushed an update to your PR to fix this, please review and let me know if it looks good. Thanks!

src/Lucene.Net/Store/FSDirectory.cs
paulirwin: @NightOwl888 I pushed an update to this PR branch to fix this, please review and let me know if it looks good. Thanks!

paulirwin (approved):

PR #1299, opened 2026-05-12
Add Infra-recommended branch protection rules

paulirwin:

Adopt Apache Infrastructure recommended branch protection rules

## Description

Apache Infra recommends adopting blocking forced pushes and blocking deletions of protected branches. This adopts those, along with requiring linear history and comment resolution on PRs (which seemed like low-friction additions, and things we already do in practice).

This supersedes the legacy branch protection rule with the more modern rulesets approach.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -35,6 +35,15 @@ github:
     # Enable discussions
     discussions: true
 
-  # Disallow forced pushes
-  protected_branches:
-    master: {}
+  rulesets:
+    - name: "Branch Protection"
+      type: branch
+      branches:
+        includes:
+          - "master"
+          - "release/*"
+        excludes: []
+      restrict_deletion: true
+      restrict_force_push: true
+      required_linear_history: true
+      required_conversation_resolution: true
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -42,6 +42,7 @@ github:
         includes:
           - "master"
           - "release/*"
+          - "docs/*"
         excludes: []
       restrict_deletion: true
       restrict_force_push: true


.asf.yaml
```diff
@@ -35,6 +35,15 @@ github:
...
-  # Disallow forced pushes
-  protected_branches:
-    master: {}
+  rulesets:
+    - name: "Branch Protection"
+      type: branch
+      branches:
+        includes:
+          - "master"
+          - "release/*"
```
NightOwl888, line 44: Currently, we don't have a release branching scheme (although if we did, this is probably how we would do it).

That being said, we should disallow force pushes to the `docs/*` branches. Those are to track the API state at the time of release so we can correct and regenerate docs. I typically have been adding docs changes as separate commits, cherry picking them to the release branch, then regenerating the docs. However, I don't think we have an official policy for this, so I am sure many docs changes that should be patches on that branch end up waiting until the next release to be published.

Direct pushing should be allowed, but force pushing is a bit of a hazard for `docs/*` branches because it could wipe out that save point and we could lose all of the docs updates that happened since the release. The best we could do is restore the docs state to where it was at the original release commit.

So, please add `docs/*` here.

NightOwl888 (changes requested):
Looks good to merge after one addition. See the inline comment.

paulirwin:
Not sure what I did wrong here, but getting this error:

> An error occurred while processing the github feature in .asf.yaml:
> 
> Validation failed while creating ruleset 'Branch Protection': Invalid request.
> 
> Invalid property /rules/3: data matches no possible input. See `documentation_url`.

PR #1308, opened 2026-05-19
misc: fix typos

jbampton:
General clean up


.build/psake/en-US/psake.psm1-help.xml.old
```diff
@@ -19,35 +19,35 @@
     <command:syntax>
       <command:syntaxItem>
         <command:name>Invoke-psake</command:name>
-        <command:parameter require="false" variableLength="false" globbing="false" pipelineInput="false" postion="0">
+        <command:parameter require="false" variableLength="false" globbing="false" pipelineInput="false" position="0">
```
paulirwin: @NightOwl888 what is this file? Can it be removed? I did a git grep and nothing seems to reference it.

paulirwin (changes requested):
Looks good apart from this psake file.

.github/linters/codespell.txt
```diff
@@ -74,7 +74,6 @@ compatibity
 compatiblity
 completly
 concret
-condtions
```
paulirwin, line 77: @jbampton Question: is it possible to have codespell check this list for any suppressions that are no longer needed, to help prune the list? It's possible along the way that we might fix typos without updating this suppression list.

.build/psake/en-US/psake.psm1-help.xml.old
NightOwl888: PSake is a 3rd party module. 

https://github.com/psake/psake

Our build scripts (for all repositories) use PSake as a task manager. It helps divide the build into steps and control dependencies between steps.

Although we don't reference all of it, this is part of the module code. I am not sure we should be changing xml elements even though they are spelled wrong.

When I was prototyping a build script methodology, installing Powershell modules was brittle and requires elevated permissions and/or dependencies such as Chololatey. So, it was simpler just to make a local copy of the module and invoke it directly than to install it on the OS as part of the CI operation. I suppose we could take another look at what it takes to install this reliably across user machines.

Do note that it is not part of CI in any repository but lucenenet - it is mainly used to support the `build` command. It would be a bad UX if our `build` command couldn't run even though the user has all of the documented prerequisites.

Also note that it is an Apache requirement to be able to [build](https://www.apache.org/legal/release-policy.html#source-packages) and test on the command line using the appropriate tools.

-------------------------

Anyway, IMO we should probably add an ignore rule for spell checking the PSake module rather than modifying 3rd party materials because we don't know specifically if anything will stop functioning by changing these. We can either upgrade and/or submit a PR to fix this upstream to resolve the spelling errors in our repositories.

paulirwin: This file was removed upstream last July: https://github.com/psake/psake/commit/0eed34f9321efe68ad2b4619354f439b56a1f146#diff-44b3b2dd62b1b50187418853b3e98bc836cdb1fe45ce9c30b5285440cd7f4e05

I confirmed it is not referenced anywhere. We can remove it safely. It was called ".old" even in the version we ported.

Edit: by "confirmed" I include that to mean that I also tested the build script locally.

paulirwin: I went ahead and removed this unused file by updating this PR.

.github/linters/codespell.txt
paulirwin: To answer my own question, it is not possible to have codespell check for unused suppressions currently. I had Claude Code help me write a script to prune any unused suppressions. It found only one: `lien` (which is strange, since that is a word). I pushed that script to this PR as a new file in the same folder.

PR #1317, opened 2026-06-08
Update copyright year in NOTICE.txt

jbampton:
Change end year from 2024 to 2026

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1,5 +1,5 @@
 Apache Lucene.Net
-Copyright 2006-2024 The Apache Software Foundation
+Copyright 2006-2026 The Apache Software Foundation
 
 This product includes software developed at
 The Apache Software Foundation (http://www.apache.org/).


paulirwin (changes requested):
Thanks for tackling small things like this, but can you please include this change in the #1290 PR, so that it will be one related commit when squashed (easier to do things like cherry-pick later), and also address @NightOwl888's feedback on that PR so we can get that merged? Thanks!

NightOwl888:
Actually, since we already have this as a [documented step in the release procedure](https://lucenenet.apache.org/contributing/make-release.html#release-steps) we don't necessarily need to prioritize automating this step. But we do need this to be cherry picked onto the release/4.8.0-beta00018 branch after it is merged here. So, we should merge this now.

I would categorize "licensing" and "docs" as 2 different things that should be kept as separate commits being that we never need to update the license files as part of a docs or website update. But we will always need to update the license files prior to release if the year has flipped.

paulirwin (approved):

PR #1318, opened 2026-06-08
[CI] Switch pre-commit to use ASF approved prek-action

jbampton:
Changes to use the ASF approved prek-action:

https://github.com/apache/infrastructure-actions/blob/437608c5893288af177de21bb7e7b735a548ee26/approved_patterns.yml#L213

https://prek.j178.dev/

refs https://github.com/apache/shiro/pull/2705

https://prek.j178.dev/#who-is-using-prek

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -15,16 +15,14 @@
 # specific language governing permissions and limitations
 # under the License.
 
+---
+# https://github.com/j178/prek
 name: pre-commit
 
 on: [pull_request]
 
 permissions:
   contents: read
-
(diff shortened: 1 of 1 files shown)


paulirwin (approved):
This looks good, approved. I like this for several reasons:

1. Removes a lot of noise from the workflow file
2. Doesn't require installing Python
3. Uses an ASF-approved action, pinned to a SHA
4. It's already in use at another Apache project
5. Works well with our existing pre-commit rules
6. ... and WOW is it fast! Running prek locally is anywhere from 1.5-2.6x faster in initial testing.

Thanks for doing this!

PR #1319, opened 2026-06-08
`.asf.yaml`: use a better repository description

jbampton:
We can update the repo description to be more informative. We should use this small space or real estate as best we can.

The next image is from my GitHub profile the pinned repo sections showing the current repo descriptions:

Screenshot 2026-06-09 at 2 34 57 am

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -10,7 +10,7 @@ notifications:
   # jobs: builds@lucenenet.apache.org # This is probably too noisy to enable
 
 github:
-  description: "Apache Lucene.NET is an open-source full-text search library written in C#. It is a port of the popular Java Apache Lucene project."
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -10,7 +10,7 @@ notifications:
   # jobs: builds@lucenenet.apache.org # This is probably too noisy to enable
 
 github:
-  description: "Apache Lucene.NET"
+  description: "Apache Lucene.NET is an open-source full-text search library written in C#. It is a port of the popular Java Apache Lucene project."
```
paulirwin: I know this was originally copied from our README, but IMHO I think we can make it a little more concise for the description:

```suggestion
  description: "Apache Lucene.NET is an open-source full-text search library written in C#, ported from the Apache Lucene project."
```

paulirwin (commented):
Suggested change for consideration.

PR #1326, opened 2026-06-09
[CI] Pin all actions to latest sha

jbampton:
https://github.com/actions/checkout/commit/df4cb1c069e1874edd31b4311f1884172cec0e10

https://github.com/actions/cache/commit/27d5ce7f107fe9357f9df03efb73ab90386fccae

https://github.com/actions/setup-dotnet/commit/9a946fdbd5fb07b82b2f5a4466058b876ab72bb2

https://github.com/actions/upload-artifact/commit/043fb46d1a93c77aae656e7c1c64a875d1fc6a0a

Using mutable tags like `@v4` leaves your pipelines vulnerable to supply chain attacks if a developer's account is compromised. Pinning to a unique 40-character commit SHA ensures you run immutable, unalterable code that cannot be silently modified by bad actors. This practice guarantees absolute build reproducibility because cryptographic hashes cannot be force-pushed or rewritten. It also protects your workflows from risks like repository deletion or malicious name squatting. For maximum safety, pairs SHAs with inline comments so dependency tools can still automate your version updates.

### DIFF
--- a/.github/workflows/EditorConfig-Rules-Check.yml
+++ b/.github/workflows/EditorConfig-Rules-Check.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       - name: Install EditorConfig Checker
         run: |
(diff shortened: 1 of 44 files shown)


paulirwin:
@jbampton I'll fix the comments and push to this PR, Claude Code had already found those as well.

paulirwin (approved):
Looks good, added a few more, and confirmed that Generate-TestWorkflows.ps1 results in no diffs.

PR #1320, opened 2026-06-08
[CI] Add pre-commit hook gitleaks secret scanner

jbampton:
https://gitleaks.io

https://github.com/gitleaks/gitleaks#pre-commit

Used on Apache Sedona:

https://github.com/apache/sedona/blob/ffefe5b5140e9770e9fc0c8c0d3cafc113d7398f/.pre-commit-config.yaml#L365

PR #1328, opened 2026-06-10
[CI] Add official pre-commit hook `detect-aws-credentials`

jbampton:
https://github.com/pre-commit/pre-commit-hooks#detect-aws-credentials


paulirwin (approved):
Just like with #1320, I don't think this really applies in the normal course of contributions to our project (i.e. we don't use AWS at all with this project), but (a) it can help prevent contributors from accidentally committing their own secrets, and (b) it's built in, so nothing else to install. I think this is fine to add.

PR #1339, opened 2026-06-11
Implement IBufferWriter<char> in OpenStringBuilder and CharTermAttribute, #1315

### LINKED ISSUE #1315: Implement `IBufferWriter<char>` on `ICharTermAttribute`/`CharTermAttribute` and `OpenStringBuilder`

### Task description

https://github.com/apache/lucenenet/issues/1293#issuecomment-4462036089

The BCL `IBufferWriter<T>` interface is purpose-built for allowing arbitrary code to grow and write to a buffer when the source has either a known or unknown buffer length. This allows for scenarios with complex business logic that would be wasteful to repeat if the buffer passed in is not already large enough, without the baggage of `System.Text.StringBuilder`, which cannot even be converted to a `ReadOnlySpan<char>` without doing an extra allocation.

`IAppendable` is an interface from Java that performs a similar operation and `IBufferWriter<T>` has the potential to eliminate it from the Lucene.NET codebase. However, for now we should limit this addition to `CharTermAttribute` and the `Normalizer2` in ICU4N which will give us a cross-library proof of concept.

The `IBufferWriter<T>` interface should be inherited directly by `ICharTermAttribute`, which is a breaking API change. However, due to the limited use cases of `IAttribute`-derived types, this is unlikely to affect many end users. The implementation of the 3 members will then need to be on the `CharTermAttribute` class.

The implementation should use the existing growth semantics provided by `InternalResizeBuffer`. See the `Append(ReadOnlySpan<char>)` implementation for an example. 

See the [`ArrayBufferWriter`](https://github.com/dotnet/runtime/blob/v10.0.6/src/libraries/Common/src/System/Buffers/ArrayBufferWriter.cs) and [its tests](https://github.com/dotnet/runtime/tree/v10.0.6/src/libraries/System.Memory/tests/ArrayBufferWriter) to see an optimized implementation example along with some tests we can use to verify the implementation.

paulirwin:

Implement `IBufferWriter<char>` in OpenStringBuilder and CharTermAttribute.

Fixes #1315

## Description

This PR does the following:
1. Implements `IBufferWriter<char>` in OpenStringBuilder. This uses the existing `EnsureCapacity` for the buffer-growing logic. A new `FreeCapacity` property was added to help with unit testing as well as consumers of this new API.
2. Adds `AsSpan()` and `AsMemory()` to OpenStringBuilder. This is not only useful for unit testing, but can open up some other performance improvement opportunities by reducing allocations. (Note that `ICharTermAttribute` already has `MemoryExtensions` for this.)
3. Updates `ICUNormalizer2Filter` to use the simpler `AsSpan()` calls on the attribute, and to use the span-based `Append` overload to optimize appending to the attribute.
4. Adds `IBufferWriter<char>` to `ICharTermAttribute`. **Breaking change.**
5. Adds an `Append(ReadOnlySpan<char>)` overload to `ICharTermAttribute`. **Breaking change.** This is in lieu of adding `ISpanAppendable`, per the discussion in #1315.
6. Implements `IBufferWriter<char>` in `CharTermAttribute`. This uses the existing `InternalResizeBuffer` for the buffer-growing logic. New `FreeCapacity` and `Capacity` properties were added to help with unit testing as well as consumers of this new API.
7. `CharTermAttribute` now explicitly implements the new `Append(ReadOnlySpan<char>)` overload by delegating to its existing implementation.
8. Full unit tests (where applicable) were added, ported from .NET's ArrayBufferWriter and adapted to these two types.

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
@@ -5,6 +5,8 @@ using System;
 using System.Buffers;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using System.Text;
 using WritableArrayAttribute = Lucene.Net.Support.WritableArrayAttribute;
 
@@ -426,9 +428,9 @@ namespace Lucene.Net.Analysis.Util
         public void Advance(int count)
         {
             if (count < 0)
-                throw new ArgumentException(null, nameof(count));
+                throw new ArgumentOutOfRangeException(nameof(count));
 
-            if (m_len > m_buf.Length - count)
+            if (count > m_buf.Length - m_len)
                 throw new InvalidOperationException($"Cannot advance {count} characters beyond the end of the buffer.");
 
             m_len += count;
@@ -452,7 +454,7 @@ namespace Lucene.Net.Analysis.Util
         {
             if (sizeHint < 0)
             {
-                throw new ArgumentException("sizeHint must be non-negative", nameof(sizeHint));
+                throw new ArgumentOutOfRangeException(nameof(sizeHint), "sizeHint must be non-negative");
             }
 
             if (sizeHint == 0)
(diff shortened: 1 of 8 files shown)


src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
```diff
@@ -400,5 +416,72 @@ IAppendable IAppendable.Append(ICharSequence value, int startIndex, int count)
...
+        private void CheckAndResizeBufferForBufferWriter(int sizeHint)
+        {
+            if (sizeHint < 0)
+            {
+                throw new ArgumentException("sizeHint must be non-negative", nameof(sizeHint));
+            }
+
+            if (sizeHint == 0)
+            {
+                sizeHint = 1;
```
NightOwl888: Since the default constructor uses an initial size of 32, we should probably provide something more reasonable here. Both the Java and .NET StringBuilder types have a default value of 16. That is generally enough to write most numbers or at least one long word.

src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
```diff
@@ -400,5 +416,72 @@ IAppendable IAppendable.Append(ICharSequence value, int startIndex, int count)
...
+
+        /// <summary>
+        /// Returns the data written to the underlying buffer so far, as a <see cref="ReadOnlyMemory{T}"/>.
+        /// </summary>
+        public ReadOnlyMemory<char> AsMemory() => m_buf.AsMemory(0, m_len);
+
+        /// <summary>
+        /// Returns the data written to the underlying buffer so far, as a <see cref="ReadOnlySpan{T}"/>.
+        /// </summary>
+        public ReadOnlySpan<char> AsSpan() => m_buf.AsSpan(0, m_len);
```
NightOwl888: One thing I discovered while designing `TextBuilder` is that it is more sensible to implement `AsSpan()` as extension methods on reference types. Why? Because reference types can be `null` and `null` naturally converts to `ReadOnlySpan<char>.Empty`. If we implement it here, it will throw a `NullReferenceException` when the value is `null`

Another thing to note is that it would be most sensible to put these extension methods into the same namespace `Lucene.Net.Analysis.Util`. In this case, I would suggest simply dropping an `OpenStringBuilderExtensions` class right in this file.

Note also that there are 6 highly recommended overloads of these methods:

`public static ReadOnlySpan<char> AsSpan(this OpenStringBuilder? text)`
`public static ReadOnlySpan<char> AsSpan(this OpenStringBuilder? text, int start)`
`public static ReadOnlySpan<char> AsSpan(this OpenStringBuilder? text, int start, int length)`

`public static ReadOnlyMemory<char> AsMemory(this OpenStringBuilder? text)`
`public static ReadOnlyMemory<char> AsMemory(this OpenStringBuilder? text, int start)`
`public static ReadOnlyMemory<char> AsMemory(this OpenStringBuilder? text, int start, int length)`

There are also 4 additional overloads that are a nice finishing touch:

`public static ReadOnlySpan<char> AsSpan(this OpenStringBuilder? text, int start, int length)`
`public static ReadOnlySpan<char> AsSpan(this OpenStringBuilder? text, System.Index startIndex)`

`public static ReadOnlyMemory<char> AsMemory(this OpenStringBuilder? text, System.Index startIndex)`
`public static ReadOnlyMemory<char> AsMemory(this OpenStringBuilder? text, Range range)`

The implementation in `Lucene.Net.MemoryExtensions` is a direct adaption of the original BCL implementation and preserves the optimizations on .NET Core. The tests for them were also derived directly from the BCL.

src/Lucene.Net.Tests.Analysis.Common/Support/Analysis/Util/TestOpenStringBuilderBufferWriter.cs
```diff
@@ -0,0 +1,488 @@
+// Adapted from MIT-licensed ArrayBufferWriter tests here:
```
NightOwl888, line 1: Please move this file into a top-level `Support` folder, so it is in `Support/Analysis/Util/`. But leave the namespace `Lucene.Net.Analysis.Util`.

src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
```diff
@@ -400,5 +416,72 @@ IAppendable IAppendable.Append(ICharSequence value, int startIndex, int count)
...
+        #region IBufferWriter<char> members and support
+
+        // LUCENENET-specific: IBufferWriter<char> support.
+        // See ArrayBufferWriter for an inspiration and reference implementation:
+        // https://github.com/dotnet/runtime/blob/v10.0.6/src/libraries/Common/src/System/Buffers/ArrayBufferWriter.cs
+
+        public void Advance(int count)
+        {
+            if (count < 0)
+                throw new ArgumentException(null, nameof(count));
```
NightOwl888: Let's throw `ArgumentOutOfRangeException` here.

src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
```diff
@@ -400,5 +416,72 @@ IAppendable IAppendable.Append(ICharSequence value, int startIndex, int count)
...
+
+            if (m_len > m_buf.Length - count)
```
NightOwl888: I think the intent would be clearer if the `count` and `m_len` variables were reversed here.

```c#
if (count > m_buf.Lenth - m_len)
```

src/Lucene.Net.Tests/Support/Analysis/TokenAttributes/TestCharTermAttributeImplBufferWriter.cs
```diff
@@ -0,0 +1,445 @@
+// Adapted from MIT-licensed ArrayBufferWriter tests here:
```
NightOwl888, line 1: Please move this file into a top-level `Support` folder, so it is in `Support/Analysis/TokenAttributes/`. But leave the namespace `Lucene.Net.Analysis.TokenAttributes`.

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttributeImpl.cs
```diff
@@ -516,5 +529,121 @@ public override void CopyTo(IAttribute target) // LUCENENET specific - intention
...
+        /// available buffer space.
+        /// </exception>
+        /// <remarks>
+        /// You must request a new buffer after calling Advance to continue writing more data
+        /// and cannot write to a previously acquired buffer.
+        /// </remarks>
+        public void Advance(int count)
+        {
+            if (count < 0)
+                throw new ArgumentException(null, nameof(count));
```
NightOwl888: Please throw `ArgumentOutOfRangeException` here.

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttributeImpl.cs
```diff
@@ -516,5 +529,121 @@ public override void CopyTo(IAttribute target) // LUCENENET specific - intention
...
+
+            if (termLength > termBuffer.Length - count)
```
NightOwl888: As in the other implementation, I think the intent is clearer by putting count on the left side.

```c#
if (count > termBuffer.Length - termLength)
```

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttributeImpl.cs
```diff
@@ -516,5 +529,121 @@ public override void CopyTo(IAttribute target) // LUCENENET specific - intention
...
+            CheckAndResizeBufferForBufferWriter(sizeHint);
+            Debug.Assert(termBuffer.Length > termLength);
+            return termBuffer.AsSpan(termLength);
+        }
+
+        private void CheckAndResizeBufferForBufferWriter(int sizeHint)
+        {
+            if (sizeHint < 0)
+            {
+                throw new ArgumentException("sizeHint must be non-negative", nameof(sizeHint));
```
NightOwl888: Please throw `ArgumentOutOfRangeException` here.

src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
```diff
@@ -400,5 +416,72 @@ IAppendable IAppendable.Append(ICharSequence value, int startIndex, int count)
...
+            CheckAndResizeBufferForBufferWriter(sizeHint);
+            Debug.Assert(m_buf.Length > m_len);
+            return m_buf.AsSpan(m_len);
+        }
+
+        private void CheckAndResizeBufferForBufferWriter(int sizeHint)
+        {
+            if (sizeHint < 0)
+            {
+                throw new ArgumentException("sizeHint must be non-negative", nameof(sizeHint));
```
NightOwl888: Please throw `ArgumentOutOfRangeException` here.

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttributeImpl.cs
```diff
@@ -516,5 +529,121 @@ public override void CopyTo(IAttribute target) // LUCENENET specific - intention
...
+            }
+
+            if (sizeHint == 0)
+            {
+                sizeHint = 1;
```
NightOwl888: In this case we have a `MIN_BUFFER_SIZE` that we can use here as a logical default size.

NightOwl888 (changes requested):
Thanks for the PR.

I noticed that a couple of the tests failed that will need to be looked into. There are also a few changes I think we should make. See the inline comments.

src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
paulirwin: FYI - this behavior is what ArrayBufferWriter does: https://github.com/dotnet/runtime/blob/7706f546bac1a99b3d891afe3591dc88c67f0cc4/src/libraries/Common/src/System/Buffers/ArrayBufferWriter.cs#L206

Note that the only time you actually would get a span of length 1 (when `sizeHint == 0` or 1) is if there is only 1 char free capacity. If you don't pass in a size hint and the buffer has already grown greater than 1 (say, it has 500 chars free capacity) then you'll get back a span of length equal to FreeCapacity (i.e. 500), not 1, even though sizeHint is set to 1 in this branch. The 1-length case is probably pretty rare; and in that case, the caller has already said that they are fine if they get a span of length 1 by not providing a size hint.

Also note that if there is no more free capacity, this sizeHint of 1 will result in a doubling of the buffer, and returning a span of the prior buffer length (because it was doubled).

Example typical flow:
1. OpenStringBuilder is created with default capacity of 32
2. `GetSpan()` called with no sizeHint, returns a span of length 32
3. `Advance(24)` called to note 24 chars written
4. `GetSpan()` called again with no sizeHint, returns a span of length 8 (32 - 24)
5. Caller can tolerate an 8-char buffer because they gave no size hint. `Advance(8)` called to max out the written chars.
6. `GetSpan()` called again with no sizeHint, now there is no free capacity.
7. Internally, `EnsureCapacity` called with a sizeHint fallback of 1. 32 + 1 > 32, so `Resize(33)` is called.
8. Internally, `m_buf.Length << 1` (doubled) results in 64 which is bigger than 33, so the buffer resizes to 64.
9. Back to `GetSpan()`, it now returns a buffer of size 32 (64 capacity - 32 written)

Note that you can construct an example flow with a real span of size 1 returned above by replacing the number 24 with 31 in step 3, and 8 with 1 in steps 4-5. Then step 4 would return a span of length 1, but the caller has already said they don't care by not providing an argument.

In my opinion, we don't need to change this, as it already is memory-optimized to not unnecessarily increase the buffer size if the caller can tolerate short spans of the remaining capacity and doesn't care about what minimum size span they get back. Going to resolve this one, but feel free to reopen if you think there's a good alternative justification.

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttributeImpl.cs
paulirwin: See other comment.

paulirwin:
FYI - the failing tests on .NET Framework were surprisingly due to a JIT bug. I extracted a minimal reproduction and submitted the bug report to Microsoft. I worked around them by adding NoInlining methods around those calls which avoids the bug.

Feedback addressed or responded to; ready for re-review.

src/Lucene.Net.Tests.Analysis.Common/Support/Analysis/Util/TestOpenStringBuilderBufferWriter.cs
```diff
@@ -0,0 +1,507 @@
...
+            {
+                pinnedArray.Free();
+            }
+        }
+
+        protected static void WriteData(IBufferWriter<char> bufferWriter, int numChars)
+        {
+            Span<char> outputSpan = bufferWriter.GetSpan(numChars);
+            Debug.Assert(outputSpan.Length >= numChars);
+            var random = new Random(42);
```
paulirwin: Need to remove static from this method and use the Random property from the base class for repeatable runs.

src/Lucene.Net.Tests/Support/Analysis/TokenAttributes/TestCharTermAttributeImplBufferWriter.cs
```diff
@@ -0,0 +1,464 @@
...
+            {
+                pinnedArray.Free();
+            }
+        }
+
+        protected static void WriteData(IBufferWriter<char> bufferWriter, int numChars)
+        {
+            Span<char> outputSpan = bufferWriter.GetSpan(numChars);
+            Debug.Assert(outputSpan.Length >= numChars);
+            var random = new Random(42);
```
paulirwin: See other comment about Random use.

paulirwin (commented):
Need to update this to use the repeatable Random property.

src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
NightOwl888: Respectfully, I disagree. The `ArrayBufferWriter<T>` is a general-purpose buffer. `CharTermAttribute` and `OpenStringBuilder` are specifically designed to buffer text. It is reasonable to expect the "don't care" size to at least hold an average word. 1 `char` isn't even enough space to hold every character, since some of them require 2 chars.

`CharTermAttribute` uses a conservative value of 10 for the default size in the constructor. This seems like a good default size for that case. For `OpenStringBuilder`, I would probably choose 16 to mirror the `StringBuilder` constructors of both .NET and Java, even if this implementation uses a more aggressive 32.

Thanks for pointing out that the value returned is typically the whole remaining capacity. I had to double-check `MutableTextBuffer` is doing the same. Note that we may end up either replacing this implementation everywhere with `TextBuilder` or changing it to be a facade around the `MutableTextBuffer` implementation, anyway.

src/Lucene.Net.Tests/Support/Analysis/TokenAttributes/TestCharTermAttributeImplBufferWriter.cs
```diff
@@ -0,0 +1,463 @@
...
+        // chars. If the allocation fails with OutOfMemoryException, the test is a no-op. Marked
+        // [Slow] due to the large allocation.
+        //
+        // NOTE: The upstream ArrayBufferWriter test gated this on Windows/macOS, and so do we. The
+        // risk is real on Linux but comes from the allocation itself, not from Advance(): under memory
+        // overcommit the kernel may grant the large buffer optimistically, then the GC's
+        // zero-initialization of the array touches every page, which can invoke the OOM killer (an
+        // uncatchable SIGKILL, not an OutOfMemoryException) and tear down the whole test run.
+        [Test]
+        [Slow]
```
NightOwl888: We should probably make this `[Nightly]` rather than `[Slow]`.

src/Lucene.Net.Tests.Analysis.Common/Support/Analysis/Util/TestOpenStringBuilderBufferWriter.cs
```diff
@@ -0,0 +1,506 @@
...
+        // allocation fails with OutOfMemoryException, the test is a no-op. Marked [Slow] due to the
+        // large allocation.
+        //
+        // NOTE: The upstream ArrayBufferWriter test gated this on Windows/macOS, and so do we. The
+        // risk is real on Linux but comes from the allocation itself, not from Advance(): under memory
+        // overcommit the kernel may grant new char[2_000_000_000] optimistically, then the GC's
+        // zero-initialization of the array touches every page, which can invoke the OOM killer (an
+        // uncatchable SIGKILL, not an OutOfMemoryException) and tear down the whole test run.
+        [Test]
+        [Slow]
```
NightOwl888: We should probably make this `[Nightly]` rather than `[Slow]`.

src/Lucene.Net.Analysis.Common/Analysis/Util/OpenStringBuilder.cs
paulirwin: I updated the minimums, even though we're talking about an edge case. 

> It is reasonable to expect the "don't care" size to at least hold an average word.

Wanted to respond to this to say, it typically would in almost all cases even if the sizeHint fallback was 1, because the usual behavior would be to get a buffer of length 32 (when new), some arbitrary length less than 32 (if some small string has already been written), or a doubled-length buffer. Again, the only case where you would get one of length 1 would be if you had already written exactly 31 characters, or whatever `length - 1` is after it has doubled. But, I adjusted the minimums anyways.

NightOwl888 (approved):

PR #1345, opened 2026-06-13
Performance improvements for DataInput/DataOutput and bit-shifting logic, #1279

### LINKED ISSUE #1279: Investigate performance improvement opportunities for DataInput/DataOutput

### Task description

Upon code review, I found that DataInput and DataOutput do some potentially-expensive operations that could be improved. Doing multiple virtual calls to read/write single bytes at a time is potentially very wasteful, and especially in the VInt cases, most subclasses don't always override them with more efficient implementations (`BufferedIndexInput` is the rare exception).

DataInput:
1. `ReadInt16`/`ReadInt32`/`ReadInt64` call `ReadByte()` for every single byte, which is a virtual call done twice, four times, and eight times respectively. We could instead call `ReadBytes(Span<byte>)` once with a stackalloc byte array of length 2/4/8 respectively.
2. `ReadVInt32`/`ReadVInt64` can read up to 5 or 9 bytes respectively. Unfortunately because they are "variable" (hence the "V"), we can't always simply read the full 5 or 9 bytes, or we might go past the end of the file/stream/data. But we should explore if we could add a virtual `TryReadBytes(Span<byte>)` method that tries to read the given number of bytes of the span length and returning false if at EOF, where the default implementation just returns false, falling back to the individual `ReadByte` calls in the slow path. The XML docs should be updated to inform implementers that they should implement this method if possible for performance gains. All Lucene.NET-shipped implementations should implement this method if possible.

DataOutput:
1. `WriteInt16`/`WriteInt32`/`WriteInt64` - same as with DataInput's Read methods, we could write multiple bytes at a time via `WriteBytes(ReadOnlySpan<byte>)` and a stackalloc byte array of fixed length.
2. `WriteVInt32`/`WriteVInt64` - note that unlike with reading VInts, here we actually can know the length in advance. We should replace the body of this with a call to `WriteBytes(ReadOnlySpan<byte>)` with a stackalloc array of fixed length (where the variable length is computed at runtime based on the input, but will not exceed 5/9 bytes respectively)
3. (Minor) Investigate if we can remove the suspicious `WriteBytes(Array.Empty<byte>())` call in `WriteChars` for the empty case. We've already written 0 as the variable length before this, so implementations should not be reading anything when reading this string value back after the first 0 length read. But, maybe there is some compatibility reason to retain this.

paulirwin:

Performance improvements for DataInput/DataOutput and bit-shifting logic

Fixes #1279

## Description

First, this PR does a pass through all bit-shifting logic in the app to replace that with BinaryPrimitives where possible. At a minimum, this at least makes the code a little more maintainable and less likely to have human error. But these methods also help with JIT intrinsicification opportunities (i.e. `movbe` for `BinaryPrimitives.WriteInt32BigEndian` on x64).

But the bigger gains are in optimizing DataInput and DataOutput, which resulted in reduced allocations and faster execution (or at worst, within noise of beta 17). In addition to BinaryPrimitives use, I was able to stack-allocate some small arrays and call the Span methods to avoid multiple virtual calls. Additionally, some types override the best methods now for even greater performance gains, i.e. those that can work with byte arrays directly.

### 1. Indexed reads — NIO (`BufferedIndexInput` / FSDirectory), the read hot path

| Method | beta17 | PR | Δ time | beta17 alloc | PR alloc |
|---|--:|--:|--:|--:|--:|
| NIO_ReadInt16        | 29.0 µs | 24.6 µs | **−15.1%** | 6300 B | **1888 B** |
| NIO_ReadInt32        | 38.4 µs | 28.7 µs | **−25.3%** | 6300 B | **1888 B** |
| NIO_ReadInt64        | 58.1 µs | 36.8 µs | **−36.6%** | 6300 B | **1888 B** |
| NIO_ReadVInt32_Small | 23.8 µs | 20.4 µs | **−14.1%** | 6300 B | **1888 B** |
| NIO_ReadVInt32_Mixed | 26.9 µs | 23.7 µs | **−12.0%** | 6300 B | **1888 B** |
| NIO_ReadVInt32_Large | 42.1 µs | 31.8 µs | **−24.5%** | 6300 B | **1888 B** |
| NIO_ReadVInt64_Small | 24.7 µs | 20.7 µs | **−16.3%** | 6300 B | **1888 B** |
| NIO_ReadVInt64_Mixed | 29.5 µs | 26.0 µs | **−11.7%** | 6300 B | **1888 B** |
| NIO_ReadVInt64_Large | 62.7 µs | 43.8 µs | **−30.2%** | 6300 B | **1888 B** |

NIO reads are 12–37% faster with a ~70% allocation reduction (6300 B → 1888 B).

### 2. Indexed reads — MMap (`ByteBufferIndexInput` / MMapDirectory)

| Method | beta17 | PR | Δ time | beta17 alloc | PR alloc |
|---|--:|--:|--:|--:|--:|
| MMap_ReadInt16        | 93.2 µs  | 92.0 µs  | −1.3% | 1976 B | 1784 B |
| MMap_ReadInt32        | 170.4 µs | 170.2 µs | −0.1% | 1976 B | 1784 B |
| MMap_ReadInt64        | 334.0 µs | 333.2 µs | −0.2% | 1976 B | 1784 B |
| MMap_ReadVInt32_Small | 54.3 µs  | 52.7 µs  | −2.8% | 1976 B | 1784 B |
| MMap_ReadVInt32_Mixed | 71.2 µs  | 69.4 µs  | −2.5% | 1976 B | 1784 B |
| MMap_ReadVInt32_Large | 208.8 µs | 209.0 µs | +0.1% | 1976 B | 1784 B |
| MMap_ReadVInt64_Small | 54.2 µs  | 52.7 µs  | −2.8% | 1976 B | 1784 B |
| MMap_ReadVInt64_Mixed | 86.3 µs  | 84.7 µs  | −1.8% | 1976 B | 1784 B |
| MMap_ReadVInt64_Large | 363.3 µs | 363.9 µs | +0.2% | 1976 B | 1784 B |

MMap read timing is flat (`ByteBufferIndexInput` reads VInts via the byte-by-byte base,
matching upstream Java — a dedicated override was prototyped and measured to regress the
common small-VInt case, so it was not taken).

### 3. Indexed writes — FS (`BufferedIndexOutput` / FSDirectory)

| Method | beta17 | PR | Δ time |
|---|--:|--:|--:|
| FS_WriteInt64        | 333.4 µs | 256.7 µs | **−23.0%** |
| FS_WriteVInt32_Large | 310.0 µs | 244.2 µs | **−21.2%** |
| FS_WriteVInt64_Large | 404.9 µs | 336.8 µs | **−16.8%** |
| FS_WriteVInt32_Mixed | 196.3 µs | 189.2 µs | −3.6% |
| FS_WriteVInt64_Mixed | 209.1 µs | 202.4 µs | −3.2% |
| FS_WriteVInt32_Small | 187.7 µs | 184.3 µs | −1.8% |
| FS_WriteVInt64_Small | 197.1 µs | 199.2 µs | +1.1% |

Multi-byte VInt and Int64 writes gain 17–23%; small/mixed are near-optimal already.

### 4. Fixed-width primitives, per-call cost (ns/call)

Reads from `ByteArrayDataInput` (overrides), and the inheriting `RAMInputStream` /
`InputStreamDataInput` (ISDI) which use the base. Writes from the array-backed
`ByteArrayDataOutput` (BAD) and `GrowableByteArrayDataOutput` (Grow), both of which now
have direct-array overrides.

| Method | beta17 | PR | Δ |
|---|--:|--:|--:|
| **Reads — override (ByteArrayDataInput)** | | | |
| BAD_ReadInt16  | 0.774 | 0.413 | **−46.7%** |
| BAD_ReadInt32  | 1.303 | 0.345 | **−73.5%** |
| BAD_ReadInt64  | 2.342 | 0.334 | **−85.7%** |
| **Reads — inherit base (controls; must not regress)** | | | |
| RAM_ReadInt16  | 0.880 | 0.881 | ±0.0% |
| RAM_ReadInt32  | 1.753 | 1.753 | ±0.0% |
| RAM_ReadInt64  | 3.728 | 3.385 | −9.2% |
| ISDI_ReadInt16 | 3.536 | 3.486 | −1.4% |
| ISDI_ReadInt32 | 6.972 | 6.927 | −0.6% |
| ISDI_ReadInt64 | 14.073 | 3.508 | **−75.1%** |
| **Writes — override (ByteArrayDataOutput)** | | | |
| BAD_WriteInt16 | 0.895 | 0.534 | **−40.4%** |
| BAD_WriteInt32 | 1.685 | 0.532 | **−68.4%** |
| BAD_WriteInt64 | 2.023 | 0.533 | **−73.6%** |
| **Writes — override (GrowableByteArrayDataOutput)** | | | |
| Grow_WriteInt16 | 1.331 | 0.661 | **−50.3%** |
| Grow_WriteInt32 | 2.357 | 0.690 | **−70.7%** |
| Grow_WriteInt64 | 4.346 | 0.717 | **−83.5%** |

The override paths collapse to a single direct `BinaryPrimitives` read/write (cost
independent of width). The base `ReadInt16/32` and `WriteInt16/32` stay byte-by-byte
(upstream-Java parity) so inheriting subclasses without a cheap contiguous buffer (RAM,
ISDI) are unchanged rather than regressed; `ReadInt64`/`WriteInt64` stay span-based on
the base, which breaks even or wins there too (RAM −9%, ISDI −75% via batched
`Stream.Read`).


src/Lucene.Net.Tests.Analysis.Common/Analysis/Payloads/TestPayloadHelper.cs
```diff
@@ -0,0 +1,130 @@
+// Lucene version compatibility level 4.8.1
```
NightOwl888: Please move this file into the `Support` folder in the path:

src/Lucene.Net.Tests.Analysis.Common/Support/Analysis/Payloads/TestPayloadHelper.cs

Also remove the Lucene version compatibility level since this wasn't ported from upstream.

src/Lucene.Net/Store/ByteArrayDataOutput.cs
```diff
@@ -111,5 +112,31 @@ public override void WriteBytes(ReadOnlySpan<byte> source)
...
         }
+
+        // LUCENENET specific: write fixed-width values directly into the backing array via
+        // BinaryPrimitives, avoiding the base class's per-byte WriteByte (Int16/Int32) or
+        // stackalloc + virtual WriteBytes(Span) hop (Int64). Mirrors the ReadInt16/32/64
+        // overrides in ByteArrayDataInput. The base implementations stay byte-by-byte /
+        // span-based for subclasses without a cheap contiguous buffer. See #1279.
+        public override void WriteInt16(short i)
+        {
+            if (Debugging.AssertsEnabled) Debugging.Assert(pos + sizeof(short) <= limit);
```
NightOwl888: Okay, interesting question.

We translated the upstream asserts like this because it is sometimes nearly impossible to determine from looking at the code whether:

1. There is any user benefit to turning them on and off
2. There is any testing benefit to turning them on and off

Since in Java the same `AssertionError` type is used for both production code and JUnit, trying to figure out which of the asserts is meant as:

1. A test condition
2. An invariant that only Lucene developers need to be concerned with
3. An invariant that both Lucene developers and Lucene users need to be concerned with

is not possible.

Did you think through this scenario and decide that there is a user benefit for these? Or perhaps did you intend these to make tests fail if they are tripped? If not, if it were me, I would simply make these into `Debug.Assert()` statements because they will be compiled out of the Release build and yield slightly better performance. In fact, for a hot path like this, I would highly recommend it.

The `Debugging` class exists entirely to translate upstream asserts in a similar way to ensure we don't miss any of the benefits the Lucene designers intended them to perform (at the expense of keeping them in the Release output). See: https://github.com/apache/lucenenet/issues/346#issuecomment-697148885

src/Lucene.Net.Tests/Support/Util/TestGrowableByteArrayDataOutput.cs
```diff
@@ -0,0 +1,77 @@
+using Lucene.Net.Attributes;
```
NightOwl888, line 1: Please place this in a top level `Support` folder.

src/Lucene.Net.Tests/Support/Util/TestGrowableByteArrayDataOutput.cs

src/Lucene.Net/Support/VIntUtils.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// </summary>
+        /// <param name="source">The source bytes to read from.</param>
+        /// <param name="value">The decoded value (undefined if the method returns <c>false</c>).</param>
+        /// <param name="count">The number of bytes consumed. Always set, regardless of return value,
+        /// so that callers can advance their position even on a malformed VInt — matching the
+        /// behavior of the original byte-by-byte readers (which advanced before throwing).</param>
+        /// <returns><c>true</c> if the VInt was well-formed; <c>false</c> if the 5th byte had
+        /// extra high bits set.</returns>
+        public static bool TryReadVInt32(ReadOnlySpan<byte> source, out int value, out int count)
+        {
```
NightOwl888, line 57: Let's add a `Debug.Assert()` here - it may save some headaches later.

```c#
Debug.Assert(source.Length >= MaxVInt32Length);
```

If we later make this class public, we should consider using `ThrowHelper`-style exceptions that are not compiled inline.

```c#
if (source.Length < MaxVInt32Length)
    ThrowHelper.ThrowArgumentException_SourceTooShort();
```

src/Lucene.Net/Support/VIntUtils.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// in length to avoid possible out-of-range exceptions.
+        /// </summary>
+        /// <param name="source">The source bytes to read from.</param>
+        /// <param name="value">The decoded value (undefined if the method returns <c>false</c>).</param>
+        /// <param name="count">The number of bytes consumed. Always set, regardless of return value
+        /// — see <see cref="TryReadVInt32"/>.</param>
+        /// <returns><c>true</c> if the VInt was well-formed; <c>false</c> if the 9th byte had
+        /// the continuation bit set (which would indicate a negative value, disallowed).</returns>
+        public static bool TryReadVInt64(ReadOnlySpan<byte> source, out long value, out int count)
+        {
```
NightOwl888, line 111: Let's add a `Debug.Assert()` here - it may save some headaches later.

```c#
Debug.Assert(source.Length >= MaxVInt64Length);
```

src/Lucene.Net/Codecs/Compressing/CompressingStoredFieldsReader.cs
```diff
@@ -431,6 +431,50 @@ public override void ReadBytes(Span<byte> destination)
...
+                    return base.ReadVInt32();
+                }
+
+                // fast path: avoid repeated virtual calls
+                bool ok = VIntUtils.TryReadVInt32(bytesRef.AsSpan(), out int result, out int count);
+                bytesRef.Offset += count;
+                bytesRef.Length -= count;
+                if (!ok)
+                {
+                    throw new System.IO.IOException("Invalid VInt32 detected (too many bits)");
```
NightOwl888: We can generate cleaner IL code by moving the throw call into a dedicated helper method that is not inlined.

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/ThrowHelper.cs#L5-L35

Making a fully dedicated `ThrowHelper` with resources for the solution would be a different project (and different PR), but for now we could at least get a similar benefit by making a static method with `[NoInlining]` and `[DoesNotReturn]` attributes that wraps the throw and string. Since we are sharing the same messages, we could probably just drop those methods into `VIntUtils` for now.

Do note that I have been using helper files like:

```c#
        private static class SR
        {
            public const string IO_StreamTooLong = "Stream was too long.";
            public const string IO_EOF_ReadBeyondEOF = "Unable to read beyond the end of the stream.";
        }
```

To keep the syntax aligned with the BCL so dropping the strings into a `SR.resx` later will be simple to do and won't require touching every exception site.

src/Lucene.Net/Codecs/Compressing/CompressingStoredFieldsReader.cs
```diff
@@ -431,6 +431,50 @@ public override void ReadBytes(Span<byte> destination)
...
+                    return base.ReadVInt64();
+                }
+
+                // fast path: avoid repeated virtual calls
+                bool ok = VIntUtils.TryReadVInt64(bytesRef.AsSpan(), out long result, out int count);
+                bytesRef.Offset += count;
+                bytesRef.Length -= count;
+                if (!ok)
+                {
+                    throw new System.IO.IOException("Invalid VInt64 detected (negative values disallowed)");
```
NightOwl888: We can generate cleaner IL code by moving the throw call into a dedicated helper method that is not inlined.

See my comment on `CompressingStoredFieldsReader.ReadVInt32()` for more context.

src/Lucene.Net/Index/ByteSliceReader.cs
```diff
@@ -175,5 +179,45 @@ public override void ReadBytes(Span<byte> destination)
...
+                // slow path near end of slice
+                return base.ReadVInt32();
+            }
+
+            // fast path: avoid repeated virtual calls
+            bool ok = VIntUtils.TryReadVInt32(buffer.AsSpan(upto), out int result, out int count);
+            upto += count;
+            if (!ok)
+            {
+                throw new System.IO.IOException("Invalid VInt32 detected (too many bits)");
```
NightOwl888: We can generate cleaner IL code by moving the throw call into a dedicated helper method that is not inlined.

See my comment on `CompressingStoredFieldsReader.ReadVInt32()` for more context.

src/Lucene.Net/Index/ByteSliceReader.cs
```diff
@@ -175,5 +179,45 @@ public override void ReadBytes(Span<byte> destination)
...
+                // slow path near end of slice
+                return base.ReadVInt64();
+            }
+
+            // fast path: avoid repeated virtual calls
+            bool ok = VIntUtils.TryReadVInt64(buffer.AsSpan(upto), out long result, out int count);
+            upto += count;
+            if (!ok)
+            {
+                throw new System.IO.IOException("Invalid VInt64 detected (negative values disallowed)");
```
NightOwl888: We can generate cleaner IL code by moving the throw call into a dedicated helper method that is not inlined.

See my comment on `CompressingStoredFieldsReader.ReadVInt32()` for more context.

src/Lucene.Net/Store/BufferedIndexInput.cs
```diff
@@ -277,38 +293,14 @@ public override sealed int ReadVInt32()
...
-                b = m_buffer[bufferPosition++];
-                i |= (b & 0x7F) << 7;
-                if (b <= sbyte.MaxValue) // LUCENENET: Optimized equivalent of "if ((sbyte)b >= 0)"
+                // LUCENENET: unify logic in VIntUtils. Note that existing code was not checking bytes length.
+                bool ok = VIntUtils.TryReadVInt32(m_buffer.AsSpan(bufferPosition), out int value, out int count);
+                bufferPosition += count;
+                if (!ok)
                 {
-                    return i;
+                    throw new IOException("Invalid VInt32 detected (too many bits)");
```
NightOwl888: We can generate cleaner IL code by moving the throw call into a dedicated helper method that is not inlined.

See my comment on `CompressingStoredFieldsReader.ReadVInt32()` for more context.

src/Lucene.Net/Store/BufferedIndexInput.cs
```diff
@@ -323,61 +315,14 @@ public override sealed long ReadVInt64()
...
-                b = m_buffer[bufferPosition++];
-                i |= (b & 0x7FL) << 56;
-                if (b <= sbyte.MaxValue) // LUCENENET: Optimized equivalent of "if ((sbyte)b >= 0)"
+                // LUCENENET: unify logic in VIntUtils. Note that existing code was not checking bytes length.
+                bool ok = VIntUtils.TryReadVInt64(m_buffer.AsSpan(bufferPosition), out long value, out int count);
+                bufferPosition += count;
+                if (!ok)
                 {
-                    return i;
+                    throw new IOException("Invalid VInt64 detected (negative values disallowed)");
```
NightOwl888: We can generate cleaner IL code by moving the throw call into a dedicated helper method that is not inlined.

See my comment on `CompressingStoredFieldsReader.ReadVInt32()` for more context.

src/Lucene.Net/Store/ByteArrayDataInput.cs
```diff
@@ -94,129 +95,78 @@ public override void SkipBytes(long count)
...
-            i |= (b & 0x7FL) << 56;
-            if (b <= sbyte.MaxValue) // LUCENENET: Optimized equivalent of "if ((sbyte)b >= 0)"
+            // LUCENENET: unify logic in VIntUtils. See ReadVInt32() above for why this
+            // intentionally omits a MaxVInt64Length bounds guard and base fallback.
+            bool ok = VIntUtils.TryReadVInt64(bytes.AsSpan(pos), out long value, out int count);
+            pos += count;
+            if (!ok)
             {
-                return i;
+                throw RuntimeException.Create("Invalid VInt64 detected (negative values disallowed)");
```
NightOwl888: We can generate cleaner IL code by moving the throw call into a dedicated helper method that is not inlined.

See my comment on `CompressingStoredFieldsReader.ReadVInt32()` for more context.

src/Lucene.Net/Store/ByteArrayDataInput.cs
```diff
@@ -94,129 +95,78 @@ public override void SkipBytes(long count)
...
+            // intentionally omits a MaxVInt32Length bounds guard (and the base.ReadVInt32()
+            // fallback): per its class summary it "omits all low-level checks", matching the
+            // original byte-by-byte code which read straight from the array and let an
+            // out-of-range access throw on truncated/malformed input.
+            bool ok = VIntUtils.TryReadVInt32(bytes.AsSpan(pos), out int value, out int count);
+            pos += count;
+            if (!ok)
             {
-                return i;
+                throw RuntimeException.Create("Invalid VInt32 detected (too many bits)");
```
NightOwl888: We can generate cleaner IL code by moving the throw call into a dedicated helper method that is not inlined.

See my comment on `CompressingStoredFieldsReader.ReadVInt32()` for more context.

src/Lucene.Net/Util/Fst/ForwardBytesReader.cs
```diff
@@ -51,6 +51,44 @@ public override void ReadBytes(Span<byte> destination)
...
+                // slow path near end of array
+                return base.ReadVInt32();
+            }
+
+            // fast path: avoid repeated virtual calls
+            bool ok = VIntUtils.TryReadVInt32(bytes.AsSpan(pos), out int result, out int count);
+            pos += count;
+            if (!ok)
+            {
+                throw new System.IO.IOException("Invalid VInt32 detected (too many bits)");
```
NightOwl888: We can generate cleaner IL code by moving the throw call into a dedicated helper method that is not inlined.

See my comment on `CompressingStoredFieldsReader.ReadVInt32()` for more context.

src/Lucene.Net/Util/Fst/ForwardBytesReader.cs
```diff
@@ -51,6 +51,44 @@ public override void ReadBytes(Span<byte> destination)
...
+                // slow path near end of array
+                return base.ReadVInt64();
+            }
+
+            // fast path: avoid repeated virtual calls
+            bool ok = VIntUtils.TryReadVInt64(bytes.AsSpan(pos), out long result, out int count);
+            pos += count;
+            if (!ok)
+            {
+                throw new System.IO.IOException("Invalid VInt64 detected (negative values disallowed)");
```
NightOwl888: We can generate cleaner IL code by moving the throw call into a dedicated helper method that is not inlined.

See my comment on `CompressingStoredFieldsReader.ReadVInt32()` for more context.

src/Lucene.Net/Util/StringHelper.cs
```diff
@@ -215,15 +220,15 @@ public static int Murmurhash3_x86_32(byte[] data, int offset, int len, int seed)
             switch (len & 0x03)
             {
                 case 3:
-                    k2 = (((sbyte)data[roundedEnd + 2]) & 0xff) << 16;
+                    k2 = data[roundedEnd + 2] << 16;
```
NightOwl888, line 227: Please add a comment to explain why we diverged from upstream here.

src/Lucene.Net/Util/StringHelper.cs
```diff
@@ -215,15 +220,15 @@ public static int Murmurhash3_x86_32(byte[] data, int offset, int len, int seed)
...
                     // fallthrough
                     goto case 2;
                 case 2:
-                    k2 |= (((sbyte)data[roundedEnd + 1]) & 0xff) << 8;
+                    k2 |= data[roundedEnd + 1] << 8;
```
NightOwl888, line 231: Please add a comment to explain why we diverged from upstream here.

src/Lucene.Net/Util/StringHelper.cs
```diff
@@ -215,15 +220,15 @@ public static int Murmurhash3_x86_32(byte[] data, int offset, int len, int seed)
...
                     // fallthrough
                     goto case 1;
                 case 1:
-                    k2 |= (((sbyte)data[roundedEnd]) & 0xff);
+                    k2 |= data[roundedEnd];
```
NightOwl888, line 235: Please add a comment to explain why we diverged from upstream here.

NightOwl888 (changes requested):
Thanks for the PR.

I am glad to see that we are moving in the right direction performance-wise. However, the more interesting benchmark would be to compare it against the point where this PR branched from `master` to consider the changes here alone.

I suspect a huge chunk of the NIO gains we are seeing are due to switching from using `seek` operations to simulate random access to actually calling the `RandomAccess.Read()` method.

src/Lucene.Net/Store/ByteArrayDataOutput.cs
paulirwin: I agree that this is the right call for a hot path, so I went ahead and made that change. However, benchmarks showed negligible benefit. That might have been because of JIT speculation because the value is always false. I still think we should keep it regardless.

src/Lucene.Net/Support/VIntUtils.cs
```diff
@@ -0,0 +1,238 @@
...
+        // pattern this mirrors: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/ThrowHelper.cs
+
+        /// <summary>
+        /// Throws an <see cref="IOException"/> indicating a malformed 32-bit VInt. Used by the
+        /// base <see cref="Lucene.Net.Store.DataInput"/> and the buffered/stream-backed readers,
+        /// which surface decode failures as I/O errors.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        [DoesNotReturn]
+        public static void ThrowInvalidVInt32()
```
NightOwl888: Oops, I didn't notice we are throwing 2 different exception types. Let's name these using the same convention as the BCL `ThrowHelper` classes.

```
ThrowIOException_InvalidVInt32()
ThrowIOException_InvalidVInt64()
ThrowRuntimeException_InvalidVInt32()
ThrowRuntimeException_InvalidVInt64()
```

These could then later be moved to a `ThrowHelper` class without renaming them and they would be sensible to sit side-by-side with `ThrowIOException(ExeptionResource.XXX)`, etc. overloads.

More importantly, it is straightforward at the call site which exception this method throws (or is intended to mimic in the case of the JDK exceptions).

NightOwl888 (changes requested):

src/Lucene.Net/Support/VIntUtils.cs
paulirwin: I like these names better; done.

NightOwl888 (approved):

PR #1346, opened 2026-06-13
Finish port of WindowsDirectory and NativeUnixDirectory, #1342

### LINKED ISSUE #1342: Finish port of NativeUnixDirectory and WindowsDirectory

The ports for NativeUnixDirectory and WindowsDirectory were never completed, and are currently excluded from compilation from the Lucene.Net.Misc project, Store directory. We should complete the port of these types for API parity with upstream.

https://github.com/apache/lucene/tree/releases/lucene-solr/4.8.1/lucene/misc/src/java/org/apache/lucene/store

Note that these were originally implemented in a mix of Java and C++ to be able to do native calls. However, they are quite simple, and I am fairly certain we can accomplish these via P/Invoke instead of having to add C++ to our project, without having any significant performance impact. This appears to be the direction that was started already but never completed.

Also note that as of Lucene 4.8.1, these types had no unit tests. We should add unit test coverage for these, restricted to running on their respective platforms.

The types are ported and work well, with good unit test coverage added.

### Additional context

Note that these types were removed in Lucene 9.0.0, but someone might find good use for them between now and then.

See #1341 discussion.

paulirwin:

Finish port of WindowsDirectory and NativeUnixDirectory

Fixes #1342

## Description

This PR finishes the work that was started on WindowsDirectory and NativeUnixDirectory.

Upstream, this code is a mix of Java and C++. Instead of adding in true native code, I finished the approach that was started of using P/Invoke to call the Win32/POSIX native functions. 

This PR also adds tests for these types, which had no tests upstream. These tests all pass on Windows 11 x64, Ubuntu 24.04 x64 in WSL, Ubuntu 24.04 arm64 in Docker, and macOS 26 arm64. 

## Benchmark Results

These benchmarks are on macOS 26 arm64:

| Dir (64 MB) | Write | ReadSequential | ReadRandom |
|---|--:|--:|--:|
| NativeUnix | 19.1 ms | 47.8 ms | 380.6 ms |
| SimpleFS | 149 ms | 3.5 ms | 1.0 ms |
| NIOFS | 137 ms | 5.4 ms | 1.1 ms |
| MMap | 139 ms | 618 ms | 5.6 ms |
| RAM | 148 ms | 1.8 ms | 0.06 ms |

Notes:
- NativeUnix is an order of magnitude faster for writes, but is much slower for reads, especially random reads. This is due to deliberately bypassing the OS-level IO cache with `O_DIRECT`. It is worth reviewing to determine if this read performance should be improved.
- The MMap slow read performance is likely related to #1151 

I have not yet run benchmarks on WindowsDirectory.

AI: This PR was mostly written using Claude Code, Opus 4.8.


src/Lucene.Net.Misc/Store/NativePosixUtil.cs
```diff
@@ -20,45 +27,222 @@ namespace org.apache.lucene.store
...
+
+        private static IOException NewIOException(string operation, string filename)
+            => NewIOException(operation, filename, Marshal.GetLastWin32Error());
+
+        private static IOException NewIOException(string operation, string filename, int errno)
+        {
+            string where = filename is null ? operation : $"{operation} {filename}";
+            // On Unix with SetLastError, GetLastWin32Error() returns errno. Win32Exception's message
+            // is not meaningful on Unix, so include the raw errno for diagnosis.
+            return new IOException($"{where} failed (errno {errno})");
```
NightOwl888: Make sure to call the overload that accepts `hresult`. Our `NativeFSLock` depends on this.

src/Lucene.Net.Misc/Store/NativePosixUtil.cs
```diff
@@ -20,45 +27,222 @@ namespace org.apache.lucene.store
...
+        /// </summary>
+        /// <exception cref="IOException"> If <c>posix_fadvise</c> returned a non-zero code </exception>
+        public static void Advise(SafeFileHandle fd, long offset, long len, int advise)
+        {
+            EnsureUnix();
+            int code = NativeMethods.posix_fadvise(Fd(fd), offset, len, MapAdvice(advise));
+            if (code != 0)
+            {
+                // LUCENENET: upstream throws RuntimeException; we use IOException as this is an I/O failure.
+                throw new IOException("posix_fadvise failed code=" + code);
```
NightOwl888: Make sure to call the overload that accepts `hresult`. Our `NativeFSLock` depends on this.

src/Lucene.Net.Misc/Store/NativePosixUtil.cs
```diff
@@ -20,45 +27,222 @@ namespace org.apache.lucene.store
...
+        /// </summary>
+        /// <returns> the number of bytes read (may be less than <paramref name="length"/> near EOF) </returns>
+        /// <exception cref="IOException"> If the read failed </exception>
+        public static long Pread(SafeFileHandle fd, long pos, IntPtr buffer, int length)
+        {
+            EnsureUnix();
+            long n = (long)NativeMethods.pread(Fd(fd), buffer, (nuint)length, pos);
+            if (n < 0)
+            {
+                throw NewIOException("pread", null);
```
NightOwl888, line 129: Make sure to call the overload that accepts `hresult`. Our `NativeFSLock` depends on this.

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
```diff
@@ -35,493 +34,573 @@ namespace org.apache.lucene.store
...
-          //System.out.println("dump: done");
-
-          // TODO: the case where we'd seek'd back, wrote an
-          // entire buffer, we must here read the next buffer;
-          // likely Lucene won't trip on this since we only
-          // write smallish amounts on seeking back
+            long written = (long)NativeMethods.pwrite(Fd(fd), buffer, (nuint)length, pos);
+            if (written < 0)
+            {
+                throw new IOException($"pwrite failed (errno {Marshal.GetLastWin32Error()})");
```
NightOwl888: Make sure to call the overload that accepts `hresult`. Our `NativeFSLock` depends on this.

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
```diff
@@ -35,493 +34,573 @@ namespace org.apache.lucene.store
...
+            }
         }
 
-        public long FilePointer
+        private static void FTruncate(SafeFileHandle fd, long length)
         {
-          return filePos + bufferPos;
+            if (NativeMethods.ftruncate(Fd(fd), length) < 0)
+            {
+                throw new IOException($"ftruncate failed (errno {Marshal.GetLastWin32Error()})");
```
NightOwl888: Make sure to call the overload that accepts `hresult`. Our `NativeFSLock` depends on this.

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
```diff
@@ -35,493 +34,573 @@ namespace org.apache.lucene.store
...
-            int delta = (int)(pos - alignedPos);
-            buffer.position(delta);
-            bufferPos = delta;
-          }
+            // The original used FileChannel.size(). All reads here are positioned (pread), so
+            // moving the file offset with lseek(SEEK_END) to learn the size is harmless.
+            long size = NativeMethods.lseek(Fd(fd), 0, NativeMethods.SEEK_END);
+            if (size < 0)
+            {
+                throw new IOException($"lseek failed (errno {Marshal.GetLastWin32Error()})");
```
NightOwl888: Make sure to call the overload that accepts `hresult`. Our `NativeFSLock` depends on this.

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
```diff
@@ -35,493 +34,573 @@ namespace org.apache.lucene.store
...
+            public IntPtr Pointer => (IntPtr)alignedPtr;
+
+            public void Clear() { position = 0; limit = capacity; }
+            public void Flip() { limit = position; position = 0; }
+            public void Rewind() { position = 0; }
+
+            public void Put(byte b) => alignedPtr[position++] = b;
+            public byte Get() => alignedPtr[position++];
+
+            public void Put(ReadOnlySpan<byte> src)
```
NightOwl888, line 264: There are a couple of potential performance optimizations here, but I would benchmark to be sure it is worth it.

1. `Unsafe` copy.

```c#
public void Put(ReadOnlySpan<byte> src)
{
    Unsafe.CopyBlockUnaligned(
        ref Unsafe.AsRef<byte>(alignedPtr + position),
        ref MemoryMarshal.GetReference(src),
        (uint)src.Length);

    position += src.Length;
}
```

2. `ref byte` parameter type

```c#
public void Put(ref byte src, int length)
```

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
```diff
@@ -35,493 +34,573 @@ namespace org.apache.lucene.store
...
+            {
+                src.CopyTo(new Span<byte>(alignedPtr + position, src.Length));
+                position += src.Length;
+            }
+
+            public void Get(Span<byte> dst)
```
NightOwl888, line 270: Just like the Put() method, we could make similar optimizations:

1. `Unsafe` copy

```c#
public void Get(Span<byte> dst)
{
    Unsafe.CopyBlockUnaligned(
        ref MemoryMarshal.GetReference(dst),
        ref Unsafe.AsRef<byte>(alignedPtr + position),
        (uint)dst.Length);

    position += dst.Length;
}
```

2. `ref byte` parameter type

```c#
public void Get(ref byte dst, int length)
```

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
```diff
@@ -35,493 +34,573 @@ namespace org.apache.lucene.store
...
+                get => fileLength + bufferPos;
+                set { /* not supported: length is managed internally */ }
             }
-          }
-        }
 
-        public long FilePointer
-        {
-          return filePos + bufferPos;
+            public override long Checksum => throw new NotSupportedException("this directory currently does not work at all!");
```
NightOwl888: We should be able to implement the checksum pretty easily because it is just a `CRC32` instance that we need to instantiate and we just need to update it every time `WriteByte` or `WriteBytes()` is called here.

https://github.com/paulirwin/lucene.net/blob/df1029337712995e5dff434798b6664c68dedc26/src/Lucene.Net/Store/FSDirectory.cs#L513-L562

There is a minor issue with keeping multiple instances of `CRC32` around, but if we fix that we also need to ensure that the instance isn't exposed publicly because the `IChecksum` interface is baggage from Java. We made an attempt to convert this to the BCL `HashAlgorithm` base class but it turned out the approach used in Java produced far fewer allocations than the one in .NET: https://github.com/apache/lucenenet/pull/436#issuecomment-796797535

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
```diff
@@ -35,493 +34,573 @@ namespace org.apache.lucene.store
...
+                filePos += bufferSize;
+                bufferPos = 0;
+                long n;
+                try
+                {
+                    n = NativePosixUtil.Pread(sharedFd, filePos, buffer.Pointer, bufferSize);
+                }
+                catch (Exception ioe) when (ioe.IsIOException())
+                {
+                    throw new IOException(ioe.Message + ": " + this, ioe);
```
NightOwl888: This seems a bit risky. `NativeFSLock` depends on the `HResult` of the `IOException`. So, if we wrap the exception, we should at the very least surface the `HResult` from the original exception.

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
```diff
@@ -35,493 +34,573 @@ namespace org.apache.lucene.store
...
+            {
+                get
+                {
+                    try
+                    {
+                        return FileSize(sharedFd);
+                    }
+                    catch (Exception ioe) when (ioe.IsIOException())
+                    {
+                        throw RuntimeException.Create("IOException during Length: " + this, ioe);
```
NightOwl888: This seems a bit risky. `NativeFSLock` depends on the `HResult` of the `IOException`. So, if we wrap the exception, we should at the very least surface the `HResult` from the original exception.

src/Lucene.Net.Misc/Store/WindowsDirectory.cs
```diff
@@ -21,161 +28,294 @@ namespace org.apache.lucene.store
...
+            protected override void ReadInternal(Span<byte> b)
+            {
+                int bytesRead;
+                try
+                {
+                    bytesRead = WindowsDirectory.Read(fd, b, Position); // LUCENENET: Position is the file pointer (renamed from getFilePointer())
+                }
+                catch (Exception ioe) when (ioe.IsIOException())
+                {
+                    throw new IOException(ioe.Message + ": " + this, ioe);
```
NightOwl888: This seems a bit risky. `NativeFSLock` depends on the `HResult` of the `IOException`. So, if we wrap the exception, we should at the very least surface the `HResult` from the original exception.

src/Lucene.Net.Misc/Store/WindowsDirectory.cs
```diff
@@ -21,161 +28,294 @@ namespace org.apache.lucene.store
...
+            return numRead;
+        }
+
+        /// <summary>
+        /// Returns the length of a file. </summary>
+        private static long Length(SafeFileHandle fd)
+        {
+            if (!NativeMethods.GetFileInformationByHandle(fd, out NativeMethods.BY_HANDLE_FILE_INFORMATION info))
+            {
+                throw new IOException(new Win32Exception(Marshal.GetLastWin32Error()).Message);
```
NightOwl888: Make sure to call the overload that accepts `hresult`. Our `NativeFSLock` depends on this.

src/Lucene.Net.Misc/Store/WindowsDirectory.cs
```diff
@@ -21,161 +28,294 @@ namespace org.apache.lucene.store
...
+            int numRead;
+            bool success;
+            fixed (byte* p = bytes)
+            {
+                success = NativeMethods.ReadFile(fd, p, bytes.Length, out numRead, &overlapped);
+            }
+
+            if (!success)
+            {
+                throw new IOException(new Win32Exception(Marshal.GetLastWin32Error()).Message);
```
NightOwl888: Make sure to call the overload that accepts `hresult`. Our `NativeFSLock` depends on this.

src/Lucene.Net.Misc/Store/WindowsDirectory.cs
```diff
@@ -21,161 +28,294 @@ namespace org.apache.lucene.store
...
+                NativeMethods.FILE_SHARE_READ | NativeMethods.FILE_SHARE_WRITE | NativeMethods.FILE_SHARE_DELETE,
+                IntPtr.Zero,
+                NativeMethods.OPEN_EXISTING,
+                NativeMethods.FILE_FLAG_RANDOM_ACCESS,
+                IntPtr.Zero);
+            int lastError = Marshal.GetLastWin32Error();
+
+            if (handle == NativeMethods.INVALID_HANDLE_VALUE)
+            {
+                throw new IOException("Could not open file " + filename + ": " + new Win32Exception(lastError).Message);
```
NightOwl888: Make sure to call the overload that accepts `hresult`. Our `NativeFSLock` depends on this.

src/Lucene.Net.Tests.Misc/Support/Store/TestNativeUnixDirectory.cs
```diff
@@ -0,0 +1,334 @@
+using Lucene.Net.Attributes;
```
NightOwl888, line 1: Please move this file into a top level `Support` folder.

src/Lucene.Net.Tests.Misc/Support/Store/TestNativeUnixDirectory.cs

src/Lucene.Net.Tests.Misc/Support/Store/TestWindowsDirectory.cs
```diff
@@ -0,0 +1,281 @@
+using Lucene.Net.Attributes;
```
NightOwl888, line 1: Please move this file into a top level `Support` folder.

src/Lucene.Net.Tests.Misc/Support/Store/TestWindowsDirectory.cs

NightOwl888 (changes requested):
Thanks.

Based on the benchmarks, I am starting to wonder whether we should aim to use native calls (or at least `RandomAccess`) instead of `FileStream` for writes on all of our `FSDirectory` implementations, since upstream they simply used `FileChannel` (which is a thin layer over native calls). Although, whatever solution we come up with, we should aim for getting the durability patched (#894) above all else.

Shouldn't we add these native calls to our existing [native layer](https://github.com/apache/lucenenet/tree/e2e5d9738588d50f014aaa9c9559f5586b7cf0b8/src/Lucene.Net/Support/Native) so we can utilize them to make a backport of `RandomAccess` for .NET Framework and other things that we are better off using native code for?

I didn't do a full review of this, but I found a few things we should address to make this implementation more complete.

paulirwin:
Benchmarks for WindowsDirectory, before making PR feedback modifications above, are in:

64 MB file (mean, ms):

| Dir | ReadSequential | ReadRandom | ConcurrentCloneReads |
|---|--:|--:|--:|
| Windows | 11.84 | 7.84 | 103.4 |
| SimpleFS | 11.72 | 8.52 | 79.0 |
| NIOFS | 20.83 | 7.55 | 133.3 |
| MMap | 415.59 | 5.97 | 937.1 |
| RAM | 6.49 | 0.12 | 4.7 |

Notes:
- WindowsDirectory does not override CreateOutput with a custom implementation, so it was not benchmarked
- Windows is faster than the average times of SimpleFS and NIOFS for non-concurrent reads.
- Windows is within noise of SimpleFS for sequential reads, and faster for random reads. It is slightly slower than NIOFS for random reads.
- Concurrent reads track well between SimpleFS and NIOFS.
- Again, concurrent reads in MMap is likely #1151

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
paulirwin: I had Claude Code generate exhaustive benchmarks of these changes, and the small improvements were dwarfed in magnitude by I/O.

| Size | Put: Span → UnsafeSpan | Get: Span → UnsafeSpan |
|---|---|---|
| 8 B | 1.22 → 1.18 ns (0.97x) | 1.43 → 1.14 ns (0.94x) |
| 32 B | 1.51 → 1.19 ns (**0.78x**) | 1.38 → 1.11 ns (**0.73x**) |
| 256 B | 5.72 → 4.67 ns (0.82x) | 3.64 → 3.69 ns (1.0x) |
| 4 KB | 76.1 → 77.3 ns (1.02x) | 59.2 → 58.8 ns (0.99x) |
| 64 KB | 612 → 605 ns (0.99x) | 819 → 816 ns (1.00x) |

A 64MB direct write makes ~256 put calls; optimizing them saves ~77ns out of 19ms (0.0000004%; note the difference in units). It only saves a small amount for very small copies (8-32B) which are fairly rare with index directories. So I think we can skip this particular optimization.

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
paulirwin: See benchmarks response above; skipping for now.

paulirwin:
I meant to reply earlier to your top-level review comment... replies below.

> Based on the benchmarks, I am starting to wonder whether we should aim to use native calls (or at least `RandomAccess`) instead of `FileStream` for writes on all of our `FSDirectory` implementations, since upstream they simply used `FileChannel` (which is a thin layer over native calls). Although, whatever solution we come up with, we should aim for getting the durability patched (#894) above all else.

I like this, but it will not change the public API, so it can be broken out as a separate issue (even for post-beta-19) as a perf optimization opportunity.

> Shouldn't we add these native calls to our existing [native layer](https://github.com/apache/lucenenet/tree/e2e5d9738588d50f014aaa9c9559f5586b7cf0b8/src/Lucene.Net/Support/Native) so we can utilize them to make a backport of `RandomAccess` for .NET Framework and other things that we are better off using native code for?

Same as above, I think this can be broken out separately as a possible post-beta-19 perf optimization opportunity. In this particular case, these calls are very OS-dependent just like they are upstream (hence the names of the directories), so there's no harm in them being alongside the OS-specific directory code for now. That would be an easy refactor later as part of this possible perf improvement as a separate issue/PR.

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
```diff
@@ -494,7 +509,16 @@ public override long Length
                     }
                     catch (Exception ioe) when (ioe.IsIOException())
                     {
-                        throw RuntimeException.Create("IOException during Length: " + this, ioe);
+                        // LUCENENET: upstream wraps in a RuntimeException here (IndexInput.Length cannot
+                        // throw a checked IOException in Java). Preserve the original HResult so callers
+                        // that inspect it (e.g. NativeFSLock) still see the underlying error. The HResult
+                        // setter is only public on modern TFMs; on net462/netstandard2.0 it is protected,
+                        // so the detail is retained in the message there instead.
+                        Exception wrapped = RuntimeException.Create("IOException during Length: " + this, ioe);
```
NightOwl888: Given this is a `RuntimeException`, I suspect reverting this to your original implementation would be correct here. `HResult` is only relevant to `IOExceptions` in Lucene.NET. That would be preferable to conditionally compiling IMO.

Also, I think I overstated the importance of it on `IOException.HResult`. `NativeFSLock` creates its own `FileStream` instance and catches `IOException` with a very specific `HResult` value from that `FileStream` instance when testing whether the `write.lock` file is open for sharing/locking. So, exceptions from `FSDirectory` implementations aren't likely to come into play at all. However, it is still a reasonable improvement to pass the HResult value from native methods to `IOException`, since that is exactly what the BCL does.

src/Lucene.Net.Misc/Store/NativeUnixDirectory.cs
```diff
@@ -567,7 +594,14 @@ public override object Clone()
                 }
                 catch (Exception ioe) when (ioe.IsIOException())
                 {
-                    throw RuntimeException.Create("IOException during clone: " + this, ioe);
+                    // LUCENENET: as in Length above, upstream wraps in a RuntimeException (Clone cannot
+                    // throw a checked IOException in Java). Preserve the original HResult where the setter
+                    // is public (modern TFMs); on net462/netstandard2.0 the detail stays in the message.
+                    Exception wrapped = RuntimeException.Create("IOException during clone: " + this, ioe);
```
NightOwl888: Given this is a `RuntimeException`, I suspect reverting this to your original implementation would be correct here. `HResult` is only relevant to `IOExceptions` in Lucene.NET. That would be preferable to conditionally compiling IMO.

NightOwl888 (changes requested):

paulirwin:
@NightOwl888 Updated to revert to the original throw statements. I also took the opportunity to convert the file to use string interpolation to clean that up a little bit.

Additionally, I noticed that the `overview.md` file still had a commented out bit about how we hadn't ported NativeUnixDirectory. I restored most of that text except for the native build instructions which are not needed on our side, and added a note about the use of P/Invoke instead.

PR #1347, opened 2026-06-14
Finish port of Lucene.Net.Collation functionality, #1343

### LINKED ISSUE #1343: Finish port of Lucene.Net.Collation functionality
From discussion in #1341: 
> Lucene.Net.Analysis.Common is missing the [Collation functionality](https://github.com/apache/lucenenet/tree/80b2a66f74f15f5596a2e2a03078038269cf87d1/src/Lucene.Net.Analysis.Common/Collation). It probably isn't much work. .NET's collator is called [CompareInfo](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.compareinfo?view=net-10.0). The Lucene docs say that the ICU4J collator is faster than the JDK, but I suspect that in the BCL implementation will be faster than the one in ICU4N in most cases.

These types are currently excluded from compilation via the `FEATURE_COLLATION` flag not being defined. We should finish the implementation (including the tests) for feature parity with Lucene 4.8.1.

paulirwin:

Finish port of Lucene.Net.Collation functionality

Fixes #1343 

## Description

Complete the abandoned WIP under Lucene.Net.Analysis.Common/Collation, which was gated behind an undefined FEATURE_COLLATION flag and referenced the icu.net package (never a solution dependency, so it never compiled).

- Reimplement CollationKeyAnalyzer, CollationAttributeFactory, CollationKeyFilter, CollationKeyFilterFactory and CollatedTermAttributeImpl on the BCL platform collator (System.Globalization.CompareInfo / SortKey / CompareOptions), mirroring Java's java.text.Collator. Remove all FEATURE_COLLATION guards; no new dependency.
- CompareInfo has no tailoring API, so custom RuleBasedCollator rules are not supported: CollationKeyFilterFactory's "custom" path throws NotSupportedException (use the ICU package for custom rulesets).
- Normalize terms to NFC before generating sort keys so decomposed and precomposed input collate consistently across the NLS (.NET Framework) and ICU (.NET 5+) globalization backends.
- Port the collation tests to CompareInfo. TestCollationKeySort's French ordering is backend-dependent (NLS -> EACGI, matching upstream; ICU -> ECAGI), so it is detected at runtime via CollationTestBase.FranceResult.
- Re-enable the platform-collator implementation in the benchmark NewCollationAnalyzerTask (the upstream "JDK" impl, renamed to "BCL"; "jdk" kept as a legacy alias for Java .alg compatibility) and cover both impls in TestPerfTasksLogic.TestCollator.
- Port Collation/package.md examples and caveats to .NET idioms.

Verified: full Tests.Analysis.Common and Tests.Benchmark suites green on net10.0 and net48.

AI: Mostly written with Claude Code, Opus 4.8.


src/Lucene.Net.Analysis.Common/Collation/TokenAttributes/CollatedTermAttributeImpl.cs
```diff
@@ -29,27 +28,45 @@ namespace Lucene.Net.Collation.TokenAttributes
     /// </summary>
     public class CollatedTermAttributeImpl : CharTermAttribute
```
NightOwl888: Please remove the `Impl` suffx to match our other attributes, but leave the file named `CollatedTermAttributeImpl.cs`.

src/Lucene.Net.Analysis.Common/Collation/TokenAttributes/CollatedTermAttributeImpl.cs
```diff
@@ -29,27 +28,45 @@ namespace Lucene.Net.Collation.TokenAttributes
...
+            this.collator = collator;
+            this.options = options;
+        }
 
-            //LUCENENET TODO: Verify that this is correct. Java's byte[] is signed and Big Endian, .NET's is unsigned and Little Endian.
-            bytes.Bytes = this.collator.GetSortKey(this.ToString()).KeyData;
+        public override void FillBytesRef()
+        {
+            BytesRef bytes = this.BytesRef;
+            byte[] keyData = this.collator.GetSortKey(Normalize(this.ToString()), this.options).KeyData;
```
NightOwl888, line 74: Let's conditionally compile. We don't have to convert this attribute to a string at all on .NET Core when using ICU because there is a `ReadOnlySpan<char>` overload of `GetSortKey()`. So, we can use `this.AsSpan()` to provide the value without an allocation. Note that the destination is a `Span<byte>` so there is a `GetSortKeyLength()` method that can be called first to determine how much space to allocate.

However, we still need to use the slow path if the app is configured for NLS. According to [Determine if your app  is using ICU](https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu#determine-if-your-app-is-using-icu), we can check that by using:

```c#
public static bool ICUMode()
{
    SortVersion sortVersion = CultureInfo.InvariantCulture.CompareInfo.Version;
    byte[] bytes = sortVersion.SortId.ToByteArray();
    int version = bytes[3] << 24 | bytes[2] << 16 | bytes[1] << 8 | bytes[0];
    return version != 0 && version == sortVersion.FullVersion;
}
```

But the fast path will be the default on .NET core, so IMO it is worth it to factor out the string allocations.

I suggest we make the above into a property that is statically cached and we can use it at runtime to determine whether ICU is in use or we need to fallback for NLS support. Prior to .NET core, the slow path will be the only one that exists and the fast path won't compile because the APIs don't exist.

Please do the same in the `CollationKeyFilter`.

src/Lucene.Net.Analysis.Common/Collation/package.md
```diff
@@ -40,73 +40,93 @@ summary: *content
...
-      // with a Farsi Collator (or an Arabic one for the case when Farsi is not
-      // supported).
-      ScoreDoc[] result
-        = is.search(aqp.parse("[ \u062F TO \u0698 ]"), null, 1000).scoreDocs;
-      assertEquals("The index Term should not be included.", 0, result.length);
+```cs
+CompareInfo collator = CompareInfo.GetCompareInfo("ar");
+CollationKeyAnalyzer analyzer = new CollationKeyAnalyzer(LuceneVersion.LUCENE_48, collator);
+Store.Directory ramDir = new RAMDirectory();
+IndexWriter writer = new IndexWriter(ramDir, new IndexWriterConfig(LuceneVersion.LUCENE_48, analyzer));
```
NightOwl888: Let's demonstrate proper use of `using` here. If we use the inline version, the below `writer.Dispose()` approach can still be used, but `IndexWriter` will receive 2 `Dispose()` calls unless there is an error.

Let's do that across all of these code examples for `IndexReader`, `IndexWriter`. and other things where disposal is required (note that on analyzers they are optional).

src/Lucene.Net.Analysis.Common/Collation/package.md
```diff
@@ -40,73 +40,93 @@ summary: *content
...
 
- __WARNING:__ Make sure you use exactly the same `Collator` at index and query time -- `CollationKey`s are only comparable when produced by the same `Collator`. Since {@link java.text.RuleBasedCollator}s are not independently versioned, it is unsafe to search against stored `CollationKey`s unless the following are exactly the same (best practice is to store this information with the index and check that they remain the same at query time):
+ __WARNING:__ Make sure you use exactly the same collator (`System.Globalization.CompareInfo` *and* `System.Globalization.CompareOptions`) at index and query time -- `System.Globalization.SortKey`s are only comparable when produced by the same collator. Since the platform collator is not independently versioned, it is unsafe to search against stored `System.Globalization.SortKey`s unless the following are exactly the same (best practice is to store this information with the index and check that they remain the same at query time):
 
-1.  JVM vendor
+1.  The .NET runtime version, and the active globalization backend. .NET Framework uses Windows NLS, while .NET 5+ uses ICU by default; the two produce different sort keys and orderings.
```
NightOwl888: As a convenience, let's make the `IsNLS` property public somewhere so we can document using it here.

Or even better, add some helper that users can utilize to add all of these values to the index with minimal effort. Perhaps we should make a `CollationUtil` class with these settings ready to index and a comparison check to utilize whether the values from the index are a match.

src/Lucene.Net.Analysis.Common/Collation/package.md
```diff
@@ -40,73 +40,93 @@ summary: *content
...
 
- `CollationKey`s generated by `java.text.Collator`s are not compatible with those generated by ICU Collators. Specifically, if you use `CollationKeyAnalyzer` to generate index terms, do not use `ICUCollationKeyAnalyzer` on the query side, or vice versa.
\ No newline at end of file
+ `System.Globalization.SortKey`s generated by the platform collator are not compatible with those generated by ICU Collators. Specifically, if you use `CollationKeyAnalyzer` to generate index terms, do not use `ICUCollationKeyAnalyzer` on the query side, or vice versa.
```
NightOwl888, line 135: I guess it is okay to document it this way, but it isn't necessarily true. ICU4J and ICU4C use the same collation key when:

- They are based on the same ICU release.
- They use the same locale and tailoring rules.
- They use the same collation settings (strength, case level, alternate handling, numeric collation, etc.).
- They are using the same collation data.

But it would take a bit of digging to figure out how .NET maps the inputs from the user to how they are presented on the API in ICU4N to figure out how to make those guarantees. And of course, it assumes a future version of ICU4N that matches one in .NET Core. So, for the time being we can consider this true.

src/Lucene.Net.Benchmark/ByTask/Tasks/NewCollationAnalyzerTask.cs
```diff
@@ -47,11 +46,12 @@ public static Type GetCollatorType(this NewCollationAnalyzerTask.Implementation
         {
             switch (impl)
             {
-                //case NewCollationAnalyzerTask.Implementation.JDK:
-                //    return typeof(Icu.Collation.Collator);
+                case NewCollationAnalyzerTask.Implementation.BCL:
```
NightOwl888: Although BCL is understood by .NET developers, note that it is an antiquated term that refers specifically to .NET Framework. Modern terminology simply refers to ".NET runtime" or ".NET libraries".

For this case, it would be better to convert to `DotNet` (or I guess `DOT_NET` if we are keeping the uppercasing scheme).

> Side note - while I don't consider it high priority, I still think it would be best if we convert the enum and constants to follow the .NET naming conventions before the release. But I care less about the actual scheme than consistency across public APIs of the codebase.

src/Lucene.Net.Benchmark/ByTask/Tasks/NewCollationAnalyzerTask.cs
```diff
@@ -61,18 +61,24 @@ public static Type GetCollatorType(this NewCollationAnalyzerTask.Implementation
...
         public enum Implementation
         {
-            //JDK, // LUCENENET: Not supported
+            // LUCENENET: This value is named JDK in upstream Lucene (the collator comes from the
+            // Java Development Kit's java.text.Collator). It has been renamed to BCL here because the
+            // equivalent in .NET is the platform collator (System.Globalization.CompareInfo) provided
+            // by the Base Class Library. The "jdk" parameter value is still accepted as a legacy alias.
+
+            /// <summary>The .NET platform collator (<see cref="CompareInfo"/>), equivalent to the JDK's <c>java.text.Collator</c>.</summary>
+            BCL,
```
NightOwl888: See my note above about naming.

src/Lucene.Net.Analysis.Common/Collation/package.md
```diff
@@ -40,73 +40,93 @@ summary: *content
...
-      // Unicode order would include U+0633 in [ U+062F - U+0698 ], but Farsi
-      // orders the U+0698 character before the U+0633 character, so the single
-      // indexed Term above should NOT be returned by a ConstantScoreRangeQuery
-      // with a Farsi Collator (or an Arabic one for the case when Farsi is not
-      // supported).
-      ScoreDoc[] result
-        = is.search(aqp.parse("[ \u062F TO \u0698 ]"), null, 1000).scoreDocs;
-      assertEquals("The index Term should not be included.", 0, result.length);
+```cs
+CompareInfo collator = CompareInfo.GetCompareInfo("ar");
```
NightOwl888: Farsi has its own culture that should be used. It should be `fa` rather than `ar`. This works on both .NET Framework and .NET Core (with ICU).

src/Lucene.Net.Tests.Analysis.Common/Collation/TestCollationKeyAnalyzer.cs
```diff
@@ -32,42 +27,32 @@ namespace Lucene.Net.Collation
     [TestFixture]
     public class TestCollationKeyAnalyzer : CollationTestBase
     {
+        // Neither Java 1.4.2 nor 1.5.0 has Farsi Locale collation available in
+        // RuleBasedCollator.  However, the Arabic Locale seems to order the Farsi
+        // characters properly.
+        private readonly CompareInfo collator = CompareInfo.GetCompareInfo("ar");
```
NightOwl888: Farsi has its own culture that should be used. It should be `fa` rather than `ar`. This works on both .NET Framework and .NET Core (with ICU).

src/Lucene.Net.TestFramework/Analysis/CollationTestBase.cs
NightOwl888: Please remove this comment - this was only relevant to the attempt to use icu-net.

src/Lucene.Net.TestFramework/Analysis/CollationTestBase.cs
```diff
@@ -48,6 +48,22 @@ public abstract class CollationTestBase : LuceneTestCase
...
+        /// the platform collator (<see cref="System.Globalization.CompareInfo"/>) depends on the active
+        /// globalization backend. .NET Framework (NLS) applies traditional French (backward) accent ordering and
+        /// produces "EACGI" - the same result as Lucene's <c>java.text.Collator</c> with <c>Locale.FRANCE</c> -
+        /// whereas .NET 5+ (ICU) uses forward accent ordering and produces "ECAGI". <see cref="System.Globalization.CompareInfo"/>
+        /// has no option to select French reverse-accent ordering, and its sort keys are consistent with its
+        /// <see cref="System.Globalization.CompareInfo.Compare(string, string)"/>, so we detect the active behavior
+        /// here rather than hard-coding a single (backend-specific) value. This is only relevant to tests that use
+        /// the platform collator; the ICU-based collation tests have their own expected values.
+        /// </summary>
+        protected static string FranceResult =>
```
NightOwl888: This should be `FrenchResult`. `FranceResult` would be `fr-FR`.

src/Lucene.Net.TestFramework/Analysis/CollationTestBase.cs
NightOwl888: Please reflow the `sortData` array to match upstream.

https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/test-framework/src/java/org/apache/lucene/analysis/CollationTestBase.java#L196-L210

src/Lucene.Net.Tests.Analysis.Common/Collation/TestCollationKeyFilter.cs
```diff
@@ -34,44 +30,41 @@ namespace Lucene.Net.Collation
     [Obsolete("remove when CollationKeyFilter is removed.")]
     public class TestCollationKeyFilter : CollationTestBase
     {
+        // Neither Java 1.4.2 nor 1.5.0 has Farsi Locale collation available in
+        // RuleBasedCollator.  However, the Arabic Locale seems to order the Farsi
+        // characters properly.
+        private readonly CompareInfo collator = CompareInfo.GetCompareInfo("ar");
```
NightOwl888: Farsi has its own culture that should be used. It should be `fa` rather than `ar`. This works on both .NET Framework and .NET Core (with ICU).

src/Lucene.Net.Tests.Analysis.Common/Collation/TestCollationKeyFilterFactory.cs
```diff
@@ -56,7 +50,7 @@ public virtual void TestBasicUsage()
         [Test]
         public virtual void TestNormalization()
         {
-            var turkishUpperCase = "I W\u0049\u0307LL USE TURKİSH CASING";
+            var turkishUpperCase = "I WİLL USE TURKİSH CASING";
```
NightOwl888: Please revert this back to the original test condition.

src/Lucene.Net.Tests.Analysis.Common/Collation/TestCollationKeyFilterFactory.cs
```diff
@@ -93,50 +87,6 @@ public virtual void TestSecondaryStrength()
             AssertCollatesToSame(tsUpper, tsLower);
         }
 
-        /// <summary>
```
NightOwl888, line 96: Please add a comment indicating why this test was removed.

https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/analysis/common/src/test/org/apache/lucene/collation/TestCollationKeyFilterFactory.java#L108-L142

NightOwl888 (changes requested):

src/Lucene.Net.Analysis.Common/Collation/TokenAttributes/CollatedTermAttributeImpl.cs
```diff
@@ -27,29 +27,67 @@ namespace Lucene.Net.Collation.TokenAttributes
...
+            this.collator = collator;
+            this.options = options;
+        }
 
-            //LUCENENET TODO: Verify that this is correct. Java's byte[] is signed and Big Endian, .NET's is unsigned and Little Endian.
-            bytes.Bytes = this.collator.GetSortKey(this.ToString()).KeyData;
+        public override void FillBytesRef()
+        {
+            BytesRef bytes = this.BytesRef;
+#if FEATURE_COMPAREINFO_SPAN_SORTKEY
```
NightOwl888: Let's change the name of this to `FEATURE_COMPAREINFO_GETSORTKEY_SPAN` to make it more consistent with other feature names that use `FEATURE_<type name>[_<member name>[_<parameter or feature name>]]`

`GetSortKeyLength` is implied here by the new contract.

src/Lucene.Net.Analysis.Common/Collation/CollationUtil.cs
```diff
@@ -0,0 +1,59 @@
+// Lucene version compatibility level 4.8.1
```
NightOwl888: Please remove the `// Lucene version compatibility` comment and move this file in to a top level `Support` folder (without changing the namespace).

src/Lucene.Net.Analysis.Common/Support/Collation/CollationUtil.cs

src/Lucene.Net.Analysis.Common/Collation/TokenAttributes/CollatedTermAttributeImpl.cs
```diff
@@ -27,29 +27,67 @@ namespace Lucene.Net.Collation.TokenAttributes
...
+            byte[] keyData = this.collator.GetSortKey(Normalize(this.ToString()), this.options).KeyData;
+            bytes.Bytes = keyData;
             bytes.Offset = 0;
-            bytes.Length = bytes.Bytes.Length;
+            bytes.Length = keyData.Length;
         }
+
+#if FEATURE_COMPAREINFO_SPAN_SORTKEY
+        // LUCENENET: Returns the term text as a span over the backing buffer, avoiding a string allocation.
+        private ReadOnlySpan<char> AsSpan() => this.Buffer.AsSpan(0, this.Length);
```
NightOwl888: We can use the extension method here, but it is currently in `Lucene.Net.MemoryExtensions` and requires that we import the `Lucene.Net` namespace. We really ought to move these extension methods into the `Lucene.Net.Analysis.TokenAttributes` namespace (in a `CharTermAttributeExtensions` class) so they are harder to miss (in a different PR, that is).

This is the kind of thing that takes experimentation to get right, and I missed this on the first pass.

NightOwl888 (changes requested):

src/Lucene.Net.Analysis.Common/Collation/CollationUtil.cs
paulirwin: Sorry I missed this comment, Claude Code did this automatically. I've course-corrected its memory to help prevent it in the future.

src/Lucene.Net.Analysis.Common/Collation/CollationKeyFilter.cs
```diff
@@ -104,23 +109,24 @@ public override bool IncrementToken()
...
-            Arrays.Copy(arr, 0, unsigned, 0, arr.Length);
-            return unsigned;
+#if FEATURE_COMPAREINFO_GETSORTKEY_SPAN
+            // LUCENENET: On .NET 5+ with ICU, CompareInfo has a ReadOnlySpan<char> overload of GetSortKey
+            // that lets us generate the sort key without converting the term to a string first. The NLS
+            // backend does not normalize internally (see CollatedTermAttribute.Normalize), so we still need
+            // the slow path when the app is configured for NLS.
+            if (CollationUtil.IsICU)
+            {
+                ReadOnlySpan<char> source = this.termAtt.Buffer.AsSpan(0, this.termAtt.Length);
```
NightOwl888: Please use the extension method for `termAtt`. It is unnecessary to access the `termAtt.Length` property here.

NightOwl888 (changes requested):
Sorry, one more quick thing to address, then this is approved.

src/Lucene.Net.Analysis.Common/Collation/CollationKeyFilter.cs
paulirwin: Good catch, fixed.

NightOwl888 (approved):

Issue #1340, opened 2026-06-12
Add support for C# 14.0


NightOwl888:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Task description

In both [J2N](https://github.com/NightOwl888/J2N/issues/172) and [ICU4N](https://github.com/NightOwl888/ICU4N/issues/120), we have migrated to C# 14.0. 

C# 14.0 comes with 2 very important features:

1. [First-class span types](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-14.0/first-class-span-types) allow `string` to be intrinsically converted to `ReadOnlySpan<char>` without calling the (slower) `.AsSpan()` method. This will save us from having to add `string` overloads just for the sake of cross-target compilation.
2. [Extension members](https://devblogs.microsoft.com/dotnet/csharp-exploring-extension-members/) allow us to extend classes with *static* methods to polyfill methods on legacy target frameworks. For example, we can add `string.Concat()` overloads that accept `ReadOnlySpan<char>`.

This is enough motivation for adding support ASAP, since it will drastically shape how we implement support for `ReadOnlySpan<char>`. Since we already build with the .NET 10 SDK (done in #1222), changing this should be straightforward, although there may be some cleanup required.

We should also do a sweep to ensure:

- `<LangVersion>` is only defined at the solution level
- Any calls to `.AsSpan()` that can be removed are removed

paulirwin:
+∞

Added to the milestone for the next major beta, due to "ASAP" which I agree with.

paulirwin:
@NightOwl888 I'm going to mark this fixed once #1374 is merged. If there are any specific examples of extension members that should be added, that can and should be its own issue since it's not specifically about C# 14 or optimizations directly due to it (like removing AsSpan).

PR #1352, opened 2026-06-15
Use C# 14 and fix field warnings/errors, #1340

paulirwin:

Use C# 14 and fix field warnings/errors

Partial #1340 

## Description

This is part 1 of 3 for #1340. I will do the ReadOnlySpan and extension block changes as follow-up PRs.

This sets LangVersion to 14 for the solution, and removes the explicit set of that from one place. It also fixes the `field` keyword warnings and errors, by either renaming local variables or adding `this.` qualification (to avoid a more extensive refactoring of renaming the fields).

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -26,7 +26,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <LangVersion>11.0</LangVersion>
+    <LangVersion>14.0</LangVersion>
     <GitHubOrganization>apache</GitHubOrganization>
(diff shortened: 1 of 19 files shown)


NightOwl888 (approved):
Thanks for the PR.

Approved with an optional correction.

I forgot about this weird `field` naming issue that made me initially reject this. For other keywords such as `@in`, `@out`, and `@delgate`, we use the `@` character to escape them so we can use the same variable name as upstream. I think it would be better to escape them than to rename the variables to `f`. But `this.field` is fine.

PR #1353, opened 2026-06-16
Bumped ICU4N to 60.1.0-alpha.440

### LINKED ISSUE #998: Failing random test seed/culture: en-IE with seed 0x9a2b7430d6d33f0d

### Describe the bug

Found a random failing test seed/culture pair. Set the following in a `lucene.testsettings.json` file in your repo root (or somewhere between the test executable and the root of your drive):

```json
{
  "tests": {
     "seed": "0x9a2b7430d6d33f0d",
     "culture": "en-IE"
  }
}
```

This causes the following tests to fail on these frameworks:

`Lucene.Net.Index.TestIndexReaderClose.TestCloseUnderException()` (net5.0, net48, net472)

Message: 
> System.ArgumentException : The supplied DateTime represents an invalid time.  For example, when the clock is adjusted forward, any time in the period that is skipped is invalid. (Parameter 'dateTime')

`Lucene.Net.Util.Events.TestPubSubEvent.CanAddSubscriptionWhileEventIsFiringNonGeneric()` (net472)

Message:
> Expected: True  Actual: False

`Lucene.Net.Analysis.Icu.TestICUNormalizer2CharFilter.TestRandomStrings()` (net5.0, net48)

Message:
> Lucene.Net.Diagnostics.AssertionException : End() called before IncrementToken() returned false!

### Expected Behavior

The tests all pass.

### Steps To Reproduce

_No response_

### Exceptions (if any)

_No response_

### Lucene.NET Version

4.8.0-beta00017

### .NET Version

Various (listed above)

### Operating System

Windows

### Anything else?

x64 platform, running `./build.ps1 -t`

NightOwl888:

Fixes #998

## Description

This bumps ICU4N to 60.1.0-alpha.440 which contains the `Normalizer2` concurrency patch in https://github.com/NightOwl888/ICU4N/pull/122 that addresses the problems with random failures of `TestICUNormalizer2CharFilter.TestRandomStrings()` in concurrent environments.

### `ThaiTokenizer`

I made an attempt at documenting the "differences" between how `ThaiTokenizer` behaves and how `ICUTokenizer` behaves with Thai text. However, after adding several tests to check how they deal with transitions between Thai and non-Thai words and numerals, it turns out they behave identically in all cases I checked. I ended up temporarily replacing the `ThaiTokenizer` with the implementation of `ICUTokenizer` and discovered that all of the tests pass. So, the `ThaiWordBreaker` effectively is a poor-man's way of implementing part of the UAX #29 http://unicode.org/reports/tr29/ spec, but we cannot guarantee that it does completely comply for the Thai language.

I ended up adding documentation explaining that it does not guarantee to be compliant with the Unicode spec and recommended to use `ICUTokenizer` instead. But I am wondering whether we should just rewire the `ICUTokenizer` into the `ThaiTokenizer` and eliminate the "JDK compatible" attempt that is not guaranteed to be compatible with the JDK, anyway.

The JDK `RuleBasedBreakIterator` behaves differently across different vendors and versions, so there really is no stable target to hit, anyway. Some implementations don't even include a dictionary-based `BreakIterator` implementation, so Thai tokenization is impossible. Historically, this is one of the main motivations for creating the `lucene-analysis-icu` package which guarantees stability across JDK versions.

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/Th/ThaiTokenizer.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Th/ThaiTokenizer.cs
@@ -48,7 +48,7 @@ namespace Lucene.Net.Analysis.Th
     /// <para/>
     /// Unlike the JDK, this implementation is guaranteed to be stable across all supported target frameworks.
     /// While it does use ICU4N's <see cref="RuleBasedBreakIterator"/>, this implementation doesn't follow
-    /// the UAX #29 specification (http://unicode.org/reports/tr29) and is not guranteed to behave the same as
+    /// the UAX #29 specification (http://unicode.org/reports/tr29) and is not guaranteed to behave the same as
     /// either the one in the JDK or in ICU4J.
     /// <para/>
     /// This implementation is provided primarily for API compatibility with Lucene. If strict Unicode compliance
--- a/src/Lucene.Net.Analysis.Common/Analysis/Th/ThaiTokenizer.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Th/ThaiTokenizer.cs
@@ -1,18 +1,15 @@
 // Lucene version compatibility level 4.8.1
 #if FEATURE_BREAKITERATOR
-using ICU4N.Support.Text;
 using ICU4N.Text;
 using J2N;
 using Lucene.Net.Analysis.TokenAttributes;
 using Lucene.Net.Analysis.Util;
 using Lucene.Net.Support;
-using Lucene.Net.Support.Threading;
 using Lucene.Net.Util;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
-using System.Text.RegularExpressions;
 
 namespace Lucene.Net.Analysis.Th
 {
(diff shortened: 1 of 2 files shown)


src/Lucene.Net.Analysis.Common/Analysis/Th/ThaiTokenizer.cs
```diff
@@ -44,6 +41,19 @@ namespace Lucene.Net.Analysis.Th
...
     /// Tokenizer that use <see cref="BreakIterator"/> to tokenize Thai text.
     /// </summary>
+    /// <remarks>
+    /// This is an attempt to mimic the behavior of the JDK's <c>java.Text.BreakIterator</c> approach
+    /// to tokenizing Thai text. While it passes the Lucene tests, there may be innumerable differences
+    /// between this implementation and the one in the JDK.
+    /// <para/>
+    /// Unlike the JDK, this implementation is guaranteed to be stable across all supported target frameworks.
+    /// While it does use ICU4N's <see cref="RuleBasedBreakIterator"/>, this implementation doesn't follow
+    /// the UAX #29 specification (http://unicode.org/reports/tr29) and is not guranteed to behave the same as
```
paulirwin: Found by pre-commit: typo with "guranteed"

paulirwin (approved):
Approved; note the typo that is failing the pre-commit check. Once fixed looks good to merge.

PR #1354, opened 2026-06-16
Reduce Azure DevOps Windows parallelism from 8 to 2, #1350

### LINKED ISSUE #1350: Failing TestPerformance test on Azure DevOps

### Task description

https://dev.azure.com/lucene-net-temp4/Main/_build/results?buildId=392&view=ms.vss-test-web.build-test-results-tab&runId=232884&resultId=100024&paneView=debug (123s)
https://dev.azure.com/lucene-net-temp4/Main/_build/results?buildId=391&view=ms.vss-test-web.build-test-results-tab&runId=232948&resultId=100024&paneView=debug (124s)
https://dev.azure.com/lucene-net-temp4/Main/_build/results?buildId=387&view=ms.vss-test-web.build-test-results-tab&runId=230975&resultId=100016&paneView=debug (183s)

This TestPerformance test has started failing randomly on Azure DevOps. Error:

```
training took more than 2 mins : 123s
(Test: Lucene.Net.Classification.KNearestNeighborClassifierTest.TestPerformance)
```

Note that it is usually just barely past the 2 minute (120s) cutoff, although that one occasion above is 183s.

Since this test is naturally performance-sensitive, it is entirely possible that Azure DevOps had a performance regression at some point, i.e. due to an OS update.

Speaking of OS, I have only seen this fail on Windows agents, and it should be noted that Azure DevOps windows agents are, in my professional experience, _much_ slower than Linux agents. We often get builds to succeed in half the time by switching agents from Windows to Linux. That said, we should probably do due diligence and investigate this properly. There might also be real performance gain opportunities by looking into this.

If it turns out to just be an Azure DevOps issue, we could work around this by setting an environment variable to increase this timeout to something like 5 minutes. 

Although this is unlikely to result in an API change, it is a failing test that takes time out of our day to investigate each time, so it would be good to get this fixed sooner than later. So I've added this to the beta milestone, but it certainly could be moved to the final milestone.

paulirwin:

Reduce Azure DevOps Windows parallelism from 8 to 2

Fixes #1350

## Description

The KNearestNeighborClassifierTest.TestPerformance failures on Azure DevOps Windows agents are caused by CPU starvation. This could have changed over time due to an OS/agent upgrade, a SDK change, or simply that we're running more tests now than we were before (including more concurrent-load tests). The timed region (indexing 1000 docs + opening a reader) runs in under 400ms locally even for the exact seeds that took 123s-183s on Azure, and two of the three failing seeds used an in-memory RAMDirectory (so it is not disk/fsync/AV).

The test runner launches up to maximumParallelJobs dotnet test processes concurrently on a single shared agent VM. On the standard 2-core Windows hosted agents, 8 concurrent assemblies is ~4x oversubscription, which starves the perf-sensitive test and pushes its wall-clock past the 120s threshold. Lowering the Windows concurrency to 2 reduces oversubscription. Linux/macOS are left at 7 since the failures are (currently) Windows-only.

I ran this 9 times on ADO (3 in parallel, then 5 in parallel, then one isolated) and the #1350 failing test never came back. It did expose a few other known and previously-unknown failures which all seem concurrency-related, which is a good thing! These tests were not failing previously, likely because they were taking too long to preemptively multitask on the oversubscribed CPUs. These failing tests will be evaluated independently of this PR.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -346,7 +346,7 @@ stages:
         Windows:
           osName: 'Windows'
           imageName: 'windows-latest'
-          maximumParallelJobs: 8
+          maximumParallelJobs: 2
           maximumAllowedFailures: 0 # Maximum allowed failures for a successful build
         Linux:
           osName: 'Linux'
@@ -381,7 +381,7 @@ stages:
         Windows:
           osName: 'Windows'
           imageName: 'windows-latest'
-          maximumParallelJobs: 8
+          maximumParallelJobs: 2
           maximumAllowedFailures: 0 # Maximum allowed failures for a successful build
     displayName: 'Test net10.0,x86 on'
     pool:
@@ -406,7 +406,7 @@ stages:
         Windows:
           osName: 'Windows'
           imageName: 'windows-latest'
-          maximumParallelJobs: 8
+          maximumParallelJobs: 2
           maximumAllowedFailures: 0 # Maximum allowed failures for a successful build
         Linux:
           osName: 'Linux'
@@ -441,7 +441,7 @@ stages:
         Windows:
           osName: 'Windows'
           imageName: 'windows-latest'
-          maximumParallelJobs: 8
+          maximumParallelJobs: 2
           maximumAllowedFailures: 0 # Maximum allowed failures for a successful build
     displayName: 'Test net9.0,x86 on'
     pool:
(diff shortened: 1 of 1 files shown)


NightOwl888 (approved):
Thanks for tracking this down!

PR #1363, opened 2026-06-18
Fix GroupingSearch.TotalGroupCount null when groupOffset exceeds group count, #1362

### LINKED ISSUE #1362: Grouping Search - TotalGroupCount not returned

### Describe the bug

Hi, I've found a bug with group searching when trying to work out the `TotalGroupCount` if the `groupOffset` value is larger than the amount of groups indexed, `TotalGroupCount` will be null when it shouldn't be. 

The `matchingGroups` is calculated correctly but in the code it is ignored in this return path:

In [GroupingSearch.cs](https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Grouping/GroupingSearch.cs) line 385:

```
if (topSearchGroups is null)
{
        // LUCENENET specific - optimized empty array creation
        return new TopGroups<TGroupValue>(Array.Empty<SortField>(), Array.Empty<SortField>(), 0, 0, Array.Empty<GroupDocs<TGroupValue>>(), float.NaN);
}
```

When `allGroups` is true but `topSearchGroups` is null the return ignores the calculated `matchingGroups` so `TotalGroupCount` is null even though `.SetAllGroups(true);` has been used in the `GroupingSearch` request (and has a valid value to return).

### Expected Behavior

When performing a grouped search with `SetAllGroups(true)` the `TotalGroupCount` should never be null on any return path and should be the calculated value.

### Steps To Reproduce

1) Perform a grouped search with `SetAllGroups(true)`
2) In the `Search()` set the `groupOffset` to a number larger than the amount of groups
3) The `ITopGroups<object>.TotalGroupCount` will be null

### Exceptions (if any)

_No response_

### Lucene.NET Version

4.8.0-beta00017

### .NET Version

_No response_

### Operating System

_No response_

### Anything else?

_No response_

paulirwin:

Fix GroupingSearch.TotalGroupCount null when groupOffset exceeds group count

Fixes #1362

## Description

When SetAllGroups(true) and groupOffset is larger than the number of matching groups, the early-return path now wraps the empty TopGroups with the MatchingGroups.Count so TotalGroupCount is populated correctly.

### DIFF
--- a/src/Lucene.Net.Grouping/GroupingSearch.cs
+++ b/src/Lucene.Net.Grouping/GroupingSearch.cs
@@ -450,7 +450,10 @@ namespace Lucene.Net.Search.Grouping
             if (topSearchGroups is null)
             {
                 // LUCENENET specific - optimized empty array creation
-                return new TopGroups<T>(Array.Empty<SortField>(), Array.Empty<SortField>(), 0, 0, Array.Empty<GroupDocs<T>>(), float.NaN);
+                var emptyResult = new TopGroups<T>(Array.Empty<SortField>(), Array.Empty<SortField>(), 0, 0, Array.Empty<GroupDocs<T>>(), float.NaN);
+                return AllGroups
+                    ? new TopGroups<T>(emptyResult, MatchingGroups.Count)
+                    : emptyResult;
             }
 
             int topNInsideGroup = GroupDocsOffset + GroupDocsLimit;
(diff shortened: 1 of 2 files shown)


NightOwl888 (approved):
Looks good.

This bug exists upstream, also:

https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/grouping/src/java/org/apache/lucene/search/grouping/GroupingSearch.java#L213-L215

I didn't check whether it ever got fixed, but the current `GroupingSearch` has a different design that removes the `search()` methods entirely.

PR #1364, opened 2026-06-18
`.asf.yaml`: add copilot auto code review

jbampton:
https://github.com/apache/infrastructure-asfyaml#copilot-code-review

refs https://github.com/apache/cloudstack/pull/13438

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -43,6 +43,11 @@ github:
     merge: false
     rebase: false
 
+  copilot_code_review:
+    enabled: true
+    review_drafts: true
+    review_on_push: true
+
   rulesets:
     - name: "Branch Protection"
       type: branch


.asf.yaml
```diff
@@ -43,6 +43,11 @@ github:
     merge: false
     rebase: false
 
+  copilot_code_review:
+    enabled: true
+    review_drafts: true
+    review_on_push: true
```
paulirwin, line 49: IMO, we should not re-review every push. That creates a lot of noise, and we've already gotten questions about the volume of Copilot emails on our mailing list. I am, however, in favor of having Copilot automatically review the _first_ push, and then we can manually re-request review at subsequent pushes. Sometimes you need to push multiple times to fix unit tests etc, which would create too much noise.

paulirwin (changes requested):
Comment added, but it's up for debate/discussion.

PR #1374, opened 2026-06-22
Remove unnecessary `string.AsSpan()` calls as of C# 14, #1340

### LINKED ISSUE #1340: Add support for C# 14.0

paulirwin:

Remove unnecessary `string.AsSpan()` calls as of C# 14

Fixes #1340

## Description

I reviewed all instances of calling `.AsSpan()` on strings, which are no longer necessary for method arguments for `ReadOnlySpan<char>` parameters as of C# 14. There weren't very many. I was also able to remove two internal methods from StemmerUtil. In other cases, I left the public string overload APIs so as to not make it a breaking change for non-C#-14 callers and commented as such. The calls that remain are needed to either force the overload resolution for the public API, or are not used as arguments.

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/HyphenationTree.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/HyphenationTree.cs
@@ -375,13 +375,12 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
         ///        hyphenation point. </param>
         /// <returns> a <see cref="Hyphenation"/> object representing the
         ///         hyphenated word or null if word is not hyphenated. </returns>
+        // LUCENENET: string overload retained here for public API for non-C#-14 callers
         public virtual Hyphenation Hyphenate(string word, int remainCharCount, int pushCharCount)
         {
             // LUCENENET: use Span instead of ToCharArray
             return Hyphenate(word.AsSpan(), 0, word.Length, remainCharCount, pushCharCount);
         }
-
-
 
         /// <summary>
         /// Hyphenate word and return an array of hyphenation points.
--- a/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/HyphenationTree.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/HyphenationTree.cs
@@ -366,30 +366,13 @@ namespace Lucene.Net.Analysis.Compound.Hyphenation
         }
 
         /// <summary>
-        /// Hyphenate word and return a <see cref="Hyphenation"/> object.
-        /// </summary>
-        /// <param name="word"> the word to be hyphenated </param>
-        /// <param name="remainCharCount"> Minimum number of characters allowed before the
-        ///        hyphenation point. </param>
-        /// <param name="pushCharCount"> Minimum number of characters allowed after the
-        ///        hyphenation point. </param>
-        /// <returns> a <see cref="Hyphenation"/> object representing the
-        ///         hyphenated word or null if word is not hyphenated. </returns>
-        // LUCENENET: string overload retained here for public API for non-C#-14 callers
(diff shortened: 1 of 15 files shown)


src/Lucene.Net.Analysis.Common/Analysis/Compound/Hyphenation/HyphenationTree.cs
```diff
@@ -375,14 +375,13 @@ protected virtual void SearchPatterns(ReadOnlySpan<char> word, int index, Span<b
         ///        hyphenation point. </param>
         /// <returns> a <see cref="Hyphenation"/> object representing the
         ///         hyphenated word or null if word is not hyphenated. </returns>
+        // LUCENENET: string overload retained here for public API for non-C#-14 callers
         public virtual Hyphenation Hyphenate(string word, int remainCharCount, int pushCharCount)
```
NightOwl888: When you consider that this isn't just for non-C# 14.0 users, but *only* non-.NET Core users who are not on C# 14.0, it seems far less appealing to keep the `string` overloads unless there is some other purpose, such as throwing `ArguementNullException` to inform the user the value is invalid.

Furthermore, this overload and the below overload are reversed in signature. If the intention is to keep a `string` overload, it should be the one with the `offset` and `len` parameters, not the `ReadOnlySpan<char>` overload that can be sliced before it is passed.

Frankly, the only public signature we need here is:

```c#
public virtual Hyphenation Hyphenate(ReadOnlySpan<char> w, int remainCharCount, int pushCharCount)
{
    int offset = 0;
    int len = w.Length;
```

Then we can either eliminate the `string` overload or keep it and add the `offset` and `len` parameters to it. But I think it would be better just to remove it and not worry about whether .NET Framework users are on C# 14.0.

src/Lucene.Net.Analysis.Common/Analysis/Util/CharArrayMap.cs
```diff
@@ -3827,7 +3827,7 @@ internal static CharReturnType ConvertObjectToChars<T>(T key, out char[] chars,
 
 #if FEATURE_SPANFORMATTABLE
             else if (key is ISpanFormattable spanFormattable &&
-                spanFormattable.TryFormat(reuse, out int charsWritten, string.Empty.AsSpan(), invariant))
+                spanFormattable.TryFormat(reuse, out int charsWritten, string.Empty, invariant))
```
NightOwl888: We can change this one to `ReadOnlySpan<char>.Empty`.

NightOwl888 (approved):
Looks pretty good.

I left some comments, but since we will be making another pass to deal with the `ICharSequence` and `StringBuilder` overload removals where `CharSequence` is used upstream, we probably don't need to worry about them in this PR.

PR #1379, opened 2026-06-22
ci(performance): add prek hook priorities for parallel execution

jbampton:
`prek` now seems to run much faster and hooks are grouped.

Line 24 was switched to use prek same as Apache Airflow:

refs https://github.com/apache/airflow/blob/b788d83d215ec9fe6b289c3021a726997ca3d8a3/.pre-commit-config.yaml#L19

https://prek.j178.dev/proposals/concurrency/?h=priority

The priority is also used on Apache Lucene:

https://github.com/apache/lucene/blob/610484efbcd95c14dc133fd9ea8af409101fc7ef/.pre-commit-config.yml#L20

---

In the context of the [prek](https://github.com/j178/prek) Git hook runner, Priority 0 means a hook is placed in the highest tier of the parallel execution order.
Here is how priorities work in prek: 

* Parallel Execution: Unlike the traditional Python-based pre-commit tool, which runs hooks sequentially one by one, prek can run multiple hooks concurrently if they share the same priority. 
* Priority Levels: Hooks can be assigned different priority numbers. Hooks with a priority of 0 (or a higher numerical value, depending on configuration) run first or alongside other zero-priority hooks to decrease total end-to-end execution time. 
* Performance: Grouping non-overlapping, independent hooks into the same priority allows prek to utilize your CPU more efficiently.

---

# Implementation Plan - Update .pre-commit-config.yaml for prek concurrency

This plan proposes updating `.pre-commit-config.yaml` to include prek-style `priority` keys. This enables prek to run independent hooks in parallel, improving pre-commit check times.

## Proposed Changes

We will modify [`.pre-commit-config.yaml`](file:///Users/abe/Documents/github/lucenenet/.pre-commit-config.yaml) to add `priority` configuration to every hook. The priority values are assigned to structure the hook execution into logical stages, ensuring correct ordering of file modifications and maximum concurrency of read-only checks.

### Priority Groups and Ordering Design

| Stage / Priority | Hooks | Rationale |
|---|---|---|
| **Priority 10** | `identity`, `check-hooks-apply` | **Meta & Setup**: Ensure environment is valid before executing heavy checks. |
| **Priority 20** | `fix-byte-order-marker` | **File Cleanups (Broad)**: Fix BOM first so subsequent steps work with clean UTF-8 text. |
| **Priority 30** | `trailing-whitespace` | **Formatting (Broad)**: Cleans up whitespace on files before other hooks validate them. |
| **Priority 40** | `doctoc`, `oxipng`, `file-contents-sorter` | **Disjoint Modifiers**: These modify different sets of files (`README.md`, `.png` files, and `codespell.txt` respectively). Running them concurrently is safe. |
| **Priority 50** | `codespell`, `check-ast`, `check-builtin-literals`, `check-case-conflict`, `check-executables-have-shebangs`, `check-illegal-windows-names`, `check-json`, `check-merge-conflict`, `check-vcs-permalinks`, `check-xml`, `check-yaml`, `debug-statements`, `detect-aws-credentials`, `detect-private-key`, `forbid-submodules`, `gitleaks`, `bandit` | **Read-Only / Syntax / Security**: The bulk of the suite. These hooks only read and validate files without modifying them, allowing full parallel execution. |

---


paulirwin (changes requested):
I think this looks good. Can you please add the docs of the priority levels (from your PR description) as a comment to the top of this file, so that future editors know what the meanings of the priority levels are? Thanks!

paulirwin (approved):

PR #1311, opened 2026-05-28
Fix IndexWriter.Dispose leaking file handles, #1284

paulirwin:
- [X] You've read the [Contributor Guide](https://github.com/apache/lucenenet/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html).
- [X] You've included unit or integration tests for your change, where applicable.
- [X] You've included inline docs for your change, where applicable.
- [X] There's an open issue for the PR that you are making. If you'd like to propose a change, please [open an issue](https://github.com/apache/lucenenet/issues/new/choose) to discuss the change or find an existing issue.

Fix IndexWriter.Dispose leaking file handles by backporting fix from Lucene 4.10.0

Fixes #1284

## Description

See #1284 and https://github.com/apache/lucenenet/issues/1284#issuecomment-4361730485 for more information.

This PR backports https://github.com/apache/lucene/commit/2cfcdcc from Lucene 4.10.0 to fix this issue.

Additionally, CI surfaced increased random failures of `TestMaxMergeCount`, caused by a pre-existing latent porting issue rather than something new in this PR. The fix in this PR is upstream-faithful: `Shutdown(false)` (replacing the old `CloseInternal(false, true)` path) now triggers a fresh round of merges during shutdown that the old code skipped, which surfaces a long-dormant problem in how the test was translated. `CountdownEvent` is not a drop-in replacement for Java's `CountDownLatch`: `CountdownEvent.Signal()` throws if the count is already zero, whereas Java's `countDown()` silently does nothing past zero, and Lucene's test code relies on the Java semantics.

This PR adds a port of `CountDownLatch` from Apache Harmony, adapted to use `Interlocked` and `ManualResetEventSlim` like .NET's `CountdownEvent` does (which is simpler than the Harmony `Sync`-based implementation). All Harmony tests were ported; two are `[Ignore]`'d because we don't need thread interrupt support.

Issue #1284, opened 2026-04-30
IndexWriter.Dispose can leak file handles


paulirwin:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

`IndexWriter.CloseInternal` calls `CommitInternal()`, `ProcessEvents(false, true)`, `readerPool.DropAll(true)`, `deleter.Dispose()`, and `writeLock.Dispose()` in a single straight-line sequence with no `finally` block protecting the cleanup steps. If anything before `DropAll` throws (such as `CommitInternal` due to I/O failures) execution skips the cleanup and leaks:

- All `SegmentReader` references the writer's reader pool was holding (and through them, every `IndexInput`, `IndexInputSlicer`, `MemoryMappedFile`/`FileStream`, etc)
- The deleter
- The `write.lock` `FileStream` held by `NativeFSLock`

`SegmentCoreReaders.DecRef` consequently never reaches zero on the rooted readers, so `Dispose` is never called on the underlying file/slicer/mapping chain.

An example scenario where this was reproduced in tests (when `MMapDirectory` was incorrectly using `FileShare.Read` only, as part of #1267) is in `FreeTextSuggester.Build`, which deletes the temp index path after disposing the writer. Artificially modifying a directory implementation to only use `FileShare.Read` is one way to force an I/O error (a sharing violation in this case) for initial local reproduction. A focused unit test could substitute a `MockDirectoryWrapper` that throws from `Sync()` to simulate the failure, but would need to ensure that it doesn't strictly rely on file-sharing violations alone (that is just one example that spurred the creation of this issue, and is unlikely to occur once fixed in the code).

## Upstream comparison

The upstream code has the same straight-line structure. Java was possibly getting away with it because of either differences in our FS lock implementation, or the JVM finalizer for `FileChannel` being more aggressive than the .NET equivalent. Or, perhaps it is/was leaking handles too.

### Expected Behavior

No file handles are leaked on Dispose.

### Lucene.NET Version

4.8.0-beta00017

### Anything else?

Note that this might not have a real-world concern given that we match Java here, but that should be demonstrated by adding new test coverage for this scenario at a minimum.

paulirwin:
This is indeed a bug, confirmed via TDD tests. Raising priority to high.

[...]

PR #1382, opened 2026-06-23
Implement IsTestThread to fix flaky TestFlushExceptions

paulirwin:

Implement IsTestThread to fix flaky TestFlushExceptions

(Related to flaky failure on #1311 / #1284)

## Description

`LuceneTestCase.IsTestThread` was hardcoded to always return true, so the flush-failure injection in `TestConcurrentMergeScheduler.TestFlushExceptions` fired on background ConcurrentMergeScheduler merge threads (which also flush via `CompressingStoredFieldsWriter.Flush()`), not just the test thread. When a merge thread happened to be flushing during the SetDoFail/Flush window and `Random.NextBoolean()` returned true, the merge threw, surfacing as a flaky MergeException at teardown.

Capture the test-case thread in SetUp (cleared in TearDown) in a `[ThreadStatic]` field and have IsTestThread compare against it, mirroring Java's `ThreadAndTestNameRule.testCaseThread`. Tests run with `LevelOfParallelism(1)`, and NUnit runs SetUp, the test body, and TearDown on the same thread, so the capture lines up with the thread executing the test method. Background merge threads never run SetUp, so they correctly see a null test thread and are excluded.

### DIFF
--- a/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
+++ b/src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
@@ -896,6 +896,11 @@ namespace Lucene.Net.Util
         {
             // LUCENENET TODO: Not sure how to convert these
             //ParentChainCallRule.SetupCalled = true;
+
+            // LUCENENET specific: capture the thread running the test method so that
+            // IsTestThread can distinguish it from background threads (e.g. ConcurrentMergeScheduler
+            // merge threads). This mirrors Java's ThreadAndTestNameRule.testCaseThread.
+            testCaseThread = Thread.CurrentThread;
         }
 
         /// <summary>
@@ -982,6 +987,11 @@ namespace Lucene.Net.Util
                 NUnit.Framework.TestContext.Error.WriteLine($"[ERROR] An exception occurred during TearDown:");
                 RandomizedContext.PrintStackTrace(ex, NUnit.Framework.TestContext.Error);
                 throw; // LUCENENET: Throw to preserve stack details of original throw.
+            }
+            finally
+            {
+                // LUCENENET specific: clear the captured test thread (see SetUp()).
+                testCaseThread = null;
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
```diff
@@ -1147,15 +1157,23 @@ public static SegmentReader GetOnlySegmentReader(DirectoryReader reader)
...
         /// executing the test case.
-        /// <para/>
-        /// LUCENENET: Not Implemented - always returns true
         /// </summary>
-        /*Assert.IsNotNull(ThreadAndTestNameRule.TestCaseThread, "Test case thread not set?");
-                return Thread.CurrentThread == ThreadAndTestNameRule.TestCaseThread;*/
-        internal static bool IsTestThread => true; // LUCENENET specific - changed from public to internal since there is no way to support it
+        // LUCENENET specific - changed from public to internal since IsTestThread is only meaningful
+        // within the test framework. Mirrors Java's check against ThreadAndTestNameRule.testCaseThread.
+        internal static bool IsTestThread => testCaseThread != null && Thread.CurrentThread == testCaseThread;
```
NightOwl888: Since `Thread.CurrentThread` can never be `null` and this is a reference equality check, the `== null` case will simply be `false`. So, there is no reason to check whether `testCaseThread` is `null` before comparing it.

Note that I made an attempt to do this on another branch, but I was also considering async tasks. Also note that this is one of the unfinished features listed in #1017. Ideally, we would make this public to make it available to users, but if we aren't going to do that, we probably don't need to worry about supporting async.

NightOwl888 (approved):
Approved with optional additional support.

src/Lucene.Net.TestFramework/Util/LuceneTestCase.cs
paulirwin: Well, would you like me to abandon this PR? I think supporting async tests could be worth it.

NightOwl888: The `async` attempt I made didn't work IIRC. This is an improvement over what we have now, and since you are not declaring `IsTestThread` public, we can make a later change to it without affecting users.

Issue #1338, opened 2026-06-11
Remove `IAppendable` and `ISpanAppendable` from `ICharTermAttribute` and buffer types


paulirwin:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Task description

Follow-up to #1315 (step 4-5 of the plan in https://github.com/apache/lucenenet/issues/1315#issuecomment-4683039760). Depends on #1337 landing first.

### Phase 1 - Migrate callers (non-breaking)
Audit and prep all callers still relying on `IAppendable` so they use `IBufferWriter<char>` or the concrete `Append()` methods instead:
- Includes the ICU4N integration in `ICUNormalizer2Filter`. **This step is partly gated on an ICU4N release** that ships an `IBufferWriter<char>` overload.
- Most callers can call `Append()` on the concrete buffer types without needing the interface at all.

### Phase 2 - Remove the interfaces (breaking)
Remove `IAppendable` and `ISpanAppendable` from `ICharTermAttribute`, `OpenStringBuilder`, and any other buffer types. The `Append(ReadOnlySpan<char>)` overload added in #1315 means callers no longer need `ISpanAppendable`'s extension-method/type-check path.

Note: per the discussion, `IAppendable` is less problematic to keep around than `ICharSequence` if some corner case turns up that still needs append semantics, so validate during Phase 1 that nothing genuinely requires it.

This is a breaking change and should be done in a beta release before 4.x final.

paulirwin:
This item is also blocked on ICU4N support: https://github.com/NightOwl888/ICU4N/issues/137

PR #1387, opened 2026-06-23
SWEEP: Removal of ICharSequence from most places, #1337

### LINKED ISSUE #1337: Remove `ICharSequence` from `ICharTermAttribute` and buffer types

### Task description

Follow-up to #1315 (step 2-3 of the plan in https://github.com/apache/lucenenet/issues/1315#issuecomment-4683039760).

Now that `IBufferWriter<char>` and `Append(ReadOnlySpan<char>)` are going to be in place (as of #1315), we can move off `ICharSequence` on the buffer types.

### Phase 1 - Rework callers (non-breaking)
Replace internal logic that consumes `ICharSequence` (from `CharTermAttribute`, `OpenStringBuilder`, and other buffer types) with `ReadOnlySpan<char>` or `ReadOnlyMemory<char>` where appropriate. This is purely internal call-site refactoring and must land before the removal below.

### Phase 2 - Remove the implementation (breaking)
Remove the `ICharSequence` implementation from `ICharTermAttribute` and all buffer types at the same time:
- Drop `SubSequence()` and `HasValue`.
- Promote any still-needed members directly onto `ICharTermAttribute`.

### Known consequence
Because the compiler does not honor a concrete type's implicit operators through an interface, callers will need to call `.AsSpan()` explicitly on `ICharTermAttribute` to pass it into `ReadOnlySpan<char>` APIs. On the upside, once `ICharSequence` is gone there is no risk of accidentally routing a value into an `ICharSequence` API when `ReadOnlySpan<char>` is the better choice.

This is a breaking change and should be done in a beta release before 4.x final.

paulirwin:

Removal of ICharSequence from most places.

Fixes #1337

## Description

ICharSequence is not necessary in the era of `ReadOnlySpan<char>`. This sweep removes most references to it, leaving only those for IAppendable support (falling back to a more-expensive ToString operation, to be removed in #1338) and in the query node types (that have custom implementations that should be done separately. Where possible, it was replaced with ReadOnlySpan<char>. Some tests had to be adapted with AsSpan/Slice to mimic the prior behavior.

This is a breaking change because of the removal of overloads that can affect non-modern-.NET/C#14 users.

### DIFF
--- a/src/Lucene.Net.Analysis.Common/Analysis/Compound/CompoundWordTokenFilterBase.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Compound/CompoundWordTokenFilterBase.cs
@@ -1,13 +1,14 @@
 // Lucene version compatibility level 4.8.1
-using J2N.Text;
 using Lucene.Net.Analysis.TokenAttributes;
 using Lucene.Net.Analysis.TokenAttributes.Extensions;
 using Lucene.Net.Analysis.Util;
 using Lucene.Net.Diagnostics;
 using Lucene.Net.Util;
-using Lucene.Net.Support;
 using System;
 using System.Collections.Generic;
+#if !FEATURE_QUEUE_TRYDEQUEUE_TRYPEEK
+using Lucene.Net.Support;
+#endif
 
 namespace Lucene.Net.Analysis.Compound
 {
@@ -115,7 +116,7 @@ namespace Lucene.Net.Analysis.Compound
             {
                 if (Debugging.AssertsEnabled) Debugging.Assert(current != null);
                 RestoreState(current); // keep all other attributes untouched
-                m_termAtt.SetEmpty().Append(token.Text);
+                m_termAtt.SetEmpty().Append(token.Text.Span);
                 m_offsetAtt.SetOffset(token.StartOffset, token.EndOffset);
                 posIncAtt.PositionIncrement = 0;
                 return true;
@@ -161,10 +162,10 @@ namespace Lucene.Net.Analysis.Compound
         /// </summary>
         protected class CompoundToken
         {
-            private readonly ICharSequence txt;
+            private readonly ReadOnlyMemory<char> txt;
             private readonly int startOffset, endOffset;
 
-            public ICharSequence Text => txt; // LUCENENET specific: changed public field into property backed by private field
+            public ReadOnlyMemory<char> Text => txt; // LUCENENET specific: changed public field into property backed by private field
 
             public int StartOffset => startOffset; // LUCENENET specific: changed public field into property backed by private field
 
@@ -174,7 +175,7 @@ namespace Lucene.Net.Analysis.Compound
             /// Construct the compound token based on a slice of the current <see cref="CompoundWordTokenFilterBase.m_termAtt"/>. </summary>
             public CompoundToken(CompoundWordTokenFilterBase compoundWordTokenFilterBase, int offset, int length)
             {
-                this.txt = compoundWordTokenFilterBase.m_termAtt.Subsequence(offset, length); // LUCENENET: Corrected 2nd Subsequence parameter
+                this.txt = compoundWordTokenFilterBase.m_termAtt.AsMemory(offset, length); // LUCENENET: Corrected 2nd Subsequence parameter
 
                 // offsets of the original word
                 int startOff = compoundWordTokenFilterBase.m_offsetAtt.StartOffset;
--- a/src/Lucene.Net.Analysis.Common/Analysis/Util/CharArrayMap.cs
+++ b/src/Lucene.Net.Analysis.Common/Analysis/Util/CharArrayMap.cs
@@ -489,6 +489,18 @@ namespace Lucene.Net.Analysis.Util
         /// <exception cref="ArgumentNullException"><paramref name="text"/> is <c>null</c>.</exception>
         public virtual bool ContainsKey(string text)
         {
+            if (text is null)
+                throw new ArgumentNullException(nameof(text));
+
+            return keys[GetSlot(text)] != null;
+        }
+
+        /// <summary>
+        /// <c>true</c> if the <paramref name="text"/> <see cref="ReadOnlySpan{T}"/> is in the <see cref="Keys"/>;
+        /// otherwise <c>false</c>
+        /// </summary>
+        public virtual bool ContainsKey(ReadOnlySpan<char> text)
+        {
             return keys[GetSlot(text)] != null;
         }
 
@@ -623,6 +635,9 @@ namespace Lucene.Net.Analysis.Util
         /// <exception cref="KeyNotFoundException"><paramref name="text"/> is not found in the dictionary.</exception>
         internal virtual TValue Get(string text, bool throwIfNotFound = true)
         {
+            if (text is null)
+                throw new ArgumentNullException(nameof(text));
+
             var value = values[GetSlot(text)];
             if (value is not null)
             {
@@ -714,7 +729,7 @@ namespace Lucene.Net.Analysis.Util
         /// Returns <c>true</c> if the <see cref="string"/> is in the set.
         /// </summary>
         /// <exception cref="ArgumentNullException"><paramref name="text"/> is <c>null</c>.</exception>
-        private int GetSlot(string text)
+        private int GetSlot(ReadOnlySpan<char> text)
         {
             int code = GetHashCode(text);
             int pos = code & (keys.Length - 1);
@@ -808,6 +823,9 @@ namespace Lucene.Net.Analysis.Util
         /// <exception cref="ArgumentNullException"><paramref name="text"/> is <c>null</c>.</exception>
         public virtual bool Put(string text, TValue value, [MaybeNullWhen(returnValue: true)] out TValue previousValue) // LUCENENET: Refactored to use out value to support value types
         {
+            if (text is null)
+                throw new ArgumentNullException(nameof(text));
+
             MapValue? oldValue = PutImpl(text, new MapValue(value));
             if (oldValue is not null)
             {
@@ -930,9 +948,9 @@ namespace Lucene.Net.Analysis.Util
             // than casting using *as* and then checking for null.
             // http://stackoverflow.com/q/1583050/181087
             if (text is string str)
-                return PutImpl(str, value);
+                return PutImpl(str.AsSpan(), value);
             if (text is char[] charArray)
-                return PutImpl(charArray, value);
+                return PutImpl(charArray.AsSpan(), value);
             if (text is ICharSequence cs)
                 return PutImpl(cs, value);
 
@@ -946,18 +964,13 @@ namespace Lucene.Net.Analysis.Util
         /// <summary>
         /// Add the given mapping.
         /// </summary>
-        /// <exception cref="ArgumentNullException"><paramref name="text"/> is <c>null</c>.</exception>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private MapValue? PutImpl(string text, MapValue value)
+        private MapValue? PutImpl(ReadOnlySpan<char> text, MapValue value)
         {
-            // LUCENENET: Added guard clause
-            if (text is null)
-                throw new ArgumentNullException(nameof(text));
-
             // LUCENENET specific - only allocate char array if it is required.
             if (ignoreCase)
             {
-                return PutImpl(text.ToCharArray(), value);
+                return PutImpl(text.ToArray(), value);
             }
             version++;
             int slot = GetSlot(text);
@@ -967,7 +980,7 @@ namespace Lucene.Net.Analysis.Util
                 values[slot] = value;
                 return oldValue;
             }
-            keys[slot] = text.ToCharArray();
+            keys[slot] = text.ToArray();
             values[slot] = value;
             count++;
 
@@ -1100,6 +1113,16 @@ namespace Lucene.Net.Analysis.Util
         }
 
         /// <summary>
+        /// Sets the value of the mapping of the chars inside this <see cref="ReadOnlySpan{T}"/>.
+        /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="text"/> is <c>null</c>.</exception>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal virtual void Set(ReadOnlySpan<char> text)
+        {
+            SetImpl(text, PLACEHOLDER);
+        }
+
+        /// <summary>
         /// Sets the value of the mapping of the chars inside this <see cref="string"/>.
         /// </summary>
         /// <exception cref="ArgumentNullException"><paramref name="text"/> is <c>null</c>.</exception>
@@ -1154,7 +1177,7 @@ namespace Lucene.Net.Analysis.Util
         void ICharArrayDictionary.Set(char[] text) => Set(text);
         void ICharArrayDictionary.Set(ICharSequence text) => Set(text);
         void ICharArrayDictionary.Set<T>(T text) => Set(text);
-        void ICharArrayDictionary.Set(string text) => Set(text);
+        void ICharArrayDictionary.Set(ReadOnlySpan<char> text) => Set(text);
 
         #endregion Set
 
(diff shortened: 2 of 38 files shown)


src/Lucene.Net.Analysis.Common/Analysis/Compound/CompoundWordTokenFilterBase.cs
```diff
@@ -174,7 +175,7 @@ protected class CompoundToken
             /// Construct the compound token based on a slice of the current <see cref="CompoundWordTokenFilterBase.m_termAtt"/>. </summary>
             public CompoundToken(CompoundWordTokenFilterBase compoundWordTokenFilterBase, int offset, int length)
             {
-                this.txt = compoundWordTokenFilterBase.m_termAtt.Subsequence(offset, length); // LUCENENET: Corrected 2nd Subsequence parameter
+                this.txt = compoundWordTokenFilterBase.m_termAtt.AsMemory(offset, length); // LUCENENET: Corrected 2nd Subsequence parameter
```
NightOwl888, line 178: This is a fundamentally different operation than upstream. There are 2 different problems here:

1. `Subsequence()` works similarly to `Substring()` - that is, it creates a new chunk of memory that is independent of the original and copies the chars from the original memory location into it.
2. Storing `ReadOnlyMemory<char>` here is dangerous without also storing a reference to the class that is behind it. In ICU4N, we use [a helper method](https://github.com/NightOwl888/ICU4N/blob/main/src/ICU4N/Support/MemoryExtensions.cs#L157-L181) and `object` field to ensure the backing type is kept in scope.

So, the new operation adds a dependency on the stability of the backing array of `m_termAtt`. And since it is protected, there are no guarantees there.

Usually, when I encounter a type like this, there are a couple of questions to ask:

1. Is this type suitable for converting into a ref struct?
2. How long is the memory kept in scope until it is garbage collected?

Keep in mind, there is nothing inherently bad about allocating heap. The problems occur when that heap is allocated, used only for a very short time, and then left for the GC to clean up. In this case, after the memory is allocated, it is added to a queue for later processing. So, more analysis is needed to determine how quickly that queue cycles its memory.

One thing I noticed that could be impactful is the fact that upstream the `queue` was a linked list and we have changed it to be a `Queue<T>`. When resetting, both of them clear out all of their memory for garbage collection and `Queue<T>` potentially reuses its backing array, while `LinkedList<T>` would not allocate a backing array at all. So, it would be interesting to benchmark to work out if either of them would be a clear winner.

There are a couple of ways I can think of that would potentially improve reuse, but both come with some extra complexity.

1. Have `CompoundToken` implement `IDisposable` and back it with chunks of memory from the shared array pool.
2. Allocate a single chunk of heap in the `CompoundWordTokenFilterBase` and have each `CompoundToken` instance use a slice of it until it is full, after which new chunks are allocated for each token.

Combining both of these might be a 3rd option.

Failing any of those, the minimal safe way to remove `ICharSequence` here is to convert the backing field into a `string` that is allocated in the constructor. It is essentially the same thing we have now, but without the baggage of `CharArrayCharSequence`.

src/Lucene.Net.Analysis.Common/Analysis/Util/CharArrayMap.cs
```diff
@@ -299,28 +270,6 @@ public virtual void Add(char[] text, TValue value)
...
-        /// <param name="value">The corresponding value for the given <paramref name="text"/>.</param>
-        /// <exception cref="ArgumentNullException">
-        /// <paramref name="text"/> is <c>null</c>.
-        /// <para/>
-        /// -or-
-        /// <para/>
-        /// The <paramref name="text"/>'s <see cref="ICharSequence.HasValue"/> property returns <c>false</c>.
-        /// </exception>
-        /// <exception cref="ArgumentException">An element with <paramref name="text"/> already exists in the dictionary.</exception>
-        public virtual void Add(ICharSequence text, TValue value)
```
NightOwl888, line 315: In this case, we can and should convert this to a `ReadOnlySpan<char>` overload, since it is missing one, although, it has all of the support to implement one. It internally converts the span to an array, but since it is immediately adding it to a collection, this isn't as bad as allocating a temporary array that immediately is left for garbage collection.

src/Lucene.Net.Analysis.Common/Analysis/Util/CharArrayMap.cs
```diff
@@ -930,11 +740,9 @@ public virtual bool Put<T>(T text, TValue value, [MaybeNullWhen(returnValue: tru
             // than casting using *as* and then checking for null.
             // http://stackoverflow.com/q/1583050/181087
             if (text is string str)
-                return PutImpl(str, value);
+                return PutImpl(str.AsSpan(), value);
             if (text is char[] charArray)
-                return PutImpl(charArray, value);
-            if (text is ICharSequence cs)
-                return PutImpl(cs, value);
+                return PutImpl(charArray.AsSpan(), value);
```
NightOwl888, line 745: Was copying the array intentional? The original implementation was set up to put the same instance as the one passed in, not a copy. I am not saying that was right, but this is a difference in behavior.

src/Lucene.Net.Facet/Taxonomy/WriterCache/CharBlockArray.cs
```diff
@@ -176,160 +177,12 @@ public virtual CharBlockArray Append(ICharSequence? value, int startIndex, int l
             if (startIndex > value.Length - length)
                 throw new ArgumentOutOfRangeException(nameof(startIndex), $"Index and length must refer to a location within the string. For example {nameof(startIndex)} + {nameof(length)} <= {nameof(Length)}.");
 
-
             int end = startIndex + length;
             for (int i = startIndex; i < end; i++)
```
NightOwl888, line 181: This slow implementation made sense for `ICharSequence`, but it doesn't here.

That said, it would be better to make this type implicitly convertible to `ReadOnlySpan<char>` and change the `ReadOnlySpan<char>` overload to be safe for inserting the same memory location. I recently did that in `MutableTextBuffer` - here are the links.

https://github.com/NightOwl888/J2N/blob/a9c6092b14bbbbca6c0f842dcee7613582fcc4f2/src/J2N/Text/MutableTextBuffer.TextSources.cs#L206-L228
https://github.com/NightOwl888/J2N/blob/a9c6092b14bbbbca6c0f842dcee7613582fcc4f2/src/J2N/Text/MutableTextBuffer.TextSources.cs#L1039-L1076

This would transform it into a bulk operation that uses SIMD optimizations rather than a slow stream of adding one character at a time.

Note that we can also remove the `startIndex`/`length` overload if we add `AsSpan()` extension methods so we can slice this type.

src/Lucene.Net.Tests.Facet/Taxonomy/WriterCache/TestCharBlockArray.cs
```diff
@@ -194,7 +194,7 @@ public virtual void TestAppendableInterface()
 
             try
             {
-                t.Append((string)t2.ToString(), 1, 0 - 1); // LUCENENET: Corrected 3rd parameter
+                t.Append(t2.ToString().AsSpan(1, 0 - 1)); // LUCENENET: Corrected parameter
```
NightOwl888, line 197: This is an odd choice. Why not just call `t2.AsSpan(1, 0 - 1)`?

Note this also applies to the `TestCharTermAttributeImpl` class where this was poached from.

src/Lucene.Net.Tests.Facet/Taxonomy/WriterCache/TestCharBlockArray.cs
```diff
@@ -210,7 +210,7 @@ public virtual void TestAppendableInterface()
             // LUCENENET specific - test char[] overloads
             try
             {
-                t.Append((char[])t2.ToString().ToCharArray(), 1, 5 - 1); // LUCENENET: Corrected 3rd parameter
+                t.Append(t2.ToString().ToCharArray().AsSpan(1, 5 - 1)); // LUCENENET: Corrected parameter
```
NightOwl888, line 213: This is an odd choice. Why not just call `t2.AsSpan(1, 0 - 1)`?

Note this also applies to the `TestCharTermAttributeImpl` class where this was poached from.

src/Lucene.Net.Tests.Facet/Taxonomy/WriterCache/TestCharBlockArray.cs
```diff
@@ -221,7 +221,7 @@ public virtual void TestAppendableInterface()
 
             try
             {
-                t.Append((char[])t2.ToString().ToCharArray(), 1, 0 - 1); // LUCENENET: Corrected 3rd parameter
+                t.Append(t2.ToString().ToCharArray().AsSpan(1, 0 - 1)); // LUCENENET: Corrected parameter
```
NightOwl888, line 224: Looks like there are several more cases where `.ToCharArray().AsSpan()` are chained together in this class, which seems odd.

Note this also applies to the TestCharTermAttributeImpl class where this was poached from.

src/Lucene.Net/Analysis/Token.cs
```diff
@@ -448,47 +442,11 @@ private void ClearNoTermBuffer()
...
-        /// <summary>
-        /// Shorthand for calling <see cref="Clear"/>,
-        /// <see cref="ICharTermAttribute.CopyBuffer(char[], int, int)"/>,
-        /// <see cref="SetOffset"/>,
-        /// <see cref="Type"/> (set) on <see cref="TypeAttribute.DEFAULT_TYPE"/> </summary>
-        /// <returns> this <see cref="Token"/> instance  </returns>
-        public virtual Token Reinit(char[] newTermBuffer, int newTermOffset, int newTermLength, int newStartOffset, int newEndOffset)
-        {
-            CheckOffsets(newStartOffset, newEndOffset);
-            ClearNoTermBuffer();
```
NightOwl888, line 477: This is a bit odd. These methods existed upstream as they were here. They didn't use `CharSequence`. And now that these 2 overloads have been removed, there are no callers to the `ClearNoTermBuffer()` method. What is odd is that there are no callers to any of these methods. So, it isn't clear how they were intended to be used or whether excluding these overloads that have different clear behavior would be breaking behavior for any callers.

Maybe the safest thing to do would be to revert this class in this PR, add tests for it in a separate PR, and then decide whether we can make this sort of change.

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttributeImpl.cs
```diff
@@ -166,101 +164,14 @@ public char this[int index]
...
-            if (startIndex > value.Length - charCount)
-                throw new ArgumentOutOfRangeException(nameof(startIndex), $"Index and length must refer to a location within the string. For example {nameof(startIndex)} + {nameof(charCount)} <= {nameof(Length)}.");
-
-            Arrays.Copy(value, startIndex, InternalResizeBuffer(termLength + charCount), termLength, charCount);
-            Length += charCount;
-
-            return this;
-        }
-
-        public CharTermAttribute Append(string value)
```
NightOwl888, line 259: This overload exists upstream. Despite the fact that `string` can be implicitly converted to `ReadOnlySpan<char>` it is still more efficient for it to have its own overload than to rely only on `ReadOnlySpan<char>` for this common case. Given that this is a high use user API, we should provide the direct overloads to optimize it.

That being said, the upstream implementation was also better than it was here. We should effectively copy the `StringBuilder` overload implementation, since using `string.CopyTo()` is the most efficient way to transfer the characters into the local buffer.

It should also be made nullable to indicate that it won't throw if the value is `null`, instead it will be a no-op.

It probably also makes sense to have the `char[]` overloads even though they don't exist upstream for the same reason.

The plan is to remove the `IAppendable` interface, not the implementation.

src/Lucene.Net/Analysis/TokenAttributes/CharTermAttribute.cs
```diff
@@ -94,146 +94,12 @@ public interface ICharTermAttribute : IAttribute, ICharSequence, IAppendable,
...
-        /// This method uses .NET semantics. In Lucene, a <c>null</c> <paramref name="value"/> would append the
-        /// string <c>"null"</c> to the instance, but in Lucene.NET a <c>null</c> value will be safely ignored.
-        /// </para>
-        /// </summary>
-        /// <param name="value">The sequence of characters to append.</param>
-        /// <remarks>
-        /// LUCENENET specific method, added because the .NET <see cref="string"/> data type
-        /// doesn't implement <see cref="ICharSequence"/>.
-        /// </remarks>
-        new ICharTermAttribute Append(string value);
```
NightOwl888, line 203: See my message in the `CharTermAttributeImpl.cs` file. We should keep supporting the overloads for specific data types because converting to `ReadOnlySpan<char>` is still less efficient than having a `string`, `char[]`, overload, etc.

src/Lucene.Net.Analysis.Common/Analysis/Util/CharArrayMap.cs
```diff
@@ -440,7 +389,7 @@ private void CopyTo(KeyValuePair<string, TValue>[] array, int arrayIndex)
         /// <exception cref="ArgumentOutOfRangeException"><paramref name="index"/> is less than zero.</exception>
         /// <exception cref="ArgumentException">The number of elements in the source is greater
         /// than the available space from <paramref name="index"/> to the end of the destination array.</exception>
-        internal void CopyTo(KeyValuePair<ICharSequence, TValue>[] array, int index) // internal for testing
+        internal void CopyTo(KeyValuePair<ReadOnlyMemory<char>, TValue>[] array, int index) // internal for testing
```
NightOwl888, line 392: `ReadOnlyMemory<char>` doesn't own the pointer that keeps the memory in scope. Cloning the `KeyValuePair` is a bit dangerous because the caller must be aware when the source of the memory backing it goes out of scope.

Note in ICU4N I created a `CharSequence` type to be used in collections that owns both the `ReadOnlyMemory<char>` instance and a reference to the memory to keep it in scope.

However, we should analyze this further.

1. We should remove all of the code that accepts arbitrary objects that are converted using formatting (generic overloads and numeric overloads).
2. Now that we have alternate lookup implemented in all target frameworks, it would probably be better to back this by a `Dictionary<string, TValue>`. Then the lookup can either be `string` or `ReadOnlySpan<char>` (which covers `char[]` and `TextBuilder`).

Anyway, that doesn't have to be done in this PR, but we should probably be less eager to use `ReadOnlyMemory<char>` due to its inability to manage its own scope.

NightOwl888 (changes requested):
Thanks for the PR. I found several issues that should be addressed here, but decided not to completely review `CharArraySet` and `CharArrayDictionary<TValue>` because I think we should roll them back and take them on as a separate issue by themselves.

There were a couple of assumptions made throughout this PR that are incorrect:

1. We should not be replacing `string` overloads that exist upstream with `ReadOnlySpan<char>`. There is a cost to converting `string` to `ReadOnlySpan<char>` and it only makes sense if it is recouped by calling methods that use SIMD and/or providing some memory reuse benefit to make up the cost. Typically, the duplication of simple business logic is justified on user-heavy APIs like `CharTermAttribute` where lower-level APIs can generally be normalized to `ReadOnlySpan<char>` to reduce duplicate business logic and take advantage of vectorization and SIMD.
2. We will not be keeping the `System.Text.StringBuilder` overloads that we added to sidestep the need for `ICharSequence`. `TextBuilder` can be converted directly to `ReadOnlySpan<char>` so we now have a string builder that we can pass directly to `ReadOnlySpan<char>` APIs. The only exceptions I would make to this rule are in public facing buffers where the source may be a `System.Text.StringBuilder` from an end user and the destination is the buffer itself. However, internally we should eliminate these APIs and use either `ValueStringBuilder` (preferred) or `TextBuilder`.

That said, I think we can accept most of this, but I would recommend that we do the following:

1. Restore the `string`, `char[]` as well as sliced `string` and `char[]` overloads to `CharTermAttribute`, `ICharTermAttribute`, and `CharBlockArray` (and optimize them better than they were before this PR).
2. Revert all of the changes to `CharArrayDictionary<TValue>` and `CharArraySet`.
3. Revert the changes to the phonetic `bm` namespace.

It would probably also make sense to roll back the `CompoundTokenFilterBase`, since this "direct port" can be defeated quite easily by a user by clearing or replacing the memory in the middle of the process. This would corrupt the state of the analyzer.

### `CharArraySet` and `CharArrayDictionary<TValue>`

There are 2 separate issues that we need to address on these classes:

1. We should remove the methods that use object-to-string conversion (generic overloads and numeric formatting overloads). This "feature" didn't exist upstream and it is not likely to be very useful.
2. We should seriously consider keeping only the public API surface from Lucene and changing these classes to be backed by `Dictionary<string, TValue>` and `HashSet<string>` and use the newly backported alternate lookup functionality to allow looking up with anything that can be converted to `ReadOnlySpan<char>` (such as a `TextBuilder`).

I will be making issues for these with more details.

### Phonetic Algorithms

This code was ported from the Commons Codec project and has many design issues because it fundamentally revolved around the `ICharSequence` interface. We should take a fresh look at redesigning it so that it doesn't rely on `System.Text.RegularExpressions` or `ICharSequence` and instead operate at a lower level, even at the cost of reducing extensibility.

I will also add an issue for this, but there are some decisions to make first - for example, should we aim to make or find a package to replace Commons Codec? Note that we have had a request to make these algorithms available outside of Lucene.NET. However, I don't think we should do a direct port of the API because much like ICU4N, it takes a high-level object-oriented approach to solve low-level problems like tracking grapheme clusters. These should definitely be geared toward System.Memory and away from higher-level APIs such as regular expressions and collections.

------------------

Note that this PR is now blocking the rest of the work of restoring `CharSequence` overloads as System.Memory types, which I will be opening issues for to divide up the work into managable chunks.

Issue #1209, opened 2025-10-10
Task: Research whether we can use Lucene/Lucene.NET book images


NightOwl888:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Task description

We ran into a snag while moving the book images over to our website in #1185. We have never asked for permission to use these images, and we need to research the Apache policies before requesting permission to determine whether it is even something we can do.

For now, we have left the broken image links in place, but perhaps we should consider commenting out the links until we know for sure that we can or cannot use these images.

Issue #1183, opened 2025-09-15
Meet Apache Website requirements


paulirwin:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description of Bug

We do not meet several requirements Apache has for top-level project websites. The full list can be seen here: https://whimsy.apache.org/site/project/lucenenet

This includes the original bug report below, about missing images. All external resources should be hosted on Apache servers.

### Original Bug Report

The "Popular Books" images on the homepage are broken, seemingly as a result of a CSP change at Amazon.

<img width="1846" height="910" alt="Image" src="https://github.com/user-attachments/assets/6ee1d598-37a2-4d40-a34d-bda1ed00c94d" />

### Expected Behavior

The images appear (or the section is otherwise corrected, perhaps with text links)

### Steps To Reproduce

Visit the https://lucenenet.apache.org home page.

vwilson:
Good afternoon!  The images themselves still exist on amazon's servers. This is a CSP issue. There are several additional issues on the index page - academicons.min.css, font-awesome.min.css, and the roboto mono font. 

I think this may need to be addressed by the Apache infra team. (This is my first attempt at contributing to the repo so please forgive my lack of familiarity.)

NightOwl888:
Being that dependency rot is an ongoing issue with both websites (in addition to GitHub Actions and our NuGet dependencies), perhaps we should look into setting it up so we can automatically upgrade them with [Renovate](#1187).

- [JavaScript](https://docs.renovatebot.com/javascript/)
- [Node.js](https://docs.renovatebot.com/node/)

This will allow us to get out in front of many security problems before someone reports them.

Currently the only way to test our websites after being generated is to host them locally, go through the DocFx logs, and click links manually to see if they work. So, even if we automate the upgrades, it would take a bit of work to review the PRs from Renovate. We potentially could also set up an automated testing workflow that does some checks on the static websites that are generated. Or at the very least, set it up so when a PR with a website change is submitted, the website in the PR is built and hosted on a temporary URL so we can preview how the changes affect it without leaving the browser.

--------------------

One more thought: We are only 10 days away from the start of [Hacktoberfest](https://hacktoberfest.com/). It would be a good idea to divide up some of the smaller tasks (less than ~30 minutes of work) on the website into "good first issue" grooming for potential contributors, which may help to encourage participation.

Perhaps we could also pin a discussion up that links to the "good first issue" URLs and the documentation for how to build the website.

[...]

PR #1389, opened 2026-06-24
website: add HTML img `alt` text for both homepage books

jbampton:
Also cleaned up the HTML

- removed unneeded blank lines
- indented markup


NightOwl888 (commented):
Thanks for the PR. 

This one is kind of strange. Currently, Apache is blocking these images from appearing on our website, because they don't allow us to use externally hosted files (#1183). But we also don't have permission to copy these images onto our site - see #1209. So, we have some research to do to figure out how to deal with these.

In the meantime, it would probably be best to leave them here, but make this section invisible so we don't show broken image links on our homepage or a title with no content below it.

PR #1404, opened 2026-07-06
misc: fix typos

jbampton:
General spelling clean up

### DIFF
--- a/src/Lucene.Net.Tests.Facet/Taxonomy/TestTaxonomyCombined.cs
+++ b/src/Lucene.Net.Tests.Facet/Taxonomy/TestTaxonomyCombined.cs
@@ -412,7 +412,7 @@ namespace Lucene.Net.Facet.Taxonomy
                 }
             }
             // (also test invalid categories:)
-            Assert.AreEqual(TaxonomyReader.INVALID_ORDINAL, tr.GetOrdinal(new FacetLabel("non-existant")));
+            Assert.AreEqual(TaxonomyReader.INVALID_ORDINAL, tr.GetOrdinal(new FacetLabel("non-existent")));
             Assert.AreEqual(TaxonomyReader.INVALID_ORDINAL, tr.GetOrdinal(new FacetLabel("Author", "Jules Verne")));
 
             tr.Dispose();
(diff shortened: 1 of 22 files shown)


src/Lucene.Net.Tests.Facet/Taxonomy/TestTaxonomyCombined.cs
```diff
@@ -412,7 +412,7 @@ public virtual void TestReaderBasic()
                 }
             }
             // (also test invalid categories:)
-            Assert.AreEqual(TaxonomyReader.INVALID_ORDINAL, tr.GetOrdinal(new FacetLabel("non-existant")));
+            Assert.AreEqual(TaxonomyReader.INVALID_ORDINAL, tr.GetOrdinal(new FacetLabel("non-existent")));
```
NightOwl888, line 416: Since we are changing the test condition from [upstream](https://github.com/apache/lucene/blob/releases/lucene-solr/4.8.1/lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyCombined.java#L377), please add the following comment above this line:

```c#
// LUCNENET specific: Fixed the spelling of the word non-existent (upstream was non-existant)
```

NightOwl888 (changes requested):
Thanks for the PR.

Looks good except for the one test condition change. But being that the intent of the test hasn't changed and the test passes, I think this is fine as long as we add a comment to explain why we diverged from Lucene. Please see the inline comment.

src/Lucene.Net.Tests.Facet/Taxonomy/TestTaxonomyCombined.cs
paulirwin: @NightOwl888 a helpful tip... you can use "```suggestion" markdown fences to give code suggestions and then it makes it easy for the other person to apply them as a batch, as well as see it as a diff:

```suggestion
            // LUCENENET specific: Fixed the spelling of the word non-existent (upstream was non-existant)
            Assert.AreEqual(TaxonomyReader.INVALID_ORDINAL, tr.GetOrdinal(new FacetLabel("non-existent")));
```

This can automatically be inserted for you with this button in the comment editor:
Image

NightOwl888 (approved):

Issue #1381, opened 2026-06-23
Backport remaining Lucene 4.8.1 bug fixes


paulirwin:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Task description

We have effectively been targeting 4.8.1 instead of 4.8.0 for a while now, so we might as well close that gap before exploring #1293. This release includes [some worthwhile bugfixes](https://lucene.apache.org/core/4_8_1/changes/Changes.html).

What follows is the analysis by Claude Code against the Lucene 4.8.1 source code and git history. There may be errors/omissions.

## Background

Lucene.NET targets Lucene 4.8.1, but an audit of the upstream `4.8.0 → 4.8.1`
diff (the 16 `LUCENE-*` bug fixes in `lucene/CHANGES.txt`) found that several
fixes were never ported. No existing issue tracks these. #1293 is explicitly
scoped to fixes *after* 4.8.1 (4.9-4.10), so this gap is currently untracked.

The fixes below match pre-fix 4.8.0 behavior in current `master`. They are
grouped into batches that can be broken out into sub-issues. Several are
trivial; one (LUCENE-5628) fixes a real `StackOverflowException` that is
**already biting the port** (`AnalyzingSuggesterTest.TestTooLongSuggestion`
sits commented out with a "currently produces a stack overflow" note).

## Already ported (no action)

LUCENE-5662, LUCENE-5599/5600, LUCENE-5653, LUCENE-5559, LUCENE-5644,
LUCENE-5671 are all present, with matching tests for 5653, 5559, and 5599/5600.

### Batch 1 — Trivial production fixes (1-5 lines each)

PR: #1413 

- [X] **LUCENE-5635**: move `stream.Reset()` inside the `try` block in
  `DocInverterPerField.cs` (~line 107) so an exception on reset is handled.
- [X] **LUCENE-5668**: change `eligible.Count >= allowedSegCountInt` to `>` in
  `TieredMergePolicy.cs` (~line 415); port the `testUnbalancedMergeSelection`
  test.
- [X] **LUCENE-5656**: move `dvGens.Add(gen)` inside the per-gen loop in
  `SegmentReader.cs` (~lines 197-209) instead of `AddRange` after the loop
  (rare fd leak on docvalues-update exception).
- [X] **LUCENE-5644 follow-up**: add the missing `PulseAll`/`notifyAll` at the
  end of `DeactivateUnreleasedStates()` in `DocumentsWriterPerThreadPool.cs`.

[...]

PR #1413, opened 2026-07-08
Port trivial production fixes from Lucene 4.8.1, #1381 Batch 1

paulirwin:

Backport trivial production fixes from Lucene 4.8.1

Partial #1381 (Batch 1)

## Description

- LUCENE-5635 (partial): move `stream.Reset()` into the `try` block
- LUCENE-5668: fix off-by-one error in TieredMergePolicy + test
- LUCENE-5656 (partial): don't leak dv producers if one of them throws exception
- LUCENE-5644 (partial): add `PulseAll` call to `DeactivateUnreleasedStates`

Note that the partial items will be fully completed in Batch 5.

### DIFF
--- a/src/Lucene.Net.Tests/Index/TestTieredMergePolicy.cs
+++ b/src/Lucene.Net.Tests/Index/TestTieredMergePolicy.cs
@@ -274,5 +274,38 @@ namespace Lucene.Net.Index
 
             // TODO: Add more checks for other non-double setters!
         }
+
+        // LUCENE-5668
+        [Test]
+        public void TestUnbalancedMergeSelection()
+        {
+            Directory dir = NewDirectory();
+            IndexWriterConfig iwc = new IndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random));
+            TieredMergePolicy tmp = (TieredMergePolicy)iwc.MergePolicy;
+            tmp.FloorSegmentMB = 0.00001;
+            iwc.MergeScheduler = new SerialMergeScheduler();
+            iwc.MaxBufferedDocs = 100;
+            iwc.RAMBufferSizeMB = -1;
+            IndexWriter w = new IndexWriter(dir, iwc);
+            for (int i = 0; i < 100000; i++)
+            {
+                Document doc = new Document();
+                doc.Add(NewTextField("id", Random.NextInt64() + "" + Random.NextInt64(), Field.Store.YES));
+                w.AddDocument(doc);
+            }
+
+            IndexReader r = DirectoryReader.Open(w, true);
+
+            // Make sure TMP always merged equal-number-of-docs segments:
+            foreach (AtomicReaderContext ctx in r.Leaves)
+            {
+                int numDocs = ctx.Reader.NumDocs;
+                Assert.IsTrue(numDocs == 100 || numDocs == 1000 || numDocs == 10000, $"got numDocs={numDocs}");
+            }
+
+            r.Dispose();
+            w.Dispose();
+            dir.Dispose();
+        }
     }
 }
--- a/src/Lucene.Net.Tests/Index/TestTieredMergePolicy.cs
+++ b/src/Lucene.Net.Tests/Index/TestTieredMergePolicy.cs
@@ -6,6 +6,10 @@ using NUnit.Framework;
 using System;
 using Assert = Lucene.Net.TestFramework.Assert;
 
+#if !FEATURE_RANDOM_NEXTINT64_NEXTSINGLE
+using RandomizedTesting.Generators; // for Random.NextInt64 extension method
+#endif
+
 namespace Lucene.Net.Index
 {
     /*
@@ -279,14 +283,14 @@ namespace Lucene.Net.Index
         [Test]
         public void TestUnbalancedMergeSelection()
         {
-            Directory dir = NewDirectory();
+            using Directory dir = NewDirectory();
             IndexWriterConfig iwc = new IndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random));
             TieredMergePolicy tmp = (TieredMergePolicy)iwc.MergePolicy;
             tmp.FloorSegmentMB = 0.00001;
             iwc.MergeScheduler = new SerialMergeScheduler();
             iwc.MaxBufferedDocs = 100;
             iwc.RAMBufferSizeMB = -1;
-            IndexWriter w = new IndexWriter(dir, iwc);
+            using IndexWriter w = new IndexWriter(dir, iwc);
             for (int i = 0; i < 100000; i++)
             {
                 Document doc = new Document();
@@ -294,7 +298,7 @@ namespace Lucene.Net.Index
                 w.AddDocument(doc);
             }
 
-            IndexReader r = DirectoryReader.Open(w, true);
+            using IndexReader r = DirectoryReader.Open(w, true);
 
             // Make sure TMP always merged equal-number-of-docs segments:
             foreach (AtomicReaderContext ctx in r.Leaves)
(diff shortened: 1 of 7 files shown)


paulirwin:
The OfflineSorter changes were reverted and moved to #1422.

src/Lucene.Net.Tests/Index/TestTieredMergePolicy.cs
```diff
@@ -274,5 +278,39 @@ public virtual void TestSetters()
...
+        // LUCENE-5668
+        [Test]
+        public void TestUnbalancedMergeSelection()
+        {
+            using Directory dir = NewDirectory();
+            IndexWriterConfig iwc = new IndexWriterConfig(TEST_VERSION_CURRENT, new MockAnalyzer(Random));
+            TieredMergePolicy tmp = (TieredMergePolicy)iwc.MergePolicy;
+            tmp.FloorSegmentMB = 0.00001;
+            iwc.MergeScheduler = new SerialMergeScheduler();
+            iwc.MaxBufferedDocs = 100;
```
NightOwl888, line 291: It is not critical to fix, but do note that there is an `IndexWriterConfigExtensions` class that can be used to bring the syntax closer to Lucene. We primarily use this in tests - for users it is an optional add-on in most cases. But there may be use cases where the fluent builder pattern is also useful.

src/Lucene.Net.Tests/Index/TestTieredMergePolicy.cs
```diff
@@ -274,5 +278,39 @@ public virtual void TestSetters()
...
+            iwc.RAMBufferSizeMB = -1;
+            using IndexWriter w = new IndexWriter(dir, iwc);
+            for (int i = 0; i < 100000; i++)
+            {
+                Document doc = new Document();
+                doc.Add(NewTextField("id", Random.NextInt64() + "" + Random.NextInt64(), Field.Store.YES));
```
NightOwl888, line 297: Not sure whether it matters for the test, but `Random.NextInt64()` produces both positive and negative values. Converting it to a string in a culture-sensitive way like this means that it will sometimes insert an alternative negative sign. So, this isn't going to be the exact same test as in Java given the same random numbers if the random culture is one with a non-default negative sign.

src/Lucene.Net/Index/DocInverterPerField.cs
```diff
@@ -103,11 +103,12 @@ public override void ProcessFields(IIndexableField[] fields, int count)
                     bool succeededInProcessingField = false;
 
                     TokenStream stream = field.GetTokenStream(docState.analyzer);
-                    // reset the TokenStream to the first token
-                    stream.Reset();
 
                     try
                     {
+                        // reset the TokenStream to the first token
```
NightOwl888, line 109: Please add a comment here with a commit hash and JIRA number to indicate why we diverged from Lucene 4.8.0 here.

src/Lucene.Net/Index/DocumentsWriterPerThreadPool.cs
```diff
@@ -283,6 +283,9 @@ internal void DeactivateUnreleasedStates()
                         threadState.Unlock();
                     }
                 }
+
+                // In case any threads are waiting for indexing:
+                UninterruptableMonitor.PulseAll(this);
```
NightOwl888, line 288: Please add a comment here with a commit hash and/or JIRA number to indicate why we diverged from Lucene 4.8.0 here.

src/Lucene.Net/Index/TieredMergePolicy.cs
```diff
@@ -412,7 +412,7 @@ public override MergeSpecification FindMerges(MergeTrigger mergeTrigger, Segment
                     return spec;
                 }
 
-                if (eligible.Count >= allowedSegCountInt)
+                if (eligible.Count > allowedSegCountInt)
```
NightOwl888, line 415: Please add a comment here with a commit hash and JIRA number to indicate why we diverged from Lucene 4.8.0 here.

src/Lucene.Net/Index/SegmentReader.cs
```diff
@@ -199,14 +199,13 @@ private void InitDocValuesProducers(Codec codec)
                 long gen = e.Key;
                 IList<FieldInfo> infos = e.Value;
                 DocValuesProducer dvp = segDocValues.GetDocValuesProducer(gen, si, IOContext.READ, dir, dvFormat, infos, TermInfosIndexDivisor);
+                dvGens.Add(gen);
```
NightOwl888, line 202: Please add a comment here with a commit hash and JIRA number to indicate why we diverged from Lucene 4.8.0 here.

src/Lucene.Net/Index/SegmentReader.cs
```diff
@@ -199,14 +199,13 @@ private void InitDocValuesProducers(Codec codec)
...
                 foreach (FieldInfo fi in infos)
                 {
                     dvProducersByField[fi.Name] = dvp;
                     dvProducers.Add(dvp);
                 }
             }
-
-            dvGens.AddRange(genInfos.Keys);
```
NightOwl888, line 209: Please add a comment here with a commit hash and JIRA number to indicate why we diverged from Lucene 4.8.0 here.

NightOwl888 (changes requested):
Mostly looks okay, but we really ought to be commenting any places where we diverged from upstream.

> Side note: I am not sure how we are going to reconstruct a description so we can determine which patches (or ugh, partial patches) we have included. But, if we keep diverging from a specific upstream release commit, it would probably be useful to construct a Lucene branch that includes all of the equivalent code changes so it is possible to debug in a coherent way. Our most useful debugging tool is to execute both in Java and .NET in order to see where the paths diverge, but this approach makes it impossible to tell if we are running a patch that doesn't exist in a specific Lucene version or if there is an actual difference that constitutes a bug.
>
> But even if keeping a Lucene version that is similar to ours isn't possible, it may still be useful to document all of the patches to 4.8.0 and/or 4.8.1 we have added.

src/Lucene.Net/Index/DocInverterPerField.cs
paulirwin: I don't understand... we're not "diverging from Lucene 4.8.0" per se; we're porting Lucene 4.8.1. This does not diverge from 4.8.1 because it is a 4.8.1 change. See issue #1381. We can fix the library versioning separately before release (i.e. calling it 4.8.1-beta00019 or whatever).

src/Lucene.Net.Tests/Index/TestTieredMergePolicy.cs
paulirwin: Good catch, this would have been caught by the other analyzer (once we can get around to fixing its findings...). Will fix.

src/Lucene.Net/Index/DocInverterPerField.cs
NightOwl888: Technically, AFAIK all of the parts that existed before I started working on Lucene.NET were ported from 4.8.0. That included:

- `Lucene.Net`
- `Lucene.Net.Classification`
- `Lucene.Net.Codecs`
- `Lucene.Net.Expressions`
- `Lucene.Net.Facet`
- `Lucene.Net.Join`
- `Lucene.Net.Queries`
- `Lucene.Net.Spatial`
- `Lucene.Net.Suggest`
- `Lucene.Net.TestFramework`

Since then, I have gone through and confirmed that there were no changes in 4.8.1 in some of them (such as Lucene.Net.Facet) and marked each file as being from 4.8.1.

```
// Lucene version compatibility level 4.8.1
```

The remainder of the projects were also from 4.8.1. So, this sits squarely in 4.8.0 territory and should have comments indicating why we diverged from 4.8.0.

> No objections if you want to start a concerted effort to bump the rest of the project to 4.8.1.
>
> I did some research and discovered that Lucene 4.10.0 did a rare divergence from semver and it included several breaking changes from 5.x. Apparently, they have not put breaking changes into another major release than 4.x. So, after the release, we must stay on this side of 4.10. Or alternatively, we should port to that level before the release (or at least include all of the breaking changes).

paulirwin: #1381 (of which this PR is one of 5 batches of work) _is_ the concerted effort to bump the rest of the project to 4.8.1. I went commit by commit (well, with the help of Claude Code, and validated by hand) and analyzed what the deltas were. Once all 5 batches are merged, Lucene.NET will be exactly matching 4.8.1.

PR #1416, opened 2026-07-08
fix(ReplicatorTestCase): Replace deprecated WebHostBuilder with HostBuilder (#1411)

### LINKED ISSUE #1411: Fix WebHostBuilder deprecation warning in tests

### Task description

WebHostBuilder is deprecated and gives this warning:

```
67>ReplicatorTestCase.cs(88,35): Warning ASPDEPR004 : 'WebHostBuilder' is obsolete: 'WebHostBuilder is deprecated in favor of HostBuilder and WebApplicationBuilder. For more information, visit https://aka.ms/aspnet/deprecate/004.' (https://aka.ms/aspnet/deprecate/004)
```

This is only used for tests, so it is low priority, but we should fix it as it suggests to both silence the warning and help future upgrades.

ShashwatSinha03:
### Description

Fixes #1411

Replaces the deprecated `WebHostBuilder` with `HostBuilder` + `ConfigureWebHostDefaults` to resolve the `ASPDEPR004` build warning in tests.

### Changes

- Replaced `new WebHostBuilder()` with `new HostBuilder().ConfigureWebHostDefaults(...)` in both paths of `NewHttpServer()`
- Changed from `new TestServer(builder)` to `host.GetTestServer()` to work with the generic host
- Added `using Microsoft.Extensions.Hosting` import
- Moved service configuration and startup class registration inside `ConfigureWebHostDefaults`

### Testing

This change only affects test infrastructure. The existing tests should continue to pass as the behavior is functionally identical.

### DIFF
--- a/src/Lucene.Net.Tests.Replicator/ReplicatorTestCase.cs
+++ b/src/Lucene.Net.Tests.Replicator/ReplicatorTestCase.cs
@@ -12,6 +12,7 @@ using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.AspNetCore.TestHost;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 #else
 using Lucene.Net.Replicator.Net;
 #endif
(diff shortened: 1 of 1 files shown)


ShashwatSinha03:
The latest commit aadd72908 adds webBuilder.UseTestServer() to the HostBuilder configuration (which was missing in the initial migration from WebHostBuilder). This should resolve the InvalidCastException where a real Kestrel server was being created instead of a TestServer.

ShashwatSinha03:
**Update:** The initial `UseTestServer()` fix alone wasn't sufficient because `ConfigureWebHostDefaults()` applies Kestrel defaults **after** our configuration delegate runs, overriding `UseTestServer()`. 

Changed both branches from `.ConfigureWebHostDefaults()` to `.ConfigureWebHost()` so that only `TestServer` is configured as the web server without Kestrel being added afterwards.

Still needs a maintainer to approve the workflow runs to verify the fix. Thank you!

paulirwin (approved):
Looks good, thanks for the contribution! Happy to have this warning resolved.

PR #1420, opened 2026-07-12
Port RateLimiter fixes from Lucene 4.8.1, #1381

paulirwin:

Backport RateLimiter fixes from Lucene 4.8.1

Partial #1381 (Batch 2)

## Description

This backports the LUCENE-5641 RateLimiter fix from Lucene 4.8.1. The original impetus for this fix in Lucene was not a problem in .NET because we were already doing the integer division correctly, but this does fix a race by synchronizing around the logic too. So it's worth backporting for that, as well as to stay in sync with upstream. And, this adds a test as well.

See this comment for some more context: https://github.com/apache/lucenenet/issues/1381#issuecomment-4949561558

### DIFF
--- a/src/Lucene.Net.Tests/Store/TestRateLimiter.cs
+++ b/src/Lucene.Net.Tests/Store/TestRateLimiter.cs
@@ -1,5 +1,11 @@
+using J2N;
+using J2N.Threading;
+using J2N.Threading.Atomic;
+using Lucene.Net.Util;
 using NUnit.Framework;
+using System;
 using Assert = Lucene.Net.TestFramework.Assert;
+using ThreadInterruptedException = Lucene.Net.Util.ThreadInterruptedException;
 
 namespace Lucene.Net.Store
 {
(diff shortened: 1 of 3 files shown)


paulirwin:
An update... there was the potential for this test to be flaky (per its own comment in Java). I ran it many times on my machine and it was passing reliably, but there was the possibility (particularly on slow ADO agents I would presume) for this to possibly fail due to timing. It turns out there were subsequent updates to this test in Lucene 6.0.0 and Lucene 7.0.0 in commits 090b804 and a893aaa that improved the reliability of this test on slow machines, so I went ahead and backported those as well.

src/Lucene.Net.Tests/Store/TestRateLimiter.cs
```diff
@@ -46,5 +52,67 @@ public void TestPause()
...
             Assert.IsTrue(convert > 1000L, "we should sleep at least 1 second but did only: " + convert + " millis");
         }
+
+        [Test]
+        public void TestThreads()
+        {
+            double targetMBPerSec = 10.0 + 20 * Random.NextDouble();
+            SimpleRateLimiter limiter = new SimpleRateLimiter(targetMBPerSec);
+
+            CountdownLatch startingGun = new CountdownLatch(1);
```
NightOwl888: Please add a `using` block here.

NightOwl888 (dismissed):
Looks good except for the undisposed `CountdownLatch`.

PR #1422, opened 2026-07-12
Port Lucene 4.8.1 fixes for AnalyzingSuggester, OfflineSorter, and SpecialOperations, #1381

paulirwin:

Backport Lucene 4.8.1 fixes for AnalyzingSuggester, OfflineSorter, and SpecialOperations

Partial #1381 (Batch 4)

## Description

This PR backports commits 8c4e8268181 (LUCENE-5628) and 9d10d6f92cc (LUCENE-5660) from Lucene 4.8.1. This has the added benefit of re-enabling the commented-out TestTooLongSuggestion test!

Batch 1 (PR #1413) incorrectly included the OfflineSorter change; that is actually part of LUCENE-5660; it was just difficult to tell that since the commit message did not mention a JIRA number. I'm going to remove that from that PR since it is necessary here for the test to pass.

### DIFF
--- a/src/Lucene.Net.Suggest/Suggest/Analyzing/AnalyzingSuggester.cs
+++ b/src/Lucene.Net.Suggest/Suggest/Analyzing/AnalyzingSuggester.cs
@@ -979,7 +979,8 @@ namespace Lucene.Net.Search.Suggest.Analyzing
             ReplaceSep(automaton);
             automaton = ConvertAutomaton(automaton);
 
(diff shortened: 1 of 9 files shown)


paulirwin:
I updated the API consistency type scanner to exclude private classes due to the PathNode failure there. These are not part of the API surface and thus should not be considered for these rules. This will help us avoid having to add other similar exclusions manually.

PR #1425, opened 2026-07-19
fix(docs): keep right menu below sticky header

### LINKED ISSUE #1407: Docs site right side menu layout bug affects desktop browsers above the 1200 break point

### Task description

There is a regression in the docs site after version "16" and version 17 and 18 have this layout display bug. See the 3 images below taken on Safari Desktop.  To open this developer view on macOS 26 Safari desktop you press: `Command + Control + r` all at once. So once you go higher than the 1200 break point the bug appears on 17 and 18. Seems to be the same on Chrome Desktop. So the right hand menu is not 100% visible:

Image

Image

Image

w3lld1:
- [x] I've read the [Contributor Guide](https://github.com/apache/lucenenet/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html).
- [x] I've included unit or integration tests for my change, where applicable.
- [x] I've included inline docs for my change, where applicable.
- [x] There's an open issue for the PR that I am making.

Keep the desktop docs side menu below the sticky header

Fixes #1407

## Description

I increased the desktop offset for the API docs side affix so its first entries no longer sit underneath the wrapped sticky header at viewport widths of 1200px and above.

I verified the change against the current 4.8.0-beta00018 docs page by replacing its stylesheet in headless Chromium with this branch's stylesheet. At 1200, 1215, 1280, 1350, and 1600px, the 195px-tall header ends before the side menu starts at 200px, and every visible menu entry remains inside the viewport.

## Validation

- `uvx prek run --files websites/apidocs/Templates/LuceneTemplateAssets/styles/main.css`
- Headless Chromium layout regression check at 1200, 1215, 1280, 1350, and 1600px
- `git diff --check origin/master...HEAD`

I did not run the complete Windows-based DocFX website build locally; the browser regression check exercises the changed stylesheet against the published docs markup.

### DIFF
--- a/websites/apidocs/Templates/LuceneTemplateAssets/styles/main.css
+++ b/websites/apidocs/Templates/LuceneTemplateAssets/styles/main.css
@@ -780,9 +780,9 @@ button:focus-visible {
       width: auto !important; /* back to natural width */
     }
 
-    /* Reverts 9.1 sideaffix margin change, so they are more uniform with the desktop mode one */
+    /* Keep the right affix below the wrapped sticky header */
     .sideaffix {
-      margin-top: 50px !important;
+      margin-top: 100px !important;
     }
   }
 


paulirwin:
@w3lld1 Thanks for the contribution! I think the HTML/CSS is going to need a bigger overhaul than this, but at least this fixes the issue for now. That way, by merging this, we at least get it fixed if we can't get around to cleaning up the HTML/CSS by the next release. Merging!

Issue #1428, opened 2026-07-29
Hanging process


marionoack:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

In rare cases on network shares it could happens, that CloseIndex has trouble with File.Delete and blocks then whole application. Additional after ending the process, the filelock is still in memory (network share don't recognise it).

### Expected Behavior

If CloseIndex fails on FileDelete, there should be no deadlock

### Steps To Reproduce

I found no way to reproduce it. I created several test applications which create and update an index regularly and sometimes I need only 1 step, sometimes more than 1000. My network expertise is not deep enough to find the real case why File.Delete will sometimes fail.

### Exceptions (if any)

If I catch the hanging thread, I get the following stacktrace:

   bei Microsoft.Win32.Win32Native.DeleteFile(String path)
   bei Microsoft.Win32.Win32Native.DeleteFile(String path)
   bei System.IO.File.InternalDelete(String path, Boolean checkHost)
   bei Lucene.Net.Store.FSDirectory.DeleteFile(String name)
   bei Lucene.Net.Index.IndexFileDeleter.DeleteFile(String fileName)
   bei Lucene.Net.Index.IndexFileDeleter.DeleteNewFiles(ICollection`1 files)
   bei Lucene.Net.Index.IndexWriter.DeleteNewFiles(ICollection`1 files)
   bei Lucene.Net.Index.IndexWriter.ProcessEvents(ConcurrentQueue`1 queue, Boolean triggerMerge, Boolean forcePurge)
   bei Lucene.Net.Index.IndexWriter.DoFlush(Boolean applyAllDeletes)
   bei Lucene.Net.Index.IndexWriter.Shutdown(Boolean waitForMerges)
   bei Lucene.Net.Index.IndexWriter.Dispose(Boolean disposing, Boolean waitForMerges)
   bei Lucene.Net.Index.IndexWriter.Dispose()

### Lucene.NET Version

4.8.0 beta 18

### .NET Version

4.8 x64

### Operating System

Windows 11

paulirwin:
Closing per discussion in #1429

PR #1373, opened 2026-06-21
Fix MMapDirectory performance and crashes via raw memory access and drain-reclaiming, #1151, #1013

paulirwin:
- [X] You've read the [Contributor Guide](https://github.com/apache/lucenenet/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://www.apache.org/foundation/policies/conduct.html).
- [X] You've included unit or integration tests for your change, where applicable.
- [X] You've included inline docs for your change, where applicable.
- [X] There's an open issue for the PR that you are making. If you'd like to propose a change, please [open an issue](https://github.com/apache/lucenenet/issues/new/choose) to discuss the change or find an existing issue.

An alternative approach to fixing MMapDirectory, via a DrainReclaimer class.

Fixes #1013
Fixes #1151

## Description

This is an alternative approach to #1361, as a stacked PR on top of that for comparison. Much of the underlying work is the same, but this fixes the dilemma of risking AVEs vs non-deterministic GC cleanup. 

First, some benchmarks:
| Design | MMap parallel | req/s | MMap serial | req/s | vs SimpleFS (parallel) | vs Beta 17 MMap (parallel) |
|---|---|---|---|---|---|---|
| **beta17** (4.8.0-beta00017) | 460.49s | 217 | 7.86s | 12,722 | **260x slower** | (n/a) |
| **finalizer** (mmap-redesign) | 0.64s | 156,000 | 3.38s | 29,600 | **2.8x faster** | **720x faster** |
| **drain reclaimer** (mmap-reclaimer) | 0.68s | 149,000 | 3.84s | 26,100 | **2.6x faster** | **677x faster** |

(Note that I have a different machine than the prior benchmarks in #1151, but this was modeled in the same way. This new machine has more cores, and thus performs even worse under heavy concurrent load for beta 17.)

This branch/approach has a very small performance penalty compared to #1361, but allows for deterministic cleanup (does not rely on GC/finalizers), and does not have AVE risk either.

## Core Idea

The core idea of this PR is that it uses a "drain reclaimer" to track entries and exits around raw memory access. When the parent requests closure, if any readers are mid-read, it will spin-wait until all readers have exited before reclaiming the memory. This is how it prevents AVEs at the readers. The Enter/Exit calls are around short critical regions that are not expected to throw or hang, but could theoretically be de-scheduled by the OS under heavy multitasking contention. The spin-wait will spin for a few cycles then yield to allow the OS to clear the threads. This is analogous to Lucene 10.x's behavior.

[...]

PR #1429, opened 2026-07-29
Reduce lock scope in `FSDirectory.DeleteFile`, #1428

marionoack:

Related to #1428 hanging process on File.Delete

## Description

I found (with help of AI) two places where locking behaviour for Fails on file.delete could be improved. I extract the relevant part of it and would be happy, you can integrate it too.

### DIFF
--- a/src/Lucene.Net/Store/FSDirectory.cs
+++ b/src/Lucene.Net/Store/FSDirectory.cs
@@ -317,33 +317,33 @@ namespace Lucene.Net.Store
             EnsureOpen();
             string file = Path.Combine(m_directory.FullName, name);
 
-                // LUCENENET specific: We need to explicitly throw when the file has already been deleted,
-                // since FileInfo doesn't do that for us.
-                // (An enhancement carried over from Lucene 8.2.0)
-                if (!File.Exists(file))
-                {
-                    throw new FileNotFoundException("Cannot delete " + file + " because it doesn't exist.");
-                }
+            // LUCENENET specific: We need to explicitly throw when the file has already been deleted,
+            // since FileInfo doesn't do that for us.
+            // (An enhancement carried over from Lucene 8.2.0)
+            if (!File.Exists(file))
+            {
+                throw new FileNotFoundException("Cannot delete " + file + " because it doesn't exist.");
+            }
 
-            // LUCENENET specific: Do NOT hold m_syncLock while deleting. On a network share (SMB),
+            // LUCENENET NOTE: Do NOT hold m_syncLock while deleting. On a network share (SMB),
             // File.Delete can block for a long time (e.g. waiting for the server to break an
             // oplock/lease on a file this client still has open). Holding m_syncLock here would
             // block every concurrent Sync() and IndexOutput.Dispose() on this directory behind
             // a single stalled delete, freezing the whole writer. Only the m_staleFiles
             // bookkeeping below needs the lock (see remarks for the m_staleFiles field);
             // upstream Java likewise performs the delete without any directory-wide lock.
-                try
+            try
+            {
+                File.Delete(file);
+                if (File.Exists(file))
                 {
-                    File.Delete(file);
-                    if (File.Exists(file))
-                    {
-                        throw new IOException("Cannot delete " + file);
-                    }
+                    throw new IOException("Cannot delete " + file);
                 }
-                catch (Exception e)
-                {
-                    throw new IOException("Cannot delete " + file, e);
-                }
+            }
+            catch (Exception e)
+            {
+                throw new IOException("Cannot delete " + file, e);
+            }
 
             UninterruptableMonitor.Enter(m_syncLock);
             try
--- a/src/Lucene.Net/Store/FSDirectory.cs
+++ b/src/Lucene.Net/Store/FSDirectory.cs
@@ -317,10 +317,6 @@ namespace Lucene.Net.Store
             EnsureOpen();
             string file = Path.Combine(m_directory.FullName, name);
 
-            // LUCENENET Specific: See remarks for m_staleFiles field.
-            UninterruptableMonitor.Enter(m_syncLock);
-            try
-            {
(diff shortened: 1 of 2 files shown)


paulirwin:
Thanks for taking the time to report the issue and submit the PR!

The `FileShare.Delete` being missing is already a known issue that is being fixed in #1373 (the MMapDirectory overhaul)... and related to that, of note, MMapDirectory currently has a nasty performance issue (see #1151) among other problems. I would recommend switching to use SimpleFSDirectory, which already uses `FileShare.Delete`, and despite its name, will likely perform better (especially under heavy load) than MMapDirectory does today.

I think your proposed change to narrow the locking to just around the stale files hashset is reasonable; upstream doesn't synchronize around this so I don't particularly see a problem with it. I'll run it through its paces and let you know.

src/Lucene.Net/Store/FSDirectory.cs
```diff
@@ -317,10 +317,6 @@ public override void DeleteFile(string name)
             EnsureOpen();
             string file = Path.Combine(m_directory.FullName, name);
 
-            // LUCENENET Specific: See remarks for m_staleFiles field.
-            UninterruptableMonitor.Enter(m_syncLock);
-            try
-            {
                 // LUCENENET specific: We need to explicitly throw when the file has already been deleted,
```
paulirwin: Please fix the indentation in this method now that the depth has shifted.

src/Lucene.Net/Store/FSDirectory.cs
```diff
@@ -329,6 +325,13 @@ public override void DeleteFile(string name)
                     throw new FileNotFoundException("Cannot delete " + file + " because it doesn't exist.");
                 }
 
+            // LUCENENET specific: Do NOT hold m_syncLock while deleting. On a network share (SMB),
```
paulirwin: Minor, but let's change this to `LUCENENET NOTE:` - `LUCENENET specific` means we're diverging from Java in some way and explaining why. In this case, we're actually reverting to something closer to the Java code.

src/Lucene.Net/Store/MMapDirectory.cs
```diff
@@ -194,7 +194,13 @@ public override IndexInput OpenInput(string name, IOContext context)
             EnsureOpen();
             EnsureCanRead(name); // LUCENENET-specific: backported call site from Lucene 6.0.0
             var file = Path.Combine(Directory.FullName, name); // LUCENENET specific: changed to use string file name instead of allocating a FileInfo (#832)
-            var fc = new FileStream(file, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
+            // LUCENENET specific: Add FileShare.Delete (a deliberate divergence from upstream Java's RandomAccessFile,
```
paulirwin: Let's undo this change in favor of #1373 which will overhaul this file (and add FileShare.Delete as well).

paulirwin (changes requested):

src/Lucene.Net/Store/FSDirectory.cs
marionoack: fixed with commit 6cfabf1e21952d86a6f27915799d9e944dcde4d1

src/Lucene.Net/Store/FSDirectory.cs
marionoack: fixed with commit 6cfabf1e21952d86a6f27915799d9e944dcde4d1

src/Lucene.Net/Store/MMapDirectory.cs
marionoack: fixed with commit 6cfabf1e21952d86a6f27915799d9e944dcde4d1

marionoack (commented):
fixed all request with commit 6cfabf1e21952d86a6f27915799d9e944dcde4d1

paulirwin:
Thanks for jumping on the feedback so fast. I investigated this further, and I think the diagnosis is a bit different than how it first appears.

Per your stack trace, the thread is blocked inside `Win32Native.DeleteFile`, waiting in the kernel. If it were a file sharing issue, you would see a sharing violation instead. A thread parked indefinitely in `DeleteFile` means the SMB redirector is waiting on a network round-trip, most likely to break a lease on the file. This is consistent with the other things you observed in the issue.

Switching to SimpleFSDirectory would be a good diagnostic to see if it helps, but is likely not the proper solution. Regardless, it should be attempted due to likely performance gains (before our next beta lands with the fix). I also confirmed that the fixes in #1373 and #1311 don't materially fix the situation of a stalled delete, either.

This PR is a real improvement to locking though, and I intend to merge it. I verified it is safe, and all relevant tests pass. And again, it more closely matches upstream while retaining the lock around the hashset. However, it's worth noting that it doesn't stop the delete from stalling and thus blocking other operations. Your app will still need to wait out the operation or determine why that is taking so long. I confirmed there is not a deadlock here, too.

I strongly suspect something is wrong with your SMB configuration, network, security scanning/protections, or underlying hardware. You may want to consider i.e. only writing to the index locally on one server, then perhaps replicating that to the readers. Or, simpler (if it works with your architecture and scaling needs), write locally and expose the local index read-only over SMB for readers, or expose an API that reads from it locally. Upstream Lucene has several warnings about running on network filesystems like NFS (which could likely be generalized to include SMB), so it's worth ensuring it's both necessary and a good fit to use SMB to solve your problem first. Hope this helps.

marionoack:
Thank you very much for this detailed and thoughtful feedback. I can fully agree with your assessment and the conclusions you've drawn.
We are already aware that the actual challenges originate from the SMB share layer rather than Lucene itself. Addressing those underlying issues, as well as the known warnings and limitations regarding network filesystems, is something we are actively working on through a parallel internal track.
Thanks again for taking the time to investigate this so thoroughly and for confirming both the improvements of the PR and the absence of a deadlock. Your insights are very helpful and appreciated.

paulirwin (approved):

PR #1432, opened 2026-08-05
Upgrade docfx to 2.78.5, docs target to net10, and increase docs build perf

paulirwin:

Upgrade docfx to 2.78.5, docs target to net10, and increase docs build perf

## Description

This PR:
- Upgrades docfx to 2.78.5 (primarily for net10 support, but also to keep it evergreen)
- Updates the docs plugins to use the docfx 2.78.5 package, which necessitated adding the CancellationToken parameter (with bonus cancellation support too)
- Updates the docs targets to use the net10.0 target (instead of net8.0)
- Parallelizes the build to dramatically improve the performance of the docs build.

## Performance

The docs build historically has taken several minutes to run. This makes iterating on improving the docs harder. By parallelizing the Powershell build, it now takes a fraction of a time.

Command (from `websites/apidocs/`):
```shell
time pwsh ./docs.ps1 -Clean -LuceneNetVersion 4.8.0-ci
```

Before (latest master):
```
331.80 real       359.01 user        41.10 sys
```

After (this PR):
```
93.96 real       348.58 user        41.17 sys
```

Result: roughly 3.5x speed-up (on macOS arm64 M4 Max), reducing it from ~5.5min to ~1.5min. Your mileage will vary based on your OS and hardware, but it should perform better the more cores you have.

## Notes

Previously, we had been encountering a file-locking issue with the toc file, that necessitated us specifying parallelism of 1 and sleeping between runs. This contributed to the poor performance. By parallelizing instead at the process level, this conflict seems to be avoided, as I could not reproduce any locking issues after many runs. It also carefully batches them into "waves" to avoid the circular dependency issues, which might have contributed to the issue previously.

Because this uses parallelism in PowerShell, it requires PowerShell 7+, aka PowerShell Core. Windows PowerShell 5.x is no longer supported for building the docs with this change, so a check is added at script launch. PS7 is cross-platform and truly open-source, so it is a better fit for our open-source project anyways.

AI: Coded and tested with the assistance of Claude Code (Opus 5).

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "docfx": {
-      "version": "2.77.0",
+      "version": "2.78.5",
       "commands": [
         "docfx"
       ],
--- a/.github/workflows/Lucene-Net-Documentation.yml
+++ b/.github/workflows/Lucene-Net-Documentation.yml
@@ -32,6 +32,14 @@ name: 'Lucene.Net.Documentation'
 
 on:
   workflow_dispatch:
+  # 2026-08-09 - build (but do not publish) the API docs for any pull request that touches the docs sources, the docs
+  # plugins, or this workflow file. This lets changes to the docs build be validated in CI before they are merged.
+  # The steps that push to the site repo are skipped for pull requests - see the `if:` conditions below.
+  pull_request:
+    paths:
+    - 'websites/apidocs/**/*'
+    - 'src/docs/LuceneDocsPlugins/**/*'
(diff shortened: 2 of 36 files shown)


paulirwin:
I forgot one thing to mention in the PR description: this PR also now restores the solution once at the start, then skips restore for each docfx operation. Previously, each docfx run was doing a `dotnet restore`, which was adding many seconds to the build time unnecessarily. Docfx 2.78.3 added a `noRestore` flag that allows for skipping this, which contributed to the performance gains. It would not have been possible to get this gain on our prior version of docfx, so that is yet another reason to upgrade.

Issue #1409, opened 2026-07-06
Fix issues with `FileSupport.GetCanonicalPath()` extension method


NightOwl888:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Task description

There are currently several issues with the implementation of `FileSupport.GetCanonicalPath()`, which is mostly a port from Apache Harmony. This method is on a required path for nearly all users, so it would be good to ensure that it is optimized and fully functional.

- [ ] No guard clause for `null` check, which should throw `ArgumentNullException`
- [ ] Logic is allocating 1 extra byte to the temporary array, appending a `0` byte at the end, converting to string, then removing the trailing `\0` character that the `0` byte creates. It would be better just to omit the null character from the buffer.
- [ ] The part that resolves symbolic links was not ported (there is a `LUCENENET TODO` for it in the code).
- [ ] The behavior of the `GetOrAdd()` method already allows racing threads to both run the business logic without blocking. If we move the logic that builds the string into the factory parameter of this method, we can eliminate the extra `TryGetValue()` check (and the unnecessary check for a `null` string in the collection).
- [ ] J2N.Text now contains an extension method for `Encoding` that accepts a `Span<byte>` so we can allocate the temporary buffer on the stack on all supported target frameworks.

tomatotomata:
I was thinking of taking a narrow first pass here on the null guard and the extra terminator allocation, with focused tests, and leaving the symlink and cache/concurrency changes separate. That should keep the change easy to review while still addressing two concrete items from the list. Does that scope fit what you had in mind?

NightOwl888:
Any small bite we can take out of this task would be welcomed and can be done as a separate PR (Although, it should not be labeled `fixes #1409` or `closes #1409` if not closing. Instead, it should have the label `see #1409` so it doesn't close this issue).

That said, the only task that is going to take some real effort here is the symlink task, which would be understandable to take on its own. The "concurrency" change is structural. Instead of calling `TryGetValue()` at the top and `GetOrAdd()` at the bottom, we will move the body of the method that creates the string inside of the `GetOrAdd()` method as a `valueFactory` argument. That method already allows races to happen and stores the result of the winner in the `ConcurrentDictionary`. This is the behavior we already have, but less code to achieve it. It is safe to allow races here because it doesn't affect correctness and only results in one extra string allocation for each losing thread until the cache is populated. If you prefer to skip this and save it for later, it is fine, but it isn't a lot of extra work and it would be appreciated if you would include it.

PR #1440, opened 2026-08-13
Guard GetCanonicalPath against null

tomatotomata:
## Summary\n\n- throw ArgumentNullException when GetCanonicalPath receives a null FileSystemInfo\n- add a focused regression test\n\nThis is the small, separate null-contract piece from #1409. I ran the focused Lucene.Net.Support.IO.TestFileSupport tests on .NET 9 with the repository's test project override: 6 passed. The repository currently targets .NET 10 by default, which is not installed in this environment, so I kept the test command and result explicit for review.


NightOwl888 (approved):

NightOwl888:
Thanks for the PR!

PR #1443, opened 2026-09-01
Remove outdated test framework `.nosupport` files

paulirwin:

Remove outdated test framework `.nosupport` files

## Description

Found during a review of the codebase. These are placeholder files for unit test projects for various alternative flavors of our test framework have not been valid in years (and not significantly touched since 2019). Since the original projects they reference are no longer there, it makes sense to remove these too. Particularly since they reference no-longer-supported TFMs, too.


NightOwl888 (approved):
Thanks for taking out the trash!

PR #1447, opened 2026-09-04
chore(xml): standardize XML declaration encoding to lowercase utf-8 (#1336)

### LINKED ISSUE #1336: Should we standardize the XML declarations and encodings case ?
Very minor issue for potential clean up and standardisation. Searching the codebase finds:
 
- 21 for `<?xml version="1.0" encoding="UTF-8"?>`
- 90 for `<?xml version="1.0" encoding="utf-8"?>`
- 2 for `<?xml version="1.0"?>`
- 120 for `<?xml version="1.0"...`

From Google below

---

The short answer is **no, it doesn't need to be capitalized, but it can be.** According to the official W3C XML specifications, character encoding names are **case-insensitive**. That means XML parsers will treat all of the following exactly the same way:

* `encoding="utf-8"` (Most common)
* `encoding="UTF-8"` (Also very common)
* `encoding="Utf-8"`

---

### Best Practices

While both work perfectly, **`utf-8` (lowercase) or `UTF-8` (uppercase) are the widely accepted standards.** Here is why developers generally lean toward one or the other:

* **Lowercase (`utf-8`)**: This is the default output for many modern web frameworks, tools, and build engines. It keeps the string uniform with the rest of your lowercased XML attributes (like `version="1.0"`).
* **Uppercase (`UTF-8`)**: This aligns with how the Internet Assigned Numbers Authority (IANA) officially registers the name.

### The One Catch

While the XML parser itself won't care, some older or poorly written legacy systems (like custom regex scripts parsing the file as raw text instead of using a proper XML engine) might specifically look for `UTF-8` or `utf-8`.

> **Recommendation:** Pick one style and stay consistent across your project. Lowercase `utf-8` is a safe, modern bet.

amasen02:
## Summary

Standardizes the XML declaration encoding casing to lowercase `encoding="utf-8"` across the codebase, as proposed and discussed in #1336.

### Changes
- Updated the remaining 21 XML test files under `src/Lucene.Net.Tests.QueryParser/Xml/` and `src/java/index-compat/pom.xml` from `encoding="UTF-8"` to lowercase `encoding="utf-8"`.
- Aligns all XML declaration headers across the repository with modern conventions and existing standards in the project.

Closes #1336.

### DIFF
--- a/src/Lucene.Net.Tests.QueryParser/Xml/BooleanFilter.xml
+++ b/src/Lucene.Net.Tests.QueryParser/Xml/BooleanFilter.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <!--
  Licensed to the Apache Software Foundation (ASF) under one or more
  contributor license agreements.  See the NOTICE file distributed with
--- a/src/Lucene.Net.Tests.QueryParser/Xml/BooleanQuery.xml
+++ b/src/Lucene.Net.Tests.QueryParser/Xml/BooleanQuery.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <!--
  Licensed to the Apache Software Foundation (ASF) under one or more
  contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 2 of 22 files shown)


amasen02:
Following up on the XML declaration encoding standardization. Ready to split or reformat the changeset if the diff is too broad.

paulirwin (approved):

paulirwin:
Thanks for the PR!

PR #1448, opened 2026-09-04
Fix editorconfig-checker url and pin to 4.0.1

### LINKED ISSUE #1455: Fix editorconfig-checker download location

### Task description

The editorconfig-checker workflow has been failing recently.

```console
ec.tar.gz
  curl -sSfL https://github.com/editorconfig-checker/editorconfig-checker/releases/latest/download/ec-linux-amd64.tar.gz -o ec.tar.gz
  mkdir -p ec-temp
  tar -xzf ec.tar.gz -C ec-temp
  
  # Dynamically find the binary
  EC_PATH=$(find ec-temp -type f -name "ec-linux-amd64" -perm -111 | head -n 1)
  
  if [ -z "$EC_PATH" ]; then
    echo "❌ Could not locate ec-linux-amd64 binary in archive"
    exit 1
  fi
  
  # Move and rename to /usr/local/bin/ec
  sudo mv "$EC_PATH" /usr/local/bin/ec
  sudo chmod +x /usr/local/bin/ec
  
  # Clean up the temp dir
  rm -rf ec-temp ec.tar.gz
  shell: /usr/bin/bash -e {0}
curl: (22) The requested URL returned error: 404
Error: Process completed with exit code 22.
```

I took a look and it seems that they changed the package naming convention in the 4.0.0 release. And since we bind to the latest release, it cannot be found.

The new filename is [editorconfig-checker-linux-amd64.tar.gz](https://github.com/editorconfig-checker/editorconfig-checker/releases/download/v4.0.1/editorconfig-checker-linux-amd64.tar.gz), so the ideal fix would be to just change the filename.

However, note that there were some other breaking changes that happened in 4.0.0 that we should review to determine whether they impact our simple use case and make the appropriate changes.

paulirwin:

Fix editorconfig-checker url and pin to 4.0.1

Fixes #1455

## Description

The download URL for the "latest" release of editorconfig-checker moved. This fixes the download URL and also pins to 4.0.1 so this isn't a moving target.

### DIFF
--- a/.github/workflows/EditorConfig-Rules-Check.yml
+++ b/.github/workflows/EditorConfig-Rules-Check.yml
@@ -34,9 +34,16 @@ jobs:
         run: |
           echo "🔍 Running EditorConfig Checker..."
 
-          if ec; then
+          # First pass emits GitHub workflow commands, which surface as inline
+          # annotations on changed lines. Annotations are not rendered for files
+          # outside the diff, so on failure we re-run in the default format to
+          # list every offending file (and line) in the log itself.
+          if ec -f github-actions; then
             echo "✅ EditorConfig Checker Success"
           else
+            echo ""
+            echo "❌ EditorConfig violations:"
+            ec -f default || true
             echo ""
             echo "📘 Common IDE Setup Help"
             echo ""
--- a/.github/workflows/EditorConfig-Rules-Check.yml
+++ b/.github/workflows/EditorConfig-Rules-Check.yml
@@ -11,7 +11,7 @@ jobs:
 
       - name: Install EditorConfig Checker
         run: |
-          curl -sSfL https://github.com/editorconfig-checker/editorconfig-checker/releases/latest/download/ec-linux-amd64.tar.gz -o ec.tar.gz
+          curl -sSfL https://github.com/editorconfig-checker/editorconfig-checker/releases/download/v4.0.1/editorconfig-checker-linux-amd64.tar.gz -o ec.tar.gz
           mkdir -p ec-temp
           tar -xzf ec.tar.gz -C ec-temp
 
--- a/.github/workflows/EditorConfig-Rules-Check.yml
+++ b/.github/workflows/EditorConfig-Rules-Check.yml
@@ -16,10 +16,10 @@ jobs:
           tar -xzf ec.tar.gz -C ec-temp
 
           # Dynamically find the binary
-          EC_PATH=$(find ec-temp -type f -name "ec-linux-amd64" -perm -111 | head -n 1)
+          EC_PATH=$(find ec-temp -type f -name "editorconfig-checker" -perm -111 | head -n 1)
 
           if [ -z "$EC_PATH" ]; then
-            echo "❌ Could not locate ec-linux-amd64 binary in archive"
+            echo "❌ Could not locate editorconfig-checker binary in archive"
             exit 1
           fi
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -291,6 +291,7 @@ dotnet_diagnostic.S907.severity = none
 
 # Files that tests read should not change from Lucene
 [**/src/Lucene.Net.Tests.*/**.txt]
+charset = unset
 trim_trailing_whitespace = unset
 insert_final_newline = unset
 resharper_enforce_empty_line_at_end_of_file = false


paulirwin:
@NightOwl888 Requesting your review since you originally added this.

I had to make a few more changes to support EditorConfig Checker 4.0.1:

- It now is more strict about a non-ASCII character in one of the Lucene test files, so that is now excluded (rather than patching the file; to keep it byte-for-byte identical to the Lucene version)
- It now no longer outputs the file name by default when run in GitHub Actions, because it is intending to add those as annotations (kinda like how `dotnet build` does if there's a warning) so you can see it in the PR. That's great, but it's also helpful to see the files in question in the command-line output, especially if the file was not touched in the PR. So I now run it twice if it fails the first time. The first run adds the GH actions annotations, the second run with the argument to use default mode outputs the files that are affected.

jbampton:
We should run the editorconfig checker with prek.  Some of the other ASF projects already do this and having the editorconfig checks/tests run on your local machine is faster and saves CI time.  You get immediate feedback.

Seems to be more effective in finding formatting and linting errors as well

https://github.com/apache/shiro/blob/ca444c3332b5996a1c7c75eed610a9598c0ca5a9/.pre-commit-config.yaml#L231

https://github.com/editorconfig-checker/editorconfig-checker.python

.github/workflows/EditorConfig-Rules-Check.yml
```diff
@@ -11,15 +11,15 @@ jobs:
 
       - name: Install EditorConfig Checker
         run: |
-          curl -sSfL https://github.com/editorconfig-checker/editorconfig-checker/releases/latest/download/ec-linux-amd64.tar.gz -o ec.tar.gz
+          curl -sSfL https://github.com/editorconfig-checker/editorconfig-checker/releases/download/v4.0.1/editorconfig-checker-linux-amd64.tar.gz -o ec.tar.gz
```
NightOwl888, line 14: I noticed in #1455 that there still is a way to target the latest version, but at the risk that a future release will change the file naming format again. But at the same time, it is probably fine to pin it and deal with periodic upgrades manually. Although, maybe there is a Renovate/Dependabot way to get notified about new releases...

NightOwl888 (approved):
Approved with minor feedback.

NightOwl888:
> We should run the editorconfig checker with prek. Some of the other ASF projects already do this and having the editorconfig checks/tests run on your local machine is faster and saves CI time. You get immediate feedback.
> 
> Seems to be more effective in finding formatting and linting errors as well
> 
> https://github.com/apache/shiro/blob/ca444c3332b5996a1c7c75eed610a9598c0ca5a9/.pre-commit-config.yaml#L231
> 
> https://github.com/editorconfig-checker/editorconfig-checker.python

@jbampton - I don't have any objections to doing this. In fact, it will reduce the number of agents we use by 1 if we add it to the combined script. Having it run locally would be great, also, although it may be most helpful if it is added either as some sort of IDE extension or possibly a build target so it isn't an extra manual step.

That said, I think that should be a new issue since we should merge this ASAP to stop the build from failing.

.github/workflows/EditorConfig-Rules-Check.yml
paulirwin: Yeah, we've been bitten by automatic upgrades without us testing it out first. It's quite annoying when they change the rules and then a new thing breaks. We should upgrade this periodically only after confirming that it doesn't break the build and we fix any new issues it finds. That's why I chose to pin it (case in point, what caused this PR in the first place).

paulirwin:
> > We should run the editorconfig checker with prek. Some of the other ASF projects already do this and having the editorconfig checks/tests run on your local machine is faster and saves CI time. You get immediate feedback.
> > Seems to be more effective in finding formatting and linting errors as well
> > https://github.com/apache/shiro/blob/ca444c3332b5996a1c7c75eed610a9598c0ca5a9/.pre-commit-config.yaml#L231
> > https://github.com/editorconfig-checker/editorconfig-checker.python
> 
> @jbampton - I don't have any objections to doing this. In fact, it will reduce the number of agents we use by 1 if we add it to the combined script. Having it run locally would be great, also, although it may be most helpful if it is added either as some sort of IDE extension or possibly a build target so it isn't an extra manual step.
> 
> That said, I think that should be a new issue since we should merge this ASAP to stop the build from failing.

Agreed on it being a new issue.

However, pre-commit hooks are better than an IDE extension. You install them once in your repo, and they run every commit, no matter how you write code (IDE, vim/notepad, or AI agent). The text output is also very AI agent friendly to fix issues. You don't have a "manual step" each commit; it runs automatically. IDE support is nice but is limited to those users using an IDE. I agree that we should pursue this in a separate issue/PR.

PR #1456, opened 2026-09-09
Group GitHub Actions test runs

paulirwin:

Group GitHub Actions test runs for less runner use and faster CI runs

## Description

While the larger test projects make sense to keep separate (the alphabetized shards, plus Analysis.Common) due to the amount of time they take to run, the smaller test projects are disproportionately harmed by being separate due to runner overhead.

Measured over the last 8 runs of each test workflow:

| Workflow | setup | restore | build | test | overhead |
|---|---|---|---|---|---|
| `_A-D` (shard) | 66s | 20s | 28s | 51s | 69% |
| `Analysis.Common` | 66s | 22s | 35s | 47s | 72% |
| `Memory` (42 tests) | 72s | 26s | 31s | 6s | **96%** |
| `Morfologik` (18 tests) | 61s | 22s | 26s | 5s | **96%** |

Median wall time is nearly flat (2.3-4.4 min) across all 36 workflows regardless of test count, because setup dominates.

By grouping these smaller projects together in GitHub Actions runs, it should allow for sharing that overhead across multiple projects and faster runs (as well as reduced agent runner use, to keep infra happy).

This PR updates the workflow generation to group projects into logically-similar groups, based on a detailed historical analysis of their runtimes:

| Workflow | Projects |
|---|---|
| `Lucene.Net.Tests.Analysis` | Kuromoji, Morfologik, Phonetic, SmartCn, Stempel, ICU |
| `Lucene.Net.Tests.Queries` | Expressions, Queries, QueryParser, Sandbox |
| `Lucene.Net.Tests.Search` | Facet, Grouping, Highlighter, Join, Spatial, Suggest |
| `Lucene.Net.Tests.TestFrameworks` | TestFramework, DependencyInjection, NUnitExtensions |
| `Lucene.Net.Tests.Misc` | AllProjects, Benchmark, Classification, Codecs, Demo, Memory, Misc, Replicator |

I will update this PR when the runs finish successfully with an analysis of the impact, but this alone will reduce 36 workflows and 209 jobs down to 14 workflows and 83 jobs.

Additionally, this will buy us headroom to add targets (or test-only TFMs) like .NET 11 soon.

AI: analysis done by and partially generated with Claude Code (Opus 5) with my review and guidance.


.github/workflows/Generate-TestWorkflows.ps1
```diff
@@ -79,6 +93,62 @@ param(
...
+# share a single build matrix. Get-TestProjectGroup intersects the frameworks of all members
+# and warns if they are not identical.
+#
+# Projects deliberately left out of the groups (each keeps its own workflow file):
+#   Lucene.Net.Tests._A-D, _E-I, _I-J, _J-S, _T-Z  - large letter shards, test time dominates
+#   Lucene.Net.Tests.Analysis.Common               - ~1,650 tests, test time dominates
+#   Lucene.Net.Tests.Cli                           - reduced matrix, plus an extra dotnet pack step
+#   Lucene.Net.Tests.CodeAnalysis                  - net8.0 only
+#   Lucene.Net.Tests.Analysis.OpenNLP              - excludes net472
+$TestProjectGroups = [ordered]@{
```
NightOwl888, line 111: Rather than tracking these groups here, it would be better if we configure this in MSBuild. Then we can also reuse this grouping for breaking the build into shards. Ideally, we would track the count of how many shards we are making and then put it into a solution-level `.props` file.

```xml
<TestShardCount>5</TestShardCount>
```

And then make a `<Target>` that can be used to look that up.

Similarly, each test project could be given its own `<TestShardNumber>` property and we could start the loop at 1 to 5 and have another target that collects the project names for a given `TestShardNumber` that is passed in on the command line.

In fact, I would go further - we currently track project dependencies below by parsing XML files. This is fine because we don't have any conditional ones, but will break if we add conditions. If we make MSBuild targets to query the project dependencies instead of parsing them, it will be more reliable.

In both cases, we don't have to think about updating this script if the projects change - we simply update project files.

